PR #8, opened 2025-12-12
support templateRef

furykerry:
add templateRef that may points to arbitrary kind so as to enable sandbox which is not based on k8s pod api


api/v1alpha1/sandbox_types.go
```diff
@@ -43,10 +43,33 @@ type SandboxSpec struct {
...
+	// +optional
+	TemplateRef *SandboxTemplateRef `json:"templateRef,omitempty"`
+
 	// Template describes the pods that will be created.
+	// Template is mutual exclusive with TemplateRef
 	// +kubebuilder:pruning:PreserveUnknownFields
 	// +kubebuilder:validation:Schemaless
-	Template v1.PodTemplateSpec `json:"template"`
+	// +optional
+	Template *v1.PodTemplateSpec `json:"template"`
```
zmberg: ,omitempty

PR #13, opened 2025-12-17
use annotations as metadata

furykerry:

this patch will use annotations instead of labels for metadata storage, and for backward compatibility, sandbox manager also try to read labels as a source of metadata. 

this patch also fix the licence and typo checks, so please  leaves the commits as is.


AiRanthem (commented):
/lgtm
/approve

PR #22, opened 2025-12-17
improve phase transition handling and update deprecated API usage

bcfre:
- **Update deprecated API**: Replace `k8s.io/utils/pointer` with `k8s.io/utils/ptr` for pointer utilities
- **Fix phase transition conditions**: 
  - Clear `Paused` condition when transitioning to `Running` phase
  - Clear `Resumed` condition when transitioning to `Paused` phase  
  - Add nil check for `Ready` condition before updating
  - Add additional validation for `Paused` condition status
- **Improve resuming logic**:
  - Add safeguard to detect pods in `Terminating` state during resume
  - Clear `Paused` condition when entering `Resuming` phase
- **Expand pause trigger**: Allow pausing from both `Running` and `Pending` phases

### DIFF
--- a/pkg/controller/sandbox/sandbox_controller.go
+++ b/pkg/controller/sandbox/sandbox_controller.go
@@ -160,7 +160,10 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			utils.SetSandboxCondition(newStatus, *cond)
 		}
 		// If it is paused, first set the sandbox to the Paused state.
+		// To prevent loss of state information, the state immediately before Paused must currently be Running.
 	} else if box.Spec.Paused && box.Status.Phase == agentsv1alpha1.SandboxRunning {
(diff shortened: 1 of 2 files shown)


pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -160,7 +160,7 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			utils.SetSandboxCondition(newStatus, *cond)
 		}
 		// If it is paused, first set the sandbox to the Paused state.
-	} else if box.Spec.Paused && box.Status.Phase == agentsv1alpha1.SandboxRunning {
+	} else if box.Spec.Paused && (box.Status.Phase == agentsv1alpha1.SandboxRunning || box.Status.Phase == agentsv1alpha1.SandboxPending) {
```
zmberg: We should not allow pending sandboxes to enter paused state.

PR #28, opened 2025-12-18
sandbox support inplace update proposal

zmberg:

### DIFF
--- a/docs/proposals/20250218-sandbox-inplace-update.md
+++ b/docs/proposals/20250218-sandbox-inplace-update.md
@@ -1,0 +1,217 @@
+---
+title: Sandbox support inplace update image
+authors:
+  - "@zmberg"
+reviewers:
+  - "@AiRanthem"
+  - "@furykerry"
+creation-date: 2025-12-18
+status: implementable
+---
+
+# Sandbox support inplace update image
+
+## Motivation
+In the warm-up pool scenario, a Sandbox may initially use a base image with lower resource allocations. If selected by an upper layer (e.g., E2B), it must be replaced with a specific image and have its resource configuration upgraded.
+
+Therefore, Sandbox must support in-place upgrades for both images and resources.
+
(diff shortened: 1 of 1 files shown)


AiRanthem (commented):
/lgtm
Is it better to add a "Upgrading" phase to Sandbox ?

docs/proposals/20250218-sandbox-inplace-update.md
```diff
@@ -0,0 +1,176 @@
...
+apiVersion: agents.kruise.io/v1alpha1
+kind: Sandbox
+metadata:
+  name: sample
+spec:
+  pause: false
+  template: 
+    ...
+status:
+  revision: f8d7967bc
```
furykerry, line 34: how to calculate revision?

docs/proposals/20250218-sandbox-inplace-update.md
```diff
@@ -0,0 +1,176 @@
...
+    pod-template-hash: f8d7967bc
+```
+
+revision: Calculated based on the sandbox template. If the sandbox revision matches the pod's revision, the versions are considered consistent. Otherwise, an in-place upgrade will be triggered.
+
+**2. Update sandbox template image or resources and revision changed.**
+
+When the sandbox template image or resources changed, the revision will change and the condition InPlaceUpdateReady, Ready will be set to false.
+
+Note: The webhook will ensure that only the sandbox template image and resources fields can be modified.
```
furykerry: we can do the validation inside the sandbox controller, i think we should avoid introduce sandbox webhook to minimize the sandbox claim flow

zmberg: Validating webhooks is essential, at the very least to ensure the template is valid. If the template fails to parse, it will cause the controller to malfunction.

furykerry (approved):
/lgtm
/approve

PR #31, opened 2025-12-19
[feat] add feature gate

yuexian1234:
Ⅰ. Describe what this PR does

This PR introduces a feature gate mechanism to selectively enable/disable controllers and webhooks in agent controller

Ⅱ. Does this pull request fix one issue?

Ⅲ. Describe how to verify it

Ⅳ. Special notes for reviews
Key changes:
Add --feature-gates flag to agent controller command-line argument

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2025 The Kruise Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package features
+
+import (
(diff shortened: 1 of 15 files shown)


pkg/features/features.go
```diff
@@ -0,0 +1,41 @@
...
+	// SandboxGate enable Sandbox-controller to create a sandbox pod.
+	SandboxGate featuregate.Feature = "Sandbox"
+
+	// SandboxSetGate enable SandboxSet-controller to create a sandbox pod.
+	SandboxSetGate featuregate.Feature = "SandboxSet"
+)
+
+var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	SandboxGate:    {Default: true, PreRelease: featuregate.Alpha},
+	SandboxSetGate: {Default: true, PreRelease: featuregate.Alpha},
```
furykerry, line 36: sandbox and sandboxset is the core controllers, it is not desirable to turn these feature off, instead i think there will be use cases where the platform users only need the sandbox controller not the sandbox manager.

pkg/utils/feature/feature.go
```diff
@@ -0,0 +1,33 @@
+/*
+Copyright 2021 The Kruise Authors.
```
furykerry: 2025

pkg/features/features.go
yuexian1234: yes, i add feature gate command-line to sandbox manager too.

PR #43, opened 2025-12-24
[Feature] Move authentication down to Sandbox

AiRanthem:
1. Remove gateway authentication
2. Activate Envd access token

### DIFF
--- a/pkg/servers/e2b/pause_resume.go
+++ b/pkg/servers/e2b/pause_resume.go
@@ -111,6 +111,6 @@ func (sc *Controller) ConnectSandbox(r *http.Request) (web.ApiResponse[*models.S
 	}
 	return web.ApiResponse[*models.Sandbox]{
 		Code: statusCode,
-		Body: sc.convertToE2BSandbox(ctx, sbx),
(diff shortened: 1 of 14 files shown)


pkg/servers/e2b/pause_resume.go
```diff
@@ -111,6 +111,6 @@ func (sc *Controller) ConnectSandbox(r *http.Request) (web.ApiResponse[*models.S
 	}
 	return web.ApiResponse[*models.Sandbox]{
 		Code: statusCode,
-		Body: sc.convertToE2BSandbox(ctx, sbx),
+		Body: sc.convertToE2BSandbox(sbx, sbx.GetAnnotations()[AnnotationEnvdAccessToken]),
```
furykerry: can we remove the second func param, and fetch the annotations inside the convertToE2BSandbox

AiRanthem: If we are going to delete this annotation later, it seems better to keep this parameter now.

AiRanthem: List API (GET /sandboxes) should not return access token

PR #44, opened 2025-12-24
[Feature] Add extensions for sandbox claiming

AiRanthem:
Currently, sandbox claiming in sandbox-manager is performed via E2B's create API. However, new features soon to be introduced in `kruise-agents` are difficult to express through the native E2B protocol. Therefore, we have defined special metadata keys to extend the semantics of the create interface.

**This patch enables extended semantics for two new capabilities:**
1. In-place upgrade: Perform an in-place image update during sandbox claiming to modify the delivered sandbox image.
2. Dynamic mounting: Trigger CSI volume mounts dynamically at the time of sandbox claiming.

### DIFF
--- a/pkg/servers/e2b/core_test.go
+++ b/pkg/servers/e2b/core_test.go
@@ -183,14 +183,6 @@ func CreateSandboxPool(t *testing.T, client versioned.Interface, name string, av
 	}
 }
 
-func AssertTimeAlmostEqual(t *testing.T, expected, actual time.Time) {
-	offset := expected.Sub(actual)
-	if offset < 0 {
-		offset = -offset
-	}
-	assert.True(t, offset < time.Second, fmt.Sprintf("actual time %s should be almost equal to expected %s", actual, expected))
-}
-
 // AvoidGetFromCache makes the resourceVersionExpectation unsatisfied to avoid getting sandbox from cache,
 // which is useful in unit tests for zero-latency update
 func AvoidGetFromCache(t *testing.T, sandboxID string, client clients.SandboxClient) {
--- a/pkg/servers/e2b/pause_resume_test.go
+++ b/pkg/servers/e2b/pause_resume_test.go
@@ -148,7 +148,7 @@ func TestConnectSandbox(t *testing.T) {
 						Type:   string(agentsv1alpha1.SandboxConditionReady),
 						Status: metav1.ConditionTrue,
 					})
-					client.ApiV1alpha1().Sandboxes(sbx.Namespace).UpdateStatus(context.Background(), sbx, metav1.UpdateOptions{})
+					_, _ = client.ApiV1alpha1().Sandboxes(sbx.Namespace).UpdateStatus(context.Background(), sbx, metav1.UpdateOptions{})
 				})
 			}
 
@@ -177,7 +177,7 @@ func TestConnectSandbox(t *testing.T) {
 				assert.Equal(t, models.SandboxStateRunning, resumeResp.Body.State)
 				endAt, err := time.Parse(time.RFC3339, resumeResp.Body.EndAt)
 				assert.NoError(t, err)
-				AssertTimeAlmostEqual(t, now.Add(time.Duration(tt.timeout)*time.Second), endAt)
+				assert.WithinDuration(t, now.Add(time.Duration(tt.timeout)*time.Second), endAt, 5*time.Second)
 			}
 		})
 	}
@@ -277,7 +277,7 @@ func TestResumeSandbox(t *testing.T) {
 						Type:   string(agentsv1alpha1.SandboxConditionReady),
 						Status: metav1.ConditionTrue,
 					})
-					client.ApiV1alpha1().Sandboxes(sbx.Namespace).UpdateStatus(context.Background(), sbx, metav1.UpdateOptions{})
+					_, _ = client.ApiV1alpha1().Sandboxes(sbx.Namespace).UpdateStatus(context.Background(), sbx, metav1.UpdateOptions{})
 				})
 			}
 
@@ -309,7 +309,7 @@ func TestResumeSandbox(t *testing.T) {
 				assert.Equal(t, models.SandboxStateRunning, describeResp.Body.State)
 				endAt, err2 := time.Parse(time.RFC3339, describeResp.Body.EndAt)
 				assert.NoError(t, err2)
-				AssertTimeAlmostEqual(t, now.Add(time.Duration(tt.timeout)*time.Second), endAt)
+				assert.WithinDuration(t, now.Add(time.Duration(tt.timeout)*time.Second), endAt, 5*time.Second)
 			}
 		})
 	}
--- a/pkg/servers/e2b/services_test.go
+++ b/pkg/servers/e2b/services_test.go
@@ -219,14 +219,14 @@ func TestCreateSandbox(t *testing.T) {
 				}
 				startedAt, err := time.Parse(time.RFC3339, sbx.StartedAt)
 				assert.NoError(t, err)
-				AssertTimeAlmostEqual(t, now, startedAt)
+				assert.WithinDuration(t, now, startedAt, 5*time.Second)
 				timeout := 300
 				if tt.request.Timeout != 0 {
 					timeout = tt.request.Timeout
 				}
 				endAt, err := time.Parse(time.RFC3339, sbx.EndAt)
 				assert.NoError(t, err)
-				AssertTimeAlmostEqual(t, startedAt.Add(time.Duration(timeout)*time.Second), endAt)
+				assert.WithinDuration(t, startedAt.Add(time.Duration(timeout)*time.Second), endAt, 5*time.Second)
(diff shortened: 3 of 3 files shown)


pkg/sandbox-manager/infra/sandboxcr/pool.go
```diff
@@ -52,66 +56,120 @@ func (p *Pool) AsSandbox(sbx *v1alpha1.Sandbox) *Sandbox {
...
-			log.Info("acquired optimistic lock of sandbox", "cost", time.Since(start), "retries", i)
-			return sbx, nil
+		utils.ResourceVersionExpectationExpect(sbx)
+		log.Info("sandbox locked")
+
+		if opts.Image != "" {
+			updateStart := time.Now()
+			log.Info("waiting for inplace update")
+			if err = p.waitForInplaceUpdate(ctx, sbx, InplaceUpdateTimeout); err != nil {
+				log.Error(err, "failed to wait for inplace update")
```
furykerry: if inplace update timeouts, the code will retry and pick another sandbox, and the previous sandbox in inplace-update process will be leaked. we should just log the error and delete the pods, asynchronously

pkg/sandbox-manager/infra/sandboxcr/pool.go
```diff
@@ -52,66 +56,120 @@ func (p *Pool) AsSandbox(sbx *v1alpha1.Sandbox) *Sandbox {
...
-		if err = p.LockSandbox(ctx, sbx, lock, user); err == nil {
-			utils.ResourceVersionExpectationExpect(sbx)
-			log.Info("acquired optimistic lock of sandbox", "cost", time.Since(start), "retries", i)
-			return sbx, nil
+		utils.ResourceVersionExpectationExpect(sbx)
+		log.Info("sandbox locked")
+
+		if opts.Image != "" {
+			updateStart := time.Now()
+			log.Info("waiting for inplace update")
```
furykerry: plz also log the sandbox name and target image name

pkg/sandbox-manager/infra/sandboxcr/pool.go
```diff
@@ -52,66 +56,120 @@ func (p *Pool) AsSandbox(sbx *v1alpha1.Sandbox) *Sandbox {
...
+				break
+			}
 		}
 	}
-	return nil, fmt.Errorf("no available sandboxes for template %s (failed to acquire optimistic lock of pod after max retries)", p.Name)
+	if len(candidates) == 0 {
+		return nil, NoAvailableError(p.Name, "no candidate")
+	}
+	obj = candidates[r.Intn(len(candidates))]
+	return p.AsSandbox(obj.DeepCopy()), nil
```
furykerry: can we delay the obj deepcopy, and only deepcopy in the modifier?

pkg/sandbox-manager/infra/sandboxcr/sandbox.go
```diff
@@ -120,6 +120,20 @@ func (s *Sandbox) SetTimeout(ttl time.Duration) {
 	s.Spec.ShutdownTime = ptr.To(metav1.NewTime(time.Now().Add(ttl)))
 }
 
+func (s *Sandbox) SetImage(image string) {
+	// TODO: 讨论存在 runtime sidecar 的情况下，如何定位到主容器
+	if s.Spec.Template != nil {
+		s.Spec.Template.Spec.Containers[0].Image = image
```
furykerry, line 143: runtime sidecar should not be visible in the sandbox template, the problem do exists if the sandbox have multiple containers.

pkg/sandbox-manager/infra/sandboxcr/pool.go
```diff
@@ -52,66 +56,120 @@ func (p *Pool) AsSandbox(sbx *v1alpha1.Sandbox) *Sandbox {
...
 		}
-		sbx.Annotations[v1alpha1.AnnotationClaimTime] = time.Now().Format(time.RFC3339)
-		if err = p.LockSandbox(ctx, sbx, lock, user); err == nil {
-			utils.ResourceVersionExpectationExpect(sbx)
-			log.Info("acquired optimistic lock of sandbox", "cost", time.Since(start), "retries", i)
-			return sbx, nil
+		utils.ResourceVersionExpectationExpect(sbx)
+		log.Info("sandbox locked")
+
+		if opts.Image != "" {
```
furykerry, line 103: we should check whether sandbox non-nil template, we cannot support inplace update for sandbox with unknown templateRef

cmd/sandbox-manager/main.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strconv"
 
 	"github.com/google/uuid"
+	"github.com/openkruise/agents/pkg/servers/e2b/models"
```
furykerry: put repository local dependency in the end of import list

pkg/sandbox-manager/infra/sandboxcr/pool.go
AiRanthem: It is checked inside sbx.SetImage, if template is nil, nothing happened

pkg/servers/e2b/services.go
```diff
@@ -68,17 +75,22 @@ func (sc *Controller) CreateSandbox(r *http.Request) (web.ApiResponse[*models.Sa
...
+			annotations := sbx.GetAnnotations()
+			if annotations == nil {
+				annotations = make(map[string]string)
+			}
+			for k, v := range request.Metadata {
+				annotations[k] = v
+			}
+			annotations[v1alpha1.AnnotationEnvdAccessToken] = accessToken
+			route := sbx.GetRoute()
+			annotations[v1alpha1.AnnotationEnvdURL] = fmt.Sprintf("http://%s:%d", route.IP, models.EnvdPort)
```
furykerry, line 90: does envd expose https url ?

pkg/sandbox-manager/infra/sandboxcr/command.go
```diff
@@ -0,0 +1,115 @@
...
+	if url == "" {
+		return RunCommandResult{}, fmt.Errorf("envd url not found on sandbox")
+	}
+	client := processconnect.NewProcessClient(
+		http.DefaultClient,
+		url,
+		connect.WithGRPC(),
+	)
+
+	clientContext, callInfo := connect.NewClientContext(ctx)
```
furykerry: instead of using timer manually(which may leak resources due to improper calling of stop func),  use context.WithTimeout to construct a child context and wait for the ctx done：

、、、
ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
clientContext, callInfo := connect.NewClientContext(ctx)defer cancel()

select {
case result := <-ch:
    // handle
case <-clientContext.Done():
	break Processor}
、、、

pkg/sandbox-manager/infra/sandboxcr/command.go
```diff
@@ -0,0 +1,115 @@
...
+	defer func() {
+		err := stream.Close()
+		if err != nil {
+			log.Error(err, "failed to close stream")
+		} else {
+			log.Info("stream closed")
+		}
+	}()
+
+	ch := make(chan *process.ProcessEvent)
```
furykerry: make ch a buffered channel

proto/envd/process/process.pb.go
```diff
@@ -0,0 +1,2353 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
furykerry, line 1: plz add README about how to generate these go source codes

pkg/servers/e2b/services.go
AiRanthem: nope

pkg/sandbox-manager/infra/sandboxcr/command.go
AiRanthem: `WithTimeout` can only handle timeout for establishing connections, and cannot control the timeout for business processing after the connection is established.

proto/envd/process/process.pb.go
AiRanthem: they are directly copied from https://github.com/e2b-dev/infra/tree/main/packages/shared/pkg/grpc/envd/process ...

pkg/sandbox-manager/infra/sandboxcr/command.go
AiRanthem: changed to WithTimeout

pkg/sandbox-manager/infra/sandboxcr/command.go
```diff
@@ -0,0 +1,94 @@
...
+			}
+
+		case *process.ProcessEvent_End:
+			result.ExitCode = evt.End.ExitCode
+			result.Exited = evt.End.Exited
+			if evt.End.Error != nil {
+				result.Error = fmt.Errorf("process error: %s", *evt.End.Error)
+			}
+
+		default: // ProcessEvent_Keepalive
```
furykerry, line 89: please add a case `case <-clientContext.Done():`

furykerry (commented):
/lgtm

Issue #30, opened 2025-12-19
[feature request] add sandbox claim CRD


furykerry:
**What would you like to be added**:
add CRD interface for sandbox claim that allocated a pre-warmed sandbox instances from sandboxset. SandboxClaim should use basically the same claim logic as the one in sandbox-manager. 

**Why is this needed**:
It is easy and fast to use high-level API such E2B API to claim a sandbox, however for platform users, they may use their own high-level apiserver, so the claim logic should be also available in K8S api

furykerry:
welcome aboard @PersistentJZH

PR #49, opened 2025-12-29
Add sandbox claim crd proposal

PersistentJZH:
add sandbox claim crd proposal

proposal for https://github.com/openkruise/agents/issues/30

### DIFF
--- a/docs/proposals/20251229-sandbox-claim-crd.md
+++ b/docs/proposals/20251229-sandbox-claim-crd.md
@@ -201,26 +201,29 @@ The SandboxClaim CRD is designed to align with the E2B HTTP API's `NewSandboxReq
 
 **Claim Process**:
 1. Track the start time when entering `Claiming` phase (for claimTimeout calculation)
-2. List available sandboxes from the specified SandboxSet (state = `available`, filtered by `IndexPoolAvailable` index)
-3. Filter candidates by checking:
+2. Calculate how many sandboxes still need to be claimed: `remaining = spec.replicas - status.claimedReplicas`
+3. List available sandboxes from the specified SandboxSet
+4. Filter candidates by checking:
    - `Status.Phase == Running` (sandbox must be running)
    - `AnnotationLock == ""` (pre-check to skip already-locked sandboxes)
-4. Randomly select a sandbox from candidates
-5. Update sandbox with:
-    - Lock annotation (UUID)
-    - Claim timestamp annotation
-    - Labels (from `claim.spec.labels`)
-    - Annotations (from `claim.spec.annotations`)
-    - Remove SandboxSet ownerReference (detach from pool)
-    - Add SandboxClaim ownerReference
-    - Set `spec.shutdownTime` (absolute time) if `shutdownTime` specified
+5. **Batch Claiming**: Attempt to claim multiple sandboxes within a single reconcile cycle:
+   - Determine batch size: `min(remaining, maxBatchSize)` where `maxBatchSize` is a configurable limit (e.g., 10)
+   - Randomly select multiple candidates (up to batch size) from the filtered list
+   - For each sandbox, update with:
+     - Lock annotation (UUID)
+     - Claim timestamp annotation
+     - Labels (from `claim.spec.labels`)
+     - Annotations (from `claim.spec.annotations`)
+     - Remove SandboxSet ownerReference (detach from pool)
+     - Add SandboxClaim ownerReference
+     - Set `spec.shutdownTime` (absolute time) if `shutdownTime` specified
 6. Inject environment variables via envd `/init` endpoint (if `claim.spec.envVars` specified)
-    - Failure to inject envVars logs error but doesn't fail the claim
-7. Check completion conditions:
-    - If `ClaimedReplicas == spec.replicas`: transition to `Completed`
-    - If `spec.claimTimeout` is set and exceeded: transition to `Completed` (regardless of claimed count)
-    - If SandboxSet is deleted: transition to `Completed` immediately
-8. If `spec.ttlAfterCompleted` is set and claim is `Completed`, schedule deletion after the TTL duration
+7. Update `status.claimedReplicas` with the actual number of successfully claimed sandboxes
+8. Check completion conditions:
+   - If `ClaimedReplicas == spec.replicas`: transition to `Completed`
+   - If `spec.claimTimeout` is set and exceeded: transition to `Completed` (regardless of claimed count)
+   - If SandboxSet is deleted: transition to `Completed` immediately
+9. If `spec.ttlAfterCompleted` is set and claim is `Completed`, schedule deletion after the TTL duration
 
 **Conflict Handling**:
 The claim process uses **optimistic locking** via Kubernetes resourceVersion to handle concurrent claims:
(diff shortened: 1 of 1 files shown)


docs/proposals/20251229-sandbox-claim-crd.md
```diff
@@ -0,0 +1,260 @@
...
+| `secure` | N/A |  Future Work | Not implemented in E2B Api |
+
+
+
+### Implementation Details
+
+**Claim Process**:
+1. List available sandboxes from the specified SandboxSet (state = Running and available)
+2. Randomly select a sandbox from candidates
+3. Update sandbox with:
```
furykerry: plz explain the conflicting handling logic, what if the sandbox is claimed by others e.g. sandbox manager

docs/proposals/20251229-sandbox-claim-crd.md
```diff
@@ -0,0 +1,260 @@
...
+    // If not specified, defaults to "<namespace>/<claim-name>"
+    // +optional
+    Owner string `json:"owner,omitempty"`
+
+    // Replicas specifies how many sandboxes to claim (default: 1)
+    // For batch claiming support
+    // +optional
+    // +kubebuilder:default=1
+    // +kubebuilder:validation:Minimum=1
+    Replicas int32 `json:"replicas,omitempty"`
```
bcfre: If the replica field is present, how should the upper-layer platform route requests to a specific sandbox instance?

PersistentJZH: @bcfre The upper-layer platform can route requests using the sandbox ID. Perhaps we could add a status item to make it easier for the upper-layer platform to get sandbox IDs in this Claim:

```
status:
  claimedSandboxes:
    - name: test-batch-0
      sandboxID: sb-abc123
      state: Running
    - name: test-batch-1
      sandboxID: sb-def456
      state: Running
```

zmberg: Is the replicas field mutable? If it is mutable, how does it correspond with the phase field when changed?

docs/proposals/20251229-sandbox-claim-crd.md
```diff
@@ -0,0 +1,434 @@
...
+reviewers:
+  - "@furykerry"
+  - "@zmberg"
+creation-date: 2025-12-29
+status: implementable
+---
+
+# SandboxClaim CRD for Declarative Sandbox Allocation
+
+This proposal introduces a **SandboxClaim** CRD that provides a Kubernetes-native, declarative API for claiming pre-warmed sandbox instances from a SandboxSet.
```
zmberg, line 14: Does the in-place upgrade and image change need to be implemented here?

docs/proposals/20251229-sandbox-claim-crd.md
PersistentJZH: @zmberg , it be better to design replicas as immutable, I'm not sure if my understanding is correct, Sandboxes are task-oriented, it's unlikely that replicas will be changed after a claim. If need to modify replicas, can create a new Claim.

or could you provide some scenarios that modifications to replicas are required?

docs/proposals/20251229-sandbox-claim-crd.md
PersistentJZH: If this is implemented in the E2B API, then we can consider implementing it in the Claim CRD.

docs/proposals/20251229-sandbox-claim-crd.md
furykerry: i agree with @PersistentJZH , claim CR is task-oriented and the fields are supposed to be immutable

docs/proposals/20251229-sandbox-claim-crd.md
```diff
@@ -0,0 +1,434 @@
...
+    - If sandbox recovers, `ReadyReplicas` increases
+
+3. **Recovery**:
+    - When all claimed sandboxes become ready
+    - Phase transitions back to `Ready`
+    - `ReadyReplicas` == `ClaimedReplicas` == `spec.replicas`
+
+#### Querying Claimed Sandboxes
+
+##### Method 1: Query SandboxClaim Status (Recommended)
```
furykerry: if the claim return allocated sandbox in status, there should be an upper limit of replicas, otherwise etcd should have problem to persist very large sandbox list

docs/proposals/20251229-sandbox-claim-crd.md
```diff
@@ -0,0 +1,434 @@
...
+    // For batch claiming support
+    // +optional
+    // +kubebuilder:default=1
+    // +kubebuilder:validation:Minimum=1
+    Replicas int32 `json:"replicas,omitempty"`
+
+    // Timeout specifies the lifetime of claimed sandboxes in seconds
+    // This will be converted to shutdownTime on the Sandbox
+    // +optional
+    Timeout *int64 `json:"timeout,omitempty"`
```
furykerry: just rename the field to shutdownTime ?

docs/proposals/20251229-sandbox-claim-crd.md
```diff
@@ -0,0 +1,434 @@
...
+    // If not specified, defaults to "<namespace>/<claim-name>"
+    // +optional
+    Owner string `json:"owner,omitempty"`
+
+    // Replicas specifies how many sandboxes to claim (default: 1)
+    // For batch claiming support
+    // +optional
+    // +kubebuilder:default=1
+    // +kubebuilder:validation:Minimum=1
+    Replicas int32 `json:"replicas,omitempty"`
```
furykerry: if the default value for replicas is 1, i think the type of replicas should be changed to *int32

docs/proposals/20251229-sandbox-claim-crd.md
```diff
@@ -0,0 +1,434 @@
...
+    // Metadata contains arbitrary key-value pairs to be added as annotations
+    // to claimed Sandbox resources
+    // +optional
+    Metadata map[string]string `json:"metadata,omitempty"`
+
+    // EnvVars contains environment variables to be injected into the sandbox
+    // These will be passed to the sandbox's init endpoint (envd) after claiming
+    // Only applicable if the SandboxSet has envd enabled
+    // +optional
+    EnvVars map[string]string `json:"envVars,omitempty"`
```
furykerry, line 129: shall we add a ttl field to cleanup the claim after claim finishes, e.g. ttlAfterFinished

docs/proposals/20251229-sandbox-claim-crd.md
```diff
@@ -0,0 +1,403 @@
...
+The claim process uses **optimistic locking** via Kubernetes resourceVersion to handle concurrent claims:
+
+- **Pre-check Filter**: Before attempting to claim, the controller filters out sandboxes that already have a `AnnotationLock` annotation. This reduces but doesn't eliminate race conditions.
+
+- **Race Window**: Between the pre-check and the actual update, another process (e.g., other sandbox-manager) may claim the same sandbox.
+
+- **Optimistic Locking**: When updating the sandbox, Kubernetes validates the resourceVersion. If the sandbox was modified by another process, the API returns a `Conflict` error.
+
+- **Declarative Retry Model**:
+  - **Per-Reconcile Attempt**: Each reconcile cycle attempts to claim one sandbox at a time
```
furykerry: why claim only one sandbox at each reconcile? Allocate more sandbox at one time should be more efficient

PersistentJZH: this should be batch processed many sandbox in one reconcile, I've updated the design for this section, please take another look.

PR #51, opened 2025-12-30
[Feature] align pause logic with native E2B

### LINKED ISSUE #50: [feature request] Sandbox pause should not consume timeout quota
# background
According to the [E2B official documentation](https://e2b.dev/docs/sandbox/persistence/pause) on sandbox pause:
`Sandbox can be used up to 30 days
After 30 days from the initial sandbox create call, the data may be deleted and you will not be able to resume it. Attempting to resume a sandbox that was deleted or does not exist will result in the NotFoundError in the JavaScript SDK or the NotFoundException in the Python SDK.`
The documentation does not explicitly state whether the pause state consumes the sandbox timeout.
Actual E2B Behavior (Verified)
Through testing, I confirmed that paused sandboxes do NOT consume timeout time in E2B's implementation.

Image

# Current Behavior
pause behavior will consume timeout time. when i create a sandbox with 60s timeout. after i paused sandbox,the sandbox will be delete after 60s. and will not be resume at all

# Proposal

To align with E2B semantics, sandbox-manager should implement the same behavior:

1. Pause should NOT consume timeout - When a sandbox is paused, the timeout timer should stop counting
Sandbox expiration - Use one of the following approaches:
2. Use maxTimeout as the absolute expiration time, OR
Follow the community standard of 30 days maximum lifetime from initial creation

AiRanthem:

1. Align pause logic with native E2B:
2. AutoPause is supported. SDK example: `Sandbox.create(auto_pause=True)`

#### if autoPause == false (default):

When setting timeout (claim / connect / resume, etc...) of a Sandbox, the following will happen to it:
- The `ShutdownTime` field of Sandbox is set to `time.Now().Add(timeoutSeconds)`

When pausing a Sandbox, the following will happen to it:
- The `ShutdownTime` field of Sandbox is set to `time.Now().Add(maxTimeoutSeconds)`

#### if autoPause == true (new feature):
A new `PauseTime` field has been added to Sandbox, similar to `ShutdownTime`, after which the Sandbox will automatically pause.

When setting timeout (claim / connect / resume, etc...) of a Sandbox, the following will happen to it:
- The `PauseTime` field of Sandbox is set to `time.Now().Add(timeoutSeconds)`
- The `ShutdownTime` field of Sandbox is set to `time.Now().Add(maxTimeoutSeconds)`

When pausing a Sandbox, the following will happen to it:
- The `ShutdownTime` field of Sandbox is set to `time.Now().Add(maxTimeoutSeconds)`

fixes #50

### DIFF
--- a/api/v1alpha1/sandbox_types.go
+++ b/api/v1alpha1/sandbox_types.go
@@ -232,6 +232,7 @@ const (
 // +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.phase"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:printcolumn:name="shutdown_time",type="string",JSONPath=".spec.shutdownTime"
+// +kubebuilder:printcolumn:name="pause_time",type="string",JSONPath=".spec.pauseTime"
 // +kubebuilder:printcolumn:name="Message",type="string",JSONPath=".status.message"
 
 // Sandbox is the Schema for the sandboxes API
--- a/config/crd/bases/agents.kruise.io_sandboxes.yaml
+++ b/config/crd/bases/agents.kruise.io_sandboxes.yaml
@@ -26,6 +26,9 @@ spec:
     - jsonPath: .spec.shutdownTime
       name: shutdown_time
       type: string
+    - jsonPath: .spec.pauseTime
+      name: pause_time
+      type: string
     - jsonPath: .status.message
       name: Message
       type: string


pkg/servers/e2b/pause_resume_test.go
```diff
@@ -277,7 +280,7 @@ func TestResumeSandbox(t *testing.T) {
 						Type:   string(agentsv1alpha1.SandboxConditionReady),
 						Status: metav1.ConditionTrue,
 					})
-					client.ApiV1alpha1().Sandboxes(sbx.Namespace).UpdateStatus(context.Background(), sbx, metav1.UpdateOptions{})
+					_, _ = client.ApiV1alpha1().Sandboxes(sbx.Namespace).UpdateStatus(context.Background(), sbx, metav1.UpdateOptions{})
```
furykerry, line 274: can we move these test that require kind environment and operator installation to e2e tests? the unit test should use fakeclient and mocked data

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -155,16 +155,26 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 		if box.Spec.ShutdownTime.Before(&now) {
-			logger.Info("Sandbox shutdown time reached，and it will be deleted.")
-			return reconcile.Result{}, r.Delete(ctx, box)
+			logger.Info("sandbox shutdown time reached, will be deleted")
+			return ctrl.Result{}, r.Delete(ctx, box)
 		}
 		requeueAfter = box.Spec.ShutdownTime.Sub(now.Time)
 	}
+	if box.Spec.PauseTime != nil && !box.Spec.Paused {
+		if box.Spec.PauseTime.Before(&now) {
```
furykerry, line 169: what if use set pause to false but not change pause time? the code seems will change back the pause to true

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -155,16 +155,26 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-	// Check Shutdown
+	// Check ShutdownTime and PauseTime
+	now := metav1.Now()
 	var requeueAfter time.Duration
 	if box.Spec.ShutdownTime != nil && box.DeletionTimestamp == nil {
-		now := metav1.Now()
 		if box.Spec.ShutdownTime.Before(&now) {
-			logger.Info("Sandbox shutdown time reached，and it will be deleted.")
-			return reconcile.Result{}, r.Delete(ctx, box)
+			logger.Info("sandbox shutdown time reached, will be deleted")
```
furykerry, line 163: the new log is syntactically incorrect, plz keep the original log,

pkg/controller/sandbox/sandbox_controller.go
AiRanthem: yes, indeed

Issue #35, opened 2025-12-22
[feature request] Force sync route in sandbox-manager after resuming some sandbox


AiRanthem:
**What would you like to be added**:

For PodIP may changed after resuming, a force-sync of routes between instances may be useful

**Why is this needed**:

PR #52, opened 2025-12-31
feat: force sync route after sandbox resume

PersistentJZH:
Add force route synchronization in sandbox-manager after resuming sandbox to ensure all peer instances are aware of route changes (especially PodIP) without waiting for informer watch events.

Changes:
- Add PauseSandbox and ResumeSandbox methods to SandboxManager
- Replace direct sbx.Pause/Resume calls in e2b controller with SandboxManager methods
- Add InplaceRefresh before route sync to ensure latest PodIP and State

https://github.com/openkruise/agents/issues/35

### DIFF
--- a/pkg/sandbox-manager/api.go
+++ b/pkg/sandbox-manager/api.go
@@ -36,14 +36,9 @@ func (m *SandboxManager) ClaimSandbox(ctx context.Context, user, template string
 
 	log.Info("sandbox claimed", "sandbox", klog.KObj(sandbox), "cost", time.Since(start))
 
-	startSync := time.Now()
-	route := sandbox.GetRoute()
-	m.proxy.SetRoute(route)
-	err = m.proxy.SyncRouteWithPeers(route)
-	if err != nil {
-		log.Error(err, "failed to sync route with peers", "cost", time.Since(startSync))
-	} else {
-		log.Info("route synced with peers", "cost", time.Since(startSync), "route", route)
+	// Sync route without refresh since sandbox was just claimed and state is already up-to-date
+	if err = m.syncRoute(ctx, sandbox, false); err != nil {
+		log.Error(err, "failed to sync route with peers after claim")
 	}
 	return sandbox, nil
 }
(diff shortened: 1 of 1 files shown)


pkg/sandbox-manager/api.go
```diff
@@ -76,3 +77,49 @@ func (m *SandboxManager) GetOwnerOfSandbox(sandboxID string) (string, bool) {
 	route, ok := m.proxy.LoadRoute(sandboxID)
 	return route.Owner, ok
 }
+
+// PauseSandbox pauses a sandbox and syncs route with peers
+func (m *SandboxManager) PauseSandbox(ctx context.Context, sbx infra.Sandbox) error {
+	if err := sbx.Pause(ctx); err != nil {
+		return err
```
AiRanthem: what if log the error ?

pkg/sandbox-manager/api.go
```diff
@@ -76,3 +77,49 @@ func (m *SandboxManager) GetOwnerOfSandbox(sandboxID string) (string, bool) {
...
+	}
+	log := klog.FromContext(ctx)
+	// Refresh sandbox to get the latest State
+	if err := sbx.InplaceRefresh(true); err != nil {
```
AiRanthem: Is the deep-copy necessary?

pkg/sandbox-manager/api.go
```diff
@@ -76,3 +77,49 @@ func (m *SandboxManager) GetOwnerOfSandbox(sandboxID string) (string, bool) {
...
+}
+
+// ResumeSandbox resumes a sandbox and syncs route with peers
+func (m *SandboxManager) ResumeSandbox(ctx context.Context, sbx infra.Sandbox) error {
+	if err := sbx.Resume(ctx); err != nil {
+		return err
+	}
+	log := klog.FromContext(ctx)
+	// Refresh sandbox to get the latest PodIP (which may change after Pod restart during resume)
+	if err := sbx.InplaceRefresh(true); err != nil {
```
AiRanthem: Is the deep-copy necessary?

pkg/sandbox-manager/api.go
```diff
@@ -76,3 +77,49 @@ func (m *SandboxManager) GetOwnerOfSandbox(sandboxID string) (string, bool) {
...
+		log.Error(err, "failed to refresh sandbox after resume, route sync may use stale IP")
+		// Continue to sync route even if refresh fails, as the route might still be valid
+	}
+	start := time.Now()
```
AiRanthem: The synchronous routing code after pause and resume looks somewhat repetitive. Could we consider extracting it into a common function?

pkg/sandbox-manager/api.go
PersistentJZH: yes, do not need deep-copy here since sbx are read only

pkg/sandbox-manager/api.go
PersistentJZH: updated

pkg/sandbox-manager/api.go
PersistentJZH: added

AiRanthem (commented):
/lgtm
@furykerry PTAL

PR #59, opened 2026-01-03
feat: Add sandbox claim label synchronization between Pod and Sandbox

### LINKED ISSUE #46: [BUG] Inconsistent sandbox tracking between status.Selector and actual ownership
**What happened**:
**Background**
1. In the scale subresource, we declare that the label `agents.kruise.io/sandbox-pool=<sbs.Name>` [link](https://github.com/openkruise/agents/blob/master/pkg/controller/sandboxset/sandboxset_controller.go#L137) is used to identify sandboxes belonging to a given SandboxSet, and thus determine the current number of child resources. However, during actual reconciliation, the SandboxSet controller tracks managed sandboxes by ownerReference.uid [link](https://github.com/openkruise/agents/blob/master/pkg/controller/sandboxset/sandboxset_controller.go#L337), not by this label.
2. The sandbox manager acquires an available sandbox by clearing its ControllerReference [link](https://github.com/openkruise/agents/blob/master/pkg/sandbox-manager/api.go#L26). When the SandboxSet controller observes that the count of sandboxes owned by a specific ownerRefUID has decreased, it creates new sandboxes to maintain the desired size of the warm pool.

**Problem**
The sandbox manager only clears the ControllerReference but does not update or remove the label agents.kruise.io/sandbox-pool=<sbs.Name>.

As a result:

1. The number of sandboxes matched by status.Selector (which relies on the label) no longer reflects the true managed count.
2. This causes a mismatch between status.Selector-based counts and SandboxSetSpec.Replicas.

This inconsistency complicates integration with autoscaling systems and makes the system state harder to reason about.

**Proposed Solution**

We should first clarify the intended semantics of SandboxSet:

1. If SandboxSet represents the entire sandbox pool (i.e., all sandboxes it manages):
    The controller should consistently use the label (agents.kruise.io/sandbox-pool) for both tracking and scaling, aligning with the scale subresource contract.

2. If SandboxSet is only a pre-warmed sandbox pool:
    Introduce a dedicated ownership label (e.g., agents.kruise.io/sandbox-owner).
      - The SandboxSet controller sets this label when assigning a sandbox.
      - The sandbox manager clears or updates this label upon taking ownership (similar to the existing v1alpha1.AnnotationOwner logic).
      This ensures a 1:1 correspondence between ownerRefUID and the label, keeping status.Selector accurate.

This change would guarantee consistency between observed state (status.Selector) and desired state (spec.replicas), improving reliability for autoscaling and observability.

I’d be happy to implement this fix!

bcfre:
This PR introduces a new label `LabelSandboxIsClaimed` to track sandbox claim status and implements label synchronization between Sandbox and Pod resources.

### 1. Core Controller Enhancement
- Implement `handleClaimSandbox()` function to synchronize labels between Sandbox and Pod
  - Add missing labels from Sandbox to Pod
  - Remove labels from Pod that don't exist in Sandbox
  - Update Pod labels when values differ from Sandbox
- Use Strategic Merge Patch with `null` values for safe label deletion (avoids conflicts with other controllers)

### 2. SandboxSet Controller Updates
- Set `LabelSandboxIsClaimed: "false"` when creating new sandboxes
- Update label selector to include `LabelSandboxIsClaimed: "false"` for better sandbox filtering

### 3. Sandbox Manager Integration
- Set `LabelSandboxIsClaimed: "true"` when a sandbox is claimed by a user

fixes #46

HPA requires the `scale` subresource to expose a `selector` field, and it uses that selector to isolate different HPAs from each other, preventing them from targeting/claiming the same Pods.
https://github.com/kubernetes/kubernetes/blob/v1.35.0/pkg/controller/podautoscaler/horizontal.go#L397
```go
func (a *HorizontalController) validateAndParseSelector(hpa *autoscalingv2.HorizontalPodAutoscaler, selector string) (labels.Selector, error) {
	if selector == "" {
		errMsg := "selector is required"
		a.eventRecorder.Event(hpa, v1.EventTypeWarning, "SelectorRequired", errMsg)
		setCondition(hpa, autoscalingv2.ScalingActive, v1.ConditionFalse, "InvalidSelector", "the HPA target's scale is missing a selector")
		return nil, errors.New(errMsg)
	}
        ...
	selectingHpas := a.hpasControllingPodsUnderSelector(pods)
	if len(selectingHpas) > 1 {
		errMsg := fmt.Sprintf("pods by selector %v are controlled by multiple HPAs: %v", selector, selectingHpas)
		a.eventRecorder.Event(hpa, v1.EventTypeWarning, "AmbiguousSelector", errMsg)
		setCondition(hpa, autoscalingv2.ScalingActive, v1.ConditionFalse, "AmbiguousSelector", "%s", errMsg)
		return nil, errors.New(errMsg)
	}
}
```
KEDA relies on HPA under the hood for autoscaling.
keda-arch

### DIFF
--- a/api/v1alpha1/sandboxset_types.go
+++ b/api/v1alpha1/sandboxset_types.go
@@ -23,8 +23,11 @@ import (
 const (
 	InternalPrefix = "agents.kruise.io/"
 
-	LabelSandboxPool  = InternalPrefix + "sandbox-pool"
-	LabelTemplateHash = InternalPrefix + "template-hash"
+	// LabelSandboxPool identifies which SandboxSet generated the sandbox
+	LabelSandboxPool = InternalPrefix + "sandbox-pool"
+	// LabelSandboxIsClaimed indicates whether the sandbox has been claimed by user
+	LabelSandboxIsClaimed = InternalPrefix + "sandbox-claimed"
+	LabelTemplateHash     = InternalPrefix + "template-hash"
 
 	AnnotationLock      = InternalPrefix + "lock"
 	AnnotationOwner     = InternalPrefix + "owner"
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/onsi/ginkgo/v2 v2.27.3
 	github.com/onsi/gomega v1.38.2
+	github.com/prometheus/client_golang v1.22.0
 	github.com/spf13/pflag v1.0.6
 	github.com/stretchr/testify v1.11.1
 	google.golang.org/grpc v1.75.1
@@ -68,7 +69,6 @@ require (
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/planetscale/vtprotobuf v0.6.1-0.20240319094008-0393e58bdf10 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	github.com/prometheus/client_golang v1.22.0 // indirect
 	github.com/prometheus/client_model v0.6.1 // indirect
 	github.com/prometheus/common v0.62.0 // indirect
 	github.com/prometheus/procfs v0.16.0 // indirect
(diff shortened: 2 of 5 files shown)


pkg/controller/sandbox/core/common_control.go
```diff
@@ -287,6 +303,120 @@ func (r *commonControl) createPod(ctx context.Context, box *agentsv1alpha1.Sandb
 	return pod, nil
 }
 
+// handleClaimSandbox synchronizes specific labels between sandbox and pod.
+// 1. If sandbox has a label but pod doesn't, add it to pod
+// 2. If sandbox doesn't have a label but pod does, delete it from pod
+// 3. If both have the label but values differ, update pod's value to match sandbox
+func (r *commonControl) handleClaimSandbox(ctx context.Context, args EnsureFuncArgs) error {
```
furykerry: why we have to add labels to the pod, who will consume these pod labels?

PR #69, opened 2026-01-06
sandbox controller add recouceversion expectation

zmberg:


pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -24,19 +24,23 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/openkruise/agents/pkg/utils/expectations"
```
furykerry: plz move the repo local dependency after the third party dependency

Issue #3, opened 2025-12-09
[feature request] support autoscaler for sandboxset


furykerry:
**What would you like to be added**:
Design a CRD for PoolingAutoScaler  (wps), which accepts two kinds of strategy to adjust the replicas of sandboxset:

1. crontab for pre-defined time-based scaling. For example "0 6 * * *" scale the replicas to 100, and "0  0 *  *  *" scale the replicas to 20. 
2. watermark control based scaling that try to keep the sandboxset utlization of a time period (e.g. ten minutes) within a user-defined range. 

**Why is this needed**:
sandboxset maintain a fixed number of ready sandboxes which can be claimed by the agent-apiserver, however the sandbox requests may undergo periodic or burst changes.  The fixed sandboxset may cause under utilized resources in idle time and sandbox claim failure in busy time. It is desirable to dynamic adjust the replicas of sandboxset based on pre-defined time schedule or dynamic watermark calculation

sivanzcw:
I'm keep on solving this, thanks.

furykerry:
welcome @sivanzcw

N3rdBot:
Is the feature in progress? It seems the last pr related to it is #77. If no one is looking into this feature yet, I'd be glad to take it over. :) @furykerry @sivanzcw

furykerry:
the contributor is pre-occupied with other work, @N3rdBot you're welcome to take it over and start the implementation

PR #77, opened 2026-01-08
initial proposal of autoscaler supporting for sandboxSet

sivanzcw:
This enhancement proposes providing autoscaler capabilities for `SandboxSet`,
enabling intelligent and dynamic management of pre-warmed sandbox resource pools.
Currently, `SandboxSet` provides sandbox pre-warming capabilities with fixed-size resource pools.
This enhancement adds two complementary autoscaling policy types:

1. **Cron-based policies**: Enable time-driven scaling based on recurring schedules
(e.g., scale up during business hours, scale down during off-peak periods)
2. **Capacity-based policies**: Enable demand-driven scaling based on available resource watermarks,
ensuring sufficient idle capacity while preventing over-provisioning

Fixes https://github.com/openkruise/agents/issues/3

### DIFF
--- a/docs/proposals/20260106-sandboxset-autoscaler.md
+++ b/docs/proposals/20260106-sandboxset-autoscaler.md
@@ -1,0 +1,1274 @@
+---
+title: SandboxSet supports autoscaler
+authors:
+  - "@sivanzcw"
+reviewers:
+  - "@furykerry"
+creation-date: 2026-01-06
+last-updated: 2026-01-15
+status: implementable
+see-also:
+replaces:
+superseded-by:
+---
+
+# SandboxSet supports autoscaler
+
+## Table of Contents
+- [Title](#title)
+    - [Table of Contents](#table-of-contents)
+    - [Summary](#summary)
+    - [Motivation](#motivation)
+        - [Goals](#goals)
+        - [Non-Goals/Future Work](#non-goalsfuture-work)
+    - [Proposal](#proposal)
+        - [User Stories](#user-stories)
+            - [Conversational Agent](#conversational-agent)
+            - [Workflow Agent](#workflow-agent)
+            - [Agent Fan-out](#agent-fan-out)
+            - [RL](#RL)
+        - [Design Details](#design-details)
+            - [API](#API)
+                - [Pool Capacity Control](#pool-capacity-control)
+                - [Pool Scaling Policy](#pool-scaling-policy)
+                    - [Cron-based Policy](#cron-based-policy)
+                    - [Capacity Availability Policy](#capacity-availability-policy)
+                - [Status](#Status)
+            - [Metrics](#Metrics)
+                - [Reconciliation Metrics](#reconciliation-metrics)
+            - [User Configuration Examples](#user-configuration-examples)
+                - [Bounds Enforcer](#bounds-enforcer)
+                - [Cron-Based Scaling](#cron-based-scaling)
+                - [Capacity-Based Scaling with Watermarks](#capacity-based-scaling-with-watermarks)
+                    - [Absolute Value Watermark Configuration](#absolute-value-watermark-configuration)
+                    - [Percentage-Based Watermark Configuration](#percentage-based-watermark-configuration)
+        - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+            - [Policy Combination Limitations](#policy-combination-limitations)
+            - [Observation Window and Sampling Configuration](#observation-window-and-sampling-configuration)
+            - [Cron Policy Maintenance Window Support](#cron-policy-maintenance-window-support)
+            - [One-to-One Relationship Between Warm Pool and Autoscaler](#one-to-one-relationship-between-warm-pool-and-autoscaler)
+        - [Risks and Mitigations](#risks-and-mitigations)
+            - [Controller Computational Complexity and Resource Consumption](#controller-computational-complexity-and-resource-consumption)
+            - [Frequent Scaling Due to Misconfiguration](#frequent-scaling-due-to-misconfiguration)
+            - [Extreme Behavior from Invalid Configuration Combinations](#extreme-behavior-from-invalid-configuration-combinations)
+            - [Observability and Debugging Challenges](#observability-and-debugging-challenges)
+    - [Alternatives](#alternatives)
+        - [Extend Existing HPA for SandboxSet](#extend-existing-hpa-for-sandboxSet)
+        - [Use External Autoscaling Tools](#use-external-autoscaling-tools)
+    - [Upgrade Strategy](#upgrade-strategy)
+        - [API Versioning](#api-versioning)
+        - [Backward Compatibility](#backward-compatibility)
+        - [Upgrade Path](#upgrade-path)
+            - [From No Autoscaler to PoolAutoscaler](#from-no-autoscaler-to-poolAutoscaler)
+            - [Upgrading Autoscaler Configuration](#upgrading-autoscaler-configuration)
+            - [Controller Upgrade](#controller-upgrade)
+        - [Downgrade Strategy](#downgrade-strategy)
+            - [Downgrading Controller Version](#downgrading-controller-version)
+            - [Removing Autoscaler](#removing-autoscaler)
+        - [Version Skew Strategy](#version-skew-strategy)
+            - [API Server and Controller Version Skew](#api-server-and-controller-version-skew)
+    - [Additional Details](#additional-details)
+    - [Test Plan](#test-plan-optional)
+        - [Unit Tests](#unit-tests)
+        - [Integration Tests](#integration-tests)
+        - [Performance Tests](#performance-tests)
+    - [Implementation History](#implementation-history)
+
+## Summary
+
+This enhancement proposes providing autoscaler capabilities for `SandboxSet`,
+enabling intelligent and dynamic management of pre-warmed sandbox resource pools.
+Currently, `SandboxSet` provides sandbox pre-warming capabilities with fixed-size resource pools.
+This enhancement adds two complementary autoscaling policy types:
+
+1. **Cron-based policies**: Enable time-driven scaling based on recurring schedules
+(e.g., scale up during business hours, scale down during off-peak periods)
+2. **Capacity-based policies**: Enable demand-driven scaling based on available resource watermarks,
+ensuring sufficient idle capacity while preventing over-provisioning
+
+The autoscaler integrates seamlessly with `SandboxSet`, providing operators with fine-grained
+control over resource pool capacity while reducing operational overhead. This enhancement addresses
+critical requirements for latency-sensitive agent workloads that require predictable startup performance
+and efficient resource utilization.
+
+## Motivation
+
+Agents are often part of latency-sensitive execution paths and operate under highly dynamic workloads.
+As a result, they have strong requirements around startup performance. By pre-warming sandbox objects
+that provide the agent runtime environment through SandboxSet, agents can achieve near-real-time startup
+and stable startup latency. However, SandboxSet currently supports only a fixed-size pre-warmed resource
+pool, which cannot handle batch agent startup scenarios nor support predictable startup behavior.
+Agents frequently create multiple sandboxes in parallel due to task decomposition or fan-out execution.
+The system must support launching many sandboxes concurrently without linear degradation.
+Startup latency should be correlated with load and capacity rather than exhibiting random jitter.
+Predictability is critical for planners and schedulers that make execution decisions based on expected
+availability.
+
+Enabling autoscaler for `SandboxSet` addresses these challenges by providing:
+
+- **Intelligent and dynamic resource pool management**: Operators can configure intelligent pre-warmed
+pool management policies based on workload concurrency demands. This helps prevent unpredictable
+startup latency, excessive request queuing during peak traffic, and situations where the planner
+is forced into serial execution. The autoscaler automatically adjusts pool capacity based on
+configured policies, ensuring optimal resource availability.
+- **Improved resource utilization**: By dynamically adjusting the capacity of the resource pre-warming pool,
+the cluster can more efficiently handle fluctuations in overall resource usage, thereby reducing
+over-reservation and over-provisioning. The capacity-based policy maintains optimal idle resource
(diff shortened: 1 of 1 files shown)


docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,317 @@
...
+}
+
+// SandboxSetAutoscalerSpec describes the desired functionality of the SandboxSetAutoscaler.
+type SandboxSetAutoscalerSpec struct {
+	// ScaleTargetRef points to the target SandboxSet to scale, and is used to the pods for which instance status
+	// should be collected, as well as to actually change the replica count.
+	// +required
+	ScaleTargetRef CrossVersionObjectReference
+	// maxReplicas is the upper limit for the number of replicas to which the autoscaler can scale up.
+	MaxReplicas int32
```
furykerry, line 233: plz also add MinReplicas, the user may want keep a minimum number of sandbox to ensure the availability of agent application

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,317 @@
...
+	// Example:
+	// "0 8 * * *"     - Every day at 8:00 AM
+	// "0 0 * * 1-5"   - Every weekday at midnight
+	// "0 */2 * * *"   - Every 2 hours
+	// "30 0 * * *"    - Every day at 00:30
+	// "0 0 1 * *"     - First day of every month at midnight
+	// +required
+	Schedule string
+	// TargetReplicas is the desired replicas.
+	TargetReplicas *int32
```
furykerry, line 280: what does freeze window means, does it means the an exception to the schedule ? then what will the autoscaler do during these freeze window

furykerry, line 280: is it possible to modified the min and max replicas in the CronScalingPolicy, so that CronScalingPolicy can be used together with CapacityPolicy

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,317 @@
...
+	// For scaleup:
+	// When the actual available replicas is below this value, the controller will
+	// scale up the SandboxSet to ensure that the number of available instances is
+	// maintained above the value.
+	// For scaledown:
+	// When the actual available replicas is greater than this value, the controller will
+	// scale down the SandboxSet to ensure that the number of available instances is
+	// maintained above the value.
+	// +optional
+	AvailableReplicas int32
```
furykerry: is AvailableReplicas a count or percentage? I think percentage can be more useful for users

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Design Details
+
+### API
+
+```go
+// SandboxSetAutoscaler is the configuration for a SandboxSet
+// autoscaler, which automatically manages the replica count of the SandboxSet
+// based on the policies specified.
+type SandboxSetAutoscaler struct {
```
furykerry: consider rename the scaler to a more neutral name, e.g. PoolAutoScaler. It might be possible to use PoolAutoScaler with other warmpool implementation.

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,317 @@
...
+
+type CapacityScalingRules struct {
+	// StabilizationWindowSeconds is the number of seconds for which past recommendations should be
+	// considered while scaling up or scaling down.
+	// StabilizationWindowSeconds must be greater than or equal to zero and less than or equal to 3600 (one hour).
+	// If not set, use the default values:
+	// - For scale up: 0 (i.e. no stabilization is done).
+	// - For scale down: 300 (i.e. the stabilization window is 300 seconds long).
+	// +optional
+	StabilizationWindowSeconds *int32
```
furykerry, line 310: if StabilizationWindowSeconds can be different for scaling up and down, then what is the time span that the pool utilisation is calculated? consider add another field in CapacityPolicy called  ObserveWindowSeconds

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,317 @@
...
+	ObservedGeneration *int64
+
+	// LastScaleTime is the last time the SandboxSetAutoscaler scaled the number of pods,
+	// used by the autoscaler to control how often the number of pods is changed.
+	// +optional
+	LastScaleTime *metav1.Time
+
+	// CurrentReplicas is current number of replicas of pods managed by this autoscaler,
+	// as last seen by the autoscaler.
+	CurrentReplicas int32
```
furykerry, line 326: shall we add AvailableReplicas so that the user can understand the CapacityScalingRules better?

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: Yes, that is the intended meaning. To handle scenarios such as holidays, a maintenance window is required. Any policies whose schedules intersect with the maintenance window will not be executed.

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: I have considered providing a more flexible definition for the watermark, supporting both absolute values and percentages, as shown below
```
// CapacityValue is a value holder that abstracts literal versus percentage based value
type CapacityValue struct {
	// The following fields are exclusive. Only the topmost non-zero field is used.

	// absolute Value.
	Value int32
	// Percentage of replicas.
	Percentage float32
}
```
For the low watermark configuration, allowing users to specify a percentage can be more appropriate. When the resource pool capacity(including both used and unused resource) is large, a percentage-based configuration can provide a large amount of pre-warmed available capacity, which may better meet expectations.

For the high watermark, it is recommended to configure it as an absolute value. This allows to explicitly control the upper limit of the entire resource pool and have clearer visibility and awareness of the associated costs.

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: Okay, let me take a look at how the capacity defined at the outer level and capacity defined within the policies are reconciled

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: I'll add a table of user scenarios along with example configuration to see how a elegant configuration can satisfy a wider range of user cases

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: Agreed. I'll incorporate this in the next push, thanks

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: Agreed. This is a better approach

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: A minimum replicas configuration is indeed required. When a user configures the upper watermark as a percentage in the capacity policy, a minimum replicas count is needed to clearly define the steady-state number of replicas.

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: adds status fields related to capacity
```
type CapacityStatus struct {
	// AvailableReplicas is current number of available pods managed by this autoscaler,
	// as last seen by the autoscaler.
	AvailableReplicas int32
}
```

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: The scaling decision can be determined based on recommended values stored in memory over a recent time window. For scale-up, it selects the minimum recommendation from the window. For scale-down, it selects the maximum recommendation from the window

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,1300 @@
...
+	// - For scale up: 0 (i.e. no stabilization is done).
+	// - For scale down: 300 (i.e. the stabilization window is 300 seconds long).
+	// +optional
+	StabilizationWindowSeconds *int32
+	// Tolerance is the tolerance between the watermark and desired
+	// value under which no updates are made to the desired number of
+	// replicas (e.g. 0.01 for 1%). Must be greater than or equal to zero. If not
+	// set, the default cluster-wide tolerance is applied (by default 10%).
+	// +optional
+	Tolerance *CapacityValue
```
furykerry: consider move Tolerance to CapacityPolicy

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,1300 @@
...
+	// as last seen by the autoscaler.
+	CurrentReplicas int32
+
+	// DesiredReplicas is the desired number of replicas of pods managed by this autoscaler,
+	// as last calculated by the autoscaler.
+	DesiredReplicas int32
+	// This flag tells the controller to suspend subsequent executions
+	// Defaults to false.
+	// +optional
+	Suspend bool
```
furykerry: plz rename Suspend to Suspended

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,1300 @@
...
+
+**Use cases**:
+- Monitor reconciliation performance
+- Identify slow reconciliations
+- Track performance degradation
+- Set SLOs for reconciliation latency
+
+##### Webhook Metrics
+
+`object_validations_total`
```
furykerry: webhook metric is redundant, controller-runtime already expose webhook validation counter for various return code

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,1300 @@
...
+	// TargetReplicas is the desired replicas.
+	// +required
+	TargetReplicas *int32
+}
+
+// CapacityPolicy defines the capacity configuration of the target resource pool.
+type CapacityPolicy struct {
+	// TargetAvailable is the desired available replicas.
+	// +required
+	TargetAvailable CapacityValue
```
furykerry: how about change type from CapacityValue to IntOrString ?

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,1300 @@
...
+	// +optional
+	LastScheduleTime *metav1.Time
+}
+
+type CapacityStatus struct {
+	// Available is current number of available pods managed by this autoscaler,
+	// as last seen by the autoscaler.
+	Available int32
+	// The available resource pool capacity range determined by the CapacityPolicy.
+	AvailableWatermark CapacityWatermark
```
furykerry: it seems redundant since AvailableWatermark can be calculated from the spec

sivanzcw: Thanks! Removed as suggested.

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: Agreed. Updated to use the standard kubernetes api pattern.

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: I've removed it . Thanks for pointing this out.

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: Updated the field.

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: This's a good suggestion. I've moved the API definition accordingly.

sivanzcw:
changes have been addressed, PTAL cc @furykerry

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,1228 @@
...
+When all watermarks (target, scale-up tolerance, and scale-down tolerance)
+are configured as percentages, they scale proportionally together,
+ensuring uniform scaling behavior regardless of pool scale.
+
+**Configuration Example**:
+```yaml
+spec:
+  capacityPolicy:
+    targetAvailable: 70  # Target: maintain 70% of replicas as idle
+    tolerance: 10  # Scale up when available < 70% - 10% = 60% of replicas. Scale down when available > 70% + 10% = 80% of replicas
```
furykerry: targetAvailable should be 70%, and tolerance should be 10%

docs/proposals/20260106-sandboxset-autoscaler.md
```diff
@@ -0,0 +1,1228 @@
...
+Currently, the autoscaler does not support simultaneous configuration of cron-based policies and
+capacity-based policies. This limitation exists to avoid potential conflicts and ensure
+predictable scaling behavior. The autoscaler will prioritize one policy type when both are configured,
+which may lead to unexpected scaling results.
+
+*Future Enhancement*: Support for combining cron-based and capacity-based policies will be added
+based on product requirements and user feedback. The implementation will include conflict
+resolution mechanisms and clear precedence rules to ensure consistent and predictable autoscaling behavior.
+
+#### Autoscaler Synchronization Delay
```
furykerry: consider rename Synchronization Delay to observe window, and there should be another related config to control the minimum sampling time. plz also add recommendation for default values

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: fixed

docs/proposals/20260106-sandboxset-autoscaler.md
sivanzcw: thanks for the clarification, done as described.

PR #78, opened 2026-01-09
feat: introduce global cache for GetSandboxState

### LINKED ISSUE #36: [feature request] Add cache for stateutils.GetSandboxState
**What would you like to be added**:

The function `stateutils.GetSandboxState` will be called very frequently. Adding a global cache (key+resourceVersion) could significantly improve performance. Considering that memory usage would not be large, this cache can be retained and updated continuously, and only destroyed when a sandbox delete event is detected.

**Why is this needed**:

DCchoudhury15:
**What this PR does / why we need it**:
This PR implements a thread-safe global cache for `stateutils.GetSandboxState` to reduce performance overhead, as identified in #36.

**Which issue(s) this PR fixes**:
Fixes #36

**Special notes for your reviewer**:
- Implemented a `sync.RWMutex` protected map in `pkg/utils/sandboxutils`.
- Added invalidation logic in `SandboxController` (clears cache on resource deletion).
- Added safeguards to bypass the cache when `ResourceVersion` or `UID` is empty, or when running unit tests (`flag.Lookup("test.v")`), ensuring no side effects on existing tests.


AiRanthem:
Thanks @DCchoudhury15 . Could you confirm that it actually improves performance through benchmarking?

DCchoudhury15:
Fair point @furykerry.

I'll run a profile under load and grab a pprof dump to verify if this is actually a bottleneck before i commit to the complexity. Will update with the results in a few hours.

PR #114, opened 2026-01-28
feat: implement SandboxClaim CRD for sandbox resource claiming

PersistentJZH:

SandboxClaim enables users to claim sandbox resources from a SandboxSet pool in a Kubernetes-native way. It implements a declarative, one-time claim operation with automatic lifecycle management and TTL-based cleanup.

Changes include:

- CRD Definition
- Controller Logic
- Webhook
- E2E Tests

https://github.com/openkruise/agents/issues/30

can see e2e cases

can first understand the behavior of CRD field by looking at the e2e cases, and then review the specific logic.

### DIFF
--- a/pkg/controller/sandboxclaim/core/common_control.go
+++ b/pkg/controller/sandboxclaim/core/common_control.go
@@ -87,7 +87,7 @@ func (c *commonControl) EnsureClaimClaiming(ctx context.Context, args ClaimArgs)
 
 	// Step 4: Use max(statusCount, actualCount) to get current count
 	currentCount := statusCount
-	if actualCount > statusCount {
+	if actualCount > currentCount {
 		log.Info("Status count mismatch, using actual count",
 			"statusCount", statusCount,
 			"actualCount", actualCount)
--- a/pkg/controller/sandboxclaim/core/util.go
+++ b/pkg/controller/sandboxclaim/core/util.go
@@ -36,11 +36,11 @@ import (
 //   - skipBusinessLogic: true if should skip Ensure* methods and go directly to status update
 //
 // Handled scenarios (in order):
-//  1. New claim (Phase == "")               → Claiming, continue
-//  2. Already Completed                     → Completed, continue (for TTL cleanup)
-//  3. SandboxSet not found                  → Completed, SKIP (terminal)
-//  4. Timeout exceeded                      → Completed, SKIP (terminal)
-//  5. All replicas claimed                  → Completed, SKIP (terminal)
+//  1. Already Completed                     → Completed, continue (for TTL cleanup)
+//  2. SandboxSet not found                  → Completed, SKIP (terminal, fail-fast)
+//  3. New claim (Phase == "")               → Claiming, continue
+//  4. All replicas claimed                  → Completed, SKIP (terminal)
+//  5. Timeout exceeded                      → Completed, SKIP (terminal)
 //  6. Otherwise                             → Current phase, continue
 //
 // Note: ObservedGeneration is always updated to track spec changes
@@ -51,7 +51,29 @@ func CalculateClaimStatus(args ClaimArgs) (*agentsv1alpha1.SandboxClaimStatus, b
 	// Always update ObservedGeneration to track spec changes
 	newStatus.ObservedGeneration = claim.Generation
 
-	// 1. Handle initial state - directly transition to Claiming phase
+	// 1. Handle terminal state
+	if newStatus.Phase == agentsv1alpha1.SandboxClaimPhaseCompleted {
+		klog.V(2).InfoS("SandboxClaim already completed, skipping state calculation",
+			"claim", klog.KObj(claim),
+			"completionTime", newStatus.CompletionTime)
+		// If already Completed, skip state calculation but allow EnsureClaimCompleted to run
+		// (for TTL cleanup logic)
+		return newStatus, false
+	}
+
+	// 2. Check if SandboxSet exists
+	// Transition: * → Completed (SandboxSet deleted)
+	if args.SandboxSet == nil {
+		klog.InfoS("SandboxSet not found, transitioning to Completed",
+			"claim", klog.KObj(claim),
+			"sandboxSet", claim.Spec.TemplateName)
+		return TransitionToCompleted(newStatus,
+			"SandboxSetNotFound",
+			"SandboxSet not found or deleted"), true
+	}
+
+	// 3. Handle initial state
+	// Transition: "" → Claiming
 	if newStatus.Phase == "" {
 		klog.InfoS("Initializing new SandboxClaim, starting claim process",
 			"claim", klog.KObj(claim),
@@ -63,28 +85,17 @@ func CalculateClaimStatus(args ClaimArgs) (*agentsv1alpha1.SandboxClaimStatus, b
 		return newStatus, false
 	}
 
-	// 2. Handle terminal state
-	if newStatus.Phase == agentsv1alpha1.SandboxClaimPhaseCompleted {
-		klog.V(2).InfoS("SandboxClaim already completed, skipping state calculation",
+	// 4. Check if desired replicas already met
+	// Transition: Claiming → Completed (All replicas claimed)
+	if isReplicasMet(claim, newStatus) {
+		klog.InfoS("All replicas claimed, transitioning to Completed",
 			"claim", klog.KObj(claim),
-			"completionTime", newStatus.CompletionTime)
-		// If already Completed, skip state calculation but allow EnsureClaimCompleted to run
-		// (for TTL cleanup logic)
-		return newStatus, false
+			"claimedReplicas", newStatus.ClaimedReplicas,
+			"desiredReplicas", getDesiredReplicas(claim))
+		return transitionToCompletedWithSuccess(newStatus, claim), true
 	}
 
-	// 3. Check if SandboxSet still exists
-	// Transition: * → Completed (SandboxSet deleted)
-	if args.SandboxSet == nil {
-		klog.InfoS("SandboxSet not found, transitioning to Completed",
-			"claim", klog.KObj(claim),
-			"sandboxSet", claim.Spec.TemplateName)
-		return transitionToCompleted(newStatus,
-			"SandboxSetNotFound",
-			"SandboxSet not found or deleted"), true
-	}
-
-	// 4. Early timeout detection
+	// 5. Early timeout detection
 	// Transition: Claiming → Completed (Timeout)
 	if isClaimTimeout(claim, newStatus) {
 		elapsed := time.Since(newStatus.ClaimStartTime.Time)
@@ -95,16 +106,6 @@ func CalculateClaimStatus(args ClaimArgs) (*agentsv1alpha1.SandboxClaimStatus, b
 			"claimedReplicas", newStatus.ClaimedReplicas,
 			"desiredReplicas", getDesiredReplicas(claim))
 		return transitionToCompletedWithTimeout(newStatus, elapsed, claim), true
-	}
-
-	// 5. Check if desired replicas already met
-	// Transition: Claiming → Completed (All replicas claimed)
-	if isReplicasMet(claim, newStatus) {
-		klog.InfoS("All replicas claimed, transitioning to Completed",
-			"claim", klog.KObj(claim),
-			"claimedReplicas", newStatus.ClaimedReplicas,
-			"desiredReplicas", getDesiredReplicas(claim))
-		return transitionToCompletedWithSuccess(newStatus, claim), true
 	}
 
 	// Continue with business logic
@@ -142,8 +143,8 @@ func isReplicasMet(claim *agentsv1alpha1.SandboxClaim, status *agentsv1alpha1.Sa
 	return status.ClaimedReplicas >= getDesiredReplicas(claim)
 }
 
-// transitionToCompleted transitions the claim to Completed state with a generic reason
-func transitionToCompleted(status *agentsv1alpha1.SandboxClaimStatus, reason, message string) *agentsv1alpha1.SandboxClaimStatus {
+// TransitionToCompleted transitions the claim to Completed state with a generic reason
+func TransitionToCompleted(status *agentsv1alpha1.SandboxClaimStatus, reason, message string) *agentsv1alpha1.SandboxClaimStatus {
 	status.Phase = agentsv1alpha1.SandboxClaimPhaseCompleted
 	status.Message = message
 	now := metav1.Now()
@@ -156,7 +157,7 @@ func transitionToCompleted(status *agentsv1alpha1.SandboxClaimStatus, reason, me
 		Message:            message,
 		LastTransitionTime: now,
 	}
-	setCondition(&status.Conditions, condition)
+	SetClaimCondition(status, condition)
 
 	return status
 }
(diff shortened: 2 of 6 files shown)


api/v1alpha1/sandboxclaim_types.go
```diff
@@ -0,0 +1,170 @@
...
+
+	// ShutdownTime specifies the absolute time when the sandbox should be shut down
+	// This will be set as spec.shutdownTime (absolute time) on the Sandbox
+	// +optional
+	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
+
+	// ClaimTimeout specifies the maximum duration to wait for claiming sandboxes
+	// If the timeout is reached, the claim will be marked as Completed regardless of
+	// whether all replicas were successfully claimed
+	// +optional
```
zmberg, line 46: Is there a default value? It seems more reasonable to have one.

pkg/controller/sandboxclaim/sandboxclaim_controller.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	// Execute business logic and get requeue strategy
+	var strategy core.RequeueStrategy
+	var err error
+
+	// State-driven execution - each Ensure method returns its own requeue strategy
+	switch newStatus.Phase {
+	case agentsv1alpha1.SandboxClaimPhasePending:
+		strategy, err = r.getControl().EnsureClaimPending(ctx, args)
```
furykerry: is phase SandboxClaimPhasePending necessary, it seems that we can just fall through and goes to the claiming phase.

pkg/controller/sandboxclaim/core/common_control.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// countClaimedSandboxes counts sandboxes that are claimed by this claim
+func (c *commonControl) countClaimedSandboxes(ctx context.Context, claim *agentsv1alpha1.SandboxClaim) (int32, error) {
+	sandboxList := &agentsv1alpha1.SandboxList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(claim.Namespace),
+		client.MatchingLabels{
+			agentsv1alpha1.LabelSandboxClaimedBy: string(claim.UID),
+		},
```
furykerry: plz add UnsafeDisableDeepCopy option to avoid unnecessary deepcopy

pkg/controller/sandboxclaim/core/common_control.go
```diff
@@ -0,0 +1,304 @@
...
+	return sandboxcr.ValidateAndInitClaimOptions(opts)
+}
+
+// countClaimedSandboxes counts sandboxes that are claimed by this claim
+func (c *commonControl) countClaimedSandboxes(ctx context.Context, claim *agentsv1alpha1.SandboxClaim) (int32, error) {
+	sandboxList := &agentsv1alpha1.SandboxList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(claim.Namespace),
+		client.MatchingLabels{
+			agentsv1alpha1.LabelSandboxClaimedBy: string(claim.UID),
```
furykerry: plz add an index named IndexTemplateAvailable, and use the IndexTemplateAvailable to list the sandboxes.

pkg/controller/sandboxclaim/core/common_control.go
```diff
@@ -0,0 +1,304 @@
...
+
+			// 1. apply labels
+			labels := sbx.GetLabels()
+			if labels == nil {
+				labels = make(map[string]string)
+			}
+			// Set UID for uniqueness (prevents conflict when claim is recreated with same name)
+			labels[agentsv1alpha1.LabelSandboxClaimedBy] = string(claim.UID)
+			// Set name for readability
+			labels[agentsv1alpha1.LabelSandboxClaimedByName] = claim.Name
```
furykerry: can we just use AnnotationOwner to filter the sandbox?

pkg/controller/sandboxclaim/core/common_control.go
```diff
@@ -0,0 +1,304 @@
...
+	// - Status update fails due to network issues
+	actualCount, err := c.countClaimedSandboxes(ctx, claim)
+	if err != nil {
+		return NoRequeue(), fmt.Errorf("failed to count claimed sandboxes: %w", err)
+	}
+
+	// Step 4: Use max(statusCount, actualCount) to get current count
+	currentCount := statusCount
+	if actualCount > statusCount {
+		log.Info("Status count mismatch, using actual count",
```
furykerry, line 91: we think we should record the claimed sandbox name in the claim status after all. If the claimed sandbox is  deleted immediately  by the user, sandboxclaim may actually claim more sandboxes than expected

pkg/webhook/sandboxclaim/validating/sandboxclaim_validating.go
```diff
@@ -0,0 +1,121 @@
...
+	if oldClaim.Spec.Replicas != nil {
+		oldReplicas = *oldClaim.Spec.Replicas
+	}
+
+	newReplicas := int32(1)
+	if newClaim.Spec.Replicas != nil {
+		newReplicas = *newClaim.Spec.Replicas
+	}
+
+	if oldReplicas != newReplicas {
```
furykerry: can we avoid webhook for sandboxclaim (which is the critical path of sandbox allocation) and use kubebuilder marker to do the validation,e.g. 

+kubebuilder:validation:Immutable

pkg/controller/sandboxclaim/sandboxclaim_controller.go
PersistentJZH: yes, we do not need SandboxClaimPhasePending phase, can enter claiming phase directly.

pkg/controller/sandboxclaim/core/common_control.go
PersistentJZH: good advice

pkg/controller/sandboxclaim/core/common_control.go
PersistentJZH: @furykerry here we should list all sandboxes, including those not available, and not list by template (but by labels ). so it be sufficient to simply create an label IndexField?

pkg/controller/sandboxclaim/core/common_control.go
PersistentJZH: yes, that would make it more consistent

pkg/webhook/sandboxclaim/validating/sandboxclaim_validating.go
PersistentJZH: @furykerry Currently, it seems to only be able to achieve field immutability through CEL expressions in kubebuilder, but this depends on the Kubernetes version (> 1.25). For older versions of Kubernetes, this validate may fail.

ref: https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/

pkg/controller/sandboxclaim/core/common_control.go
PersistentJZH: @furykerry if considering performance, the label seems to be better.

furykerry: since we have a index for AnnotationOwner, the performance is good enough

PersistentJZH: ok, noticed that, will change to use AnnotationOwner

api/v1alpha1/sandboxclaim_types.go
PersistentJZH: @zmberg it seems difficult to determine a default value because the claim time is related to the number of replicas. However, setting a relatively long time, such as 5 minutes, seems like a good option?

furykerry: what about 1 minute, 5 minutes sounds too long for me

pkg/controller/sandboxclaim/sandboxclaim_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	k8sinformers "k8s.io/client-go/informers"
+)
+
+func init() {
+	flag.IntVar(&concurrentReconciles, "sandboxclaim-workers", concurrentReconciles, "Max concurrent workers for SandboxClaim controller.")
+	flag.IntVar(&maxClaimBatchSize, "sandboxclaim-max-batch-size", maxClaimBatchSize, "Maximum batch size for claiming sandboxes in a single reconcile cycle")
+}
+
+var (
+	concurrentReconciles = 10
```
furykerry: consider increase the default concurrentReconciles to 500 , so as to match the worker number of sandbox controller

api/v1alpha1/sandboxclaim_types.go
PersistentJZH: ok, updated

pkg/controller/sandboxclaim/sandboxclaim_controller.go
PersistentJZH: updated

pkg/controller/sandboxclaim/sandboxclaim_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	logger.Info("Began to process SandboxClaim for reconcile")
+
+	// Check resourceVersion expectations
+	core.ResourceVersionExpectations.Observe(claim)
+	if isSatisfied, unsatisfiedDuration := core.ResourceVersionExpectations.IsSatisfied(claim); !isSatisfied {
+		if unsatisfiedDuration < expectations.ExpectationTimeout {
+			logger.Info("Not satisfied resourceVersion for SandboxClaim, wait for cache event")
+			return reconcile.Result{RequeueAfter: expectations.ExpectationTimeout - unsatisfiedDuration}, nil
+		}
+		klog.InfoS("Expectation unsatisfied overtime for SandboxClaim, wait for cache event timeout", "timeout", unsatisfiedDuration)
```
zmberg: use logger

pkg/controller/sandboxclaim/core/util.go
```diff
@@ -0,0 +1,240 @@
...
+			"claim", klog.KObj(claim),
+			"completionTime", newStatus.CompletionTime)
+		// If already Completed, skip state calculation but allow EnsureClaimCompleted to run
+		// (for TTL cleanup logic)
+		return newStatus, false
+	}
+
+	// 3. Check if SandboxSet still exists
+	// Transition: * → Completed (SandboxSet deleted)
+	if args.SandboxSet == nil {
```
zmberg, line 66: Move to the very beginning of this function. If can't find it there, then it must be at the Completed.

pkg/controller/sandboxclaim/core/util.go
```diff
@@ -0,0 +1,240 @@
...
+		klog.InfoS("Claim timeout reached, transitioning to Completed",
+			"claim", klog.KObj(claim),
+			"timeout", claim.Spec.ClaimTimeout.Duration,
+			"elapsed", elapsed,
+			"claimedReplicas", newStatus.ClaimedReplicas,
+			"desiredReplicas", getDesiredReplicas(claim))
+		return transitionToCompletedWithTimeout(newStatus, elapsed, claim), true
+	}
+
+	// 5. Check if desired replicas already met
```
zmberg: isReplicasMet should before isClaimTimeout

pkg/controller/sandboxclaim/core/util.go
```diff
@@ -0,0 +1,240 @@
...
+		Message:            fmt.Sprintf("Successfully claimed all %d sandboxes", status.ClaimedReplicas),
+		LastTransitionTime: now,
+	}
+	setCondition(&status.Conditions, condition)
+
+	return status
+}
+
+// setCondition sets or updates a condition in the condition list
+func setCondition(conditions *[]metav1.Condition, newCondition metav1.Condition) {
```
zmberg: reference https://github.com/openkruise/agents/blob/master/pkg/utils/utils.go#L22.

pkg/controller/sandboxclaim/core/common_control.go
```diff
@@ -0,0 +1,284 @@
...
+// EnsureClaimClaiming handles the logic for claiming sandboxes
+func (c *commonControl) EnsureClaimClaiming(ctx context.Context, args ClaimArgs) (RequeueStrategy, error) {
+	log := logf.FromContext(ctx)
+	claim, sandboxSet := args.Claim, args.SandboxSet
+
+	// Step 1: Get desired replicas
+	desiredReplicas := getDesiredReplicas(claim)
+
+	// Step 2: Get current count from status
+	statusCount := claim.Status.ClaimedReplicas
```
zmberg, line 70: Should the newStatus variable be used in all methods within this file?

pkg/controller/sandboxclaim/core/common_control.go
```diff
@@ -0,0 +1,284 @@
...
+	//   Then the controller will create new sandboxes to reach the desired replicas,
+	//   even though the user intentionally deleted them, it's an extremely rare case.
+	actualCount, err := c.countClaimedSandboxes(ctx, claim)
+	if err != nil {
+		return NoRequeue(), fmt.Errorf("failed to count claimed sandboxes: %w", err)
+	}
+
+	// Step 4: Use max(statusCount, actualCount) to get current count
+	currentCount := statusCount
+	if actualCount > statusCount {
```
zmberg: if actualCount > currentCount {

pkg/controller/sandboxclaim/core/common_control.go
```diff
@@ -0,0 +1,284 @@
...
+		log.Info("Status count mismatch, using actual count",
+			"statusCount", statusCount,
+			"actualCount", actualCount)
+		currentCount = actualCount
+	}
+
+	// Step 5: Update status with current count
+	args.NewStatus.ClaimedReplicas = currentCount
```
zmberg, line 98: Does this make sense? Shouldn't they be equal? Or do you want to use `actualCount`?

pkg/controller/sandboxclaim/sandboxclaim_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	k8sinformers "k8s.io/client-go/informers"
+)
+
+func init() {
+	flag.IntVar(&concurrentReconciles, "sandboxclaim-workers", concurrentReconciles, "Max concurrent workers for SandboxClaim controller.")
+	flag.IntVar(&maxClaimBatchSize, "sandboxclaim-max-batch-size", maxClaimBatchSize, "Maximum batch size for claiming sandboxes in a single reconcile cycle")
+}
+
+var (
+	concurrentReconciles = 500
```
zmberg, line 64: move concurrentReconciles to 50.

pkg/controller/sandboxclaim/sandboxclaim_controller_test.go
```diff
@@ -0,0 +1,402 @@
...
+	updatedClaim := &agentsv1alpha1.SandboxClaim{}
+	err = fakeClient.Get(context.Background(),
+		types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace},
+		updatedClaim)
+
+	if err != nil {
+		t.Fatalf("Failed to get updated claim: %v", err)
+	}
+
+	if updatedClaim.Status.Phase != agentsv1alpha1.SandboxClaimPhaseClaiming {
```
furykerry, line 305: plz add assertion against sandbox modification, e.g.  annotation and labels

pkg/controller/sandboxclaim/core/common_control.go
```diff
@@ -0,0 +1,284 @@
...
+	//   Then the controller will create new sandboxes to reach the desired replicas,
+	//   even though the user intentionally deleted them, it's an extremely rare case.
+	actualCount, err := c.countClaimedSandboxes(ctx, claim)
+	if err != nil {
+		return NoRequeue(), fmt.Errorf("failed to count claimed sandboxes: %w", err)
+	}
+
+	// Step 4: Use max(statusCount, actualCount) to get current count
+	currentCount := statusCount
+	if actualCount > currentCount {
```
furykerry, line 90: plz add the ut to cover the case of actualCount > currentCount

PR #141, opened 2026-02-06
add sandboxtemplate

furykerry:
add CR for sandboxtemplate and support sandbox creation using templateRef. 

1. sandboxtemplate creation should works
2. sandbox creation using templateRef should works

it is the first patch to implement [proposal](https://github.com/openkruise/agents/blob/master/docs/proposals/20260123-sandbox-template.md)


config/webhook/manifests.yaml
```diff
@@ -25,6 +25,27 @@ webhooks:
...
+      path: /default-sandboxtemplate
+  failurePolicy: Fail
+  name: md-sbt.kb.io
+  rules:
+  - apiGroups:
+    - agents.kruise.io
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
```
zmberg: Only Create?

furykerry, line 44: in the openapi schema, i have add validation that the template cannot be changed, so update webhook is not necessary

PR #167, opened 2026-02-28
sanboxset support scale maxUnavailable

zmberg:


pkg/controller/sandboxset/sandboxset_controller.go
```diff
@@ -146,10 +148,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	var allErrors error
-
 	// Step 1: perform scale
 	start := time.Now()
-	delta := int(sbs.Spec.Replicas - actualReplicas)
+	delta := calculateScaleDelta(sbs, actualReplicas)
```
furykerry: actualReplicas will includes the creating sandboxes, maybe we should only count the available sandboxes

PR #172, opened 2026-03-04
feature(sandbox-manager): add rate limiter to all apis and introduce a better limiter framework

AiRanthem:

### DIFF
--- a/cmd/sandbox-manager/main.go
+++ b/cmd/sandbox-manager/main.go
@@ -41,6 +41,7 @@ func main() {
 	var extProcMaxConcurrency int
 	var kubeClientQPS float64
 	var kubeClientBurst int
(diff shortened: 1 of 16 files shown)


cmd/sandbox-manager/main.go
```diff
@@ -47,6 +48,7 @@ func main() {
 	pflag.StringVar(&pprofAddr, "pprof-addr", ":6060", "The address the pprof debug maps to.")
 
 	// Register server configuration flags
+	pflag.IntVar(&basicQPS, "basic-qps", models.DefaultBasicQPS, "Basic QPS for the apis writing K8S APIServer")
```
zmberg: Is this description correct? I see it refers to the route's rate limiting.

PR #189, opened 2026-03-11
fix(sandbox): update some logic about timeout and pause/resume

AiRanthem:
fix(sandbox): update some logic about timeout and pause/resume

- add a new E2B extension: e2b.agents.kruise.io/never-timeout
- change pause to a sync method
- fixed a bug where some sandbox routes might be lost upon restart


furykerry (approved):
/lgtm
/approve

PR #275, opened 2026-04-15
Feat/checkpoint bugfix

BH4AWS:


AiRanthem (commented):
/lgtm

AiRanthem:
There's a small issue found by Qoder:

### Issue #1: `propagateAnnotationsToTemplate` may unnecessarily mutate template annotations

| Attribute | Details |
|-----------|---------|
| **Location** | `pkg/sandbox-manager/infra/sandboxcr/utils.go:80-86` |
| **Category** | Logic / Safety |

**Description:**

`propagateAnnotationsToTemplate` calls `sbx.GetAnnotations()` which can return `nil`. While Go allows indexing a `nil` map (returns zero value, no panic), the function always calls `template.SetAnnotations(tmplAnnotations)` even when no necessary keys were found, replacing the existing `nil` with an empty map. For `CreateCheckpoint`, this creates a template with an empty annotations map instead of `nil`, which is benign but slightly wasteful.

**Fix:**

Consider returning early if no necessary keys are found in sandbox annotations:

```go
if sbxAnnotations == nil {
    return
}

File: pkg/sandbox-manager/infra/sandboxcr/clone.go [hunk 1]
```diff
@@ -318,6 +354,8 @@ func CreateCheckpoint(ctx context.Context, sbx *v1alpha1.Sandbox, client clients
 			}
 		}
 	}
+	// to make sure the template annotations are propagated to the template
```
furykerry: consider name template to checkpoint

pkg/sandbox-manager/infra/sandboxcr/utils.go
```diff
@@ -69,3 +75,40 @@ func getCsiMountExtensionRequest(s metav1.Object) ([]v1alpha1.CSIMountConfig, er
 	}
 	return csiMountRequests, nil
 }
+
+func propagateAnnotationsToCheckpoint(sbx *v1alpha1.Sandbox, cp *v1alpha1.Checkpoint) {
```
furykerry: plz move propagateAnnotationsToCheckpoint and restoreAnnotationsFromCheckpoint to pkg/utils/checkpoint, so as to avoid cyclic dependency problem ( sandbox controller depends sandbox manager)

BH4AWS: done   /utils/checkpoint/utils.go

pkg/sandbox-manager/infra/sandboxcr/clone.go
BH4AWS: done

furykerry (approved):
/lgtm
/approve

PR #278, opened 2026-04-15
feat: supoort kruise custom api for gateway 

chengzhycn:
Now let me also check the earlier part of the diff that was truncated:
Here's the generated PR description:

---

This PR adds support for the Kruise custom API protocol in the sandbox-gateway, refactoring the filter to use a unified adapter pattern for sandbox routing.

**Gateway filter refactoring (`pkg/sandbox-gateway/filter/`)**
- Replaced the inline header/host extraction logic in `DecodeHeaders` with the unified `E2BMapper` adapter interface from `pkg/servers/e2b/adapters/`
- Introduced `FilterConfig` wrapper that holds both `Config` and `Adapter`, created via `NewFilterConfig()` / `NewE2BAdapterWithOptions()`
- Removed `ExtractHostInfo()` from `Config` — host parsing is now delegated to `NativeE2BAdapter`
- Port is now `int` throughout the filter instead of `string`
- Added support for extra headers from the adapter (e.g., `:path` rewrite for kruise custom protocol)

**NativeE2BAdapter enhancements (`pkg/servers/e2b/adapters/native_e2b.go`)**
- Extended `NativeE2BAdapter` with configurable fields: `SandboxIDHeader`, `SandboxPortHeader`, `HostHeader`, `DefaultPort`
- Implemented header-first, hostname-fallback extraction strategy with custom header support
- Added `NewE2BAdapterWithOptions()` constructor for the gateway use case

**Envoy configuration (`config/sandbox-gateway/configmap.yaml`)**
- Added `/kruise/api` route prefix that forwards to a new `manager_cluster` (sandbox-manager service on port 7788)

**Deployment cleanup (`config/sandbox-gateway/deployment.yaml`)**
- Removed the `init-sysctl` init container that required `SYS_ADMIN` capabilities

**CI & E2E (`hack/`, `.github/workflows/`, `test/e2b/`)**
- Added sandbox-gateway image build, load, and deploy steps to both E2B e2e workflows
- Updated `run-e2b-e2e-test.sh` to wait for gateway pods and port-forward through gateway instead of manager directly
- Added gateway restart count check in e2e teardown
- Added new `test_gateway.py` with header-based and host-based routing e2e tests

**Makefile**
- Added `deploy-sandbox-gateway` and `undeploy-sandbox-gateway` targets

NONE

1. **Unit tests** — run the gateway filter and adapter tests:
   ```bash
   go test ./pkg/sandbox-gateway/filter/... -v
   go test ./pkg/servers/e2b/adapters/... -v
   ```
2. **E2E tests** — the CI workflows now build and deploy sandbox-gateway; the new `test/e2b/test_gateway.py` validates both header-based and host-based routing through the gateway.
3. **Manual verification** — deploy sandbox-gateway with the updated Envoy configmap, then:
   - Send a request with `e2b-sandbox-id` and `e2b-sandbox-port` headers → should route to the correct sandbox pod
   - Send a request with host format `{port}-{ns}--{name}.example.com` → should route via hostname fallback
   - Send a request to `/kruise/ns--sandbox/3000/api/v1/data` → should route via kruise custom protocol with path rewrite

- The `init-sysctl` init container with `SYS_ADMIN` capability was removed from the gateway deployment — verify this is acceptable for your environment.
- The `NativeE2BAdapter.Map()` signature is unchanged but the behavior now supports header-first extraction. Existing callers using only hostname-based routing are unaffected (hostname fallback still works).
- The `ConfigParser.Merge` and `FilterFactory` now operate on `*FilterConfig` instead of `*Config` — this is a breaking change for the internal Envoy Go filter API surface but not for external consumers.

### DIFF
--- a/pkg/sandbox-gateway/filter/config.go
+++ b/pkg/sandbox-gateway/filter/config.go
@@ -91,10 +91,12 @@ type FilterConfig struct {
 func NewFilterConfig(cfg *Config) *FilterConfig {
 	adapter := adapters.NewE2BAdapterWithOptions(
 		0, // port not used by gateway
-		cfg.GetSandboxHeaderName(),
-		cfg.GetSandboxPortHeader(),
-		cfg.GetHostHeaderName(),
-		cfg.GetDefaultPort(),
+		adapters.E2BAdapterOptions{
+			SandboxIDHeader:   cfg.GetSandboxHeaderName(),
+			SandboxPortHeader: cfg.GetSandboxPortHeader(),
+			HostHeader:        cfg.GetHostHeaderName(),
+			DefaultPort:       cfg.GetDefaultPort(),
+		},
 	)
 	return &FilterConfig{
 		Config:  cfg,
--- a/pkg/sandbox-gateway/filter/config.go
+++ b/pkg/sandbox-gateway/filter/config.go
@@ -84,7 +84,7 @@ func (c *Config) GetDefaultPort() int {
 // FilterConfig wraps Config and holds the adapter created from the config
 type FilterConfig struct {
 	*Config
-	Adapter adapters.E2BMapper
+	Adapter *adapters.E2BAdapter
 }
 
 // NewFilterConfig creates a FilterConfig with an adapter built from the config values
--- a/pkg/sandbox-gateway/filter/config.go
+++ b/pkg/sandbox-gateway/filter/config.go
@@ -3,16 +3,14 @@ package filter
 import (
 	"encoding/json"
 	"fmt"
-	"regexp"
+	"strconv"
 
 	v3 "github.com/cncf/xds/go/xds/type/v3"
 	"github.com/envoyproxy/envoy/contrib/golang/common/go/api"
 	"google.golang.org/protobuf/types/known/anypb"
-)
 
-// hostPattern matches the host format: {port}-{namespace}--{name}.{domain}
-// Group 1: port (digits), Group 2: namespace--name (alphanumeric and hyphens)
-var hostPattern = regexp.MustCompile(`^(\d+)-([a-zA-Z0-9\-]+)\.`)
+	"github.com/openkruise/agents/pkg/servers/e2b/adapters"
+)
 
 const (
 	DefaultHostHeaderName    = "Host"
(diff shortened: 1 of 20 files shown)


pkg/servers/e2b/adapters/unified_e2b.go
```diff
@@ -28,6 +28,21 @@ func NewE2BAdapter(port int) *E2BAdapter {
 	}
 }
 
+// NewE2BAdapterWithOptions creates an E2BAdapter with configurable NativeE2BAdapter options.
+// This allows the sandbox-gateway to configure custom header names and default port.
+func NewE2BAdapterWithOptions(port int, sandboxIDHeader, sandboxPortHeader, hostHeader string, defaultPort int) *E2BAdapter {
```
AiRanthem: Add a E2BAdapterOptions struct type and use it as the param

pkg/servers/e2b/adapters/native_e2b.go
```diff
@@ -7,25 +7,98 @@ import (
...
+
+func (a *NativeE2BAdapter) getSandboxPortHeader() string {
+	if a.SandboxPortHeader != "" {
+		return a.SandboxPortHeader
+	}
+	return DefaultSandboxPortHeader
+}
+
+func (a *NativeE2BAdapter) getHostHeader() string {
+	if a.HostHeader != "" {
```
AiRanthem, line 46: Shall we also read `:authority`?

pkg/sandbox-gateway/filter/config.go
```diff
@@ -117,16 +138,16 @@ func (p *ConfigParser) Parse(any *anypb.Any, callbacks api.ConfigCallbackHandler
...
 
-	return cfg, nil
+	return NewFilterConfig(cfg), nil
 }
 
 func (p *ConfigParser) Merge(parent interface{}, child interface{}) interface{} {
-	parentCfg := parent.(*Config)
-	childCfg := child.(*Config)
+	parentCfg := parent.(*FilterConfig)
+	childCfg := child.(*FilterConfig)
```
furykerry, line 148: plz add ut for ConfigParser

pkg/servers/e2b/adapters/customized_e2b.go
```diff
@@ -11,8 +11,9 @@ type CustomizedE2BAdapter struct{}
 const CustomPrefix = "/kruise"
 
 // Map maps paths like /kruise/sandbox1234/3000/xxx to sandboxID=sandbox1234 and port=3000
-func (a *CustomizedE2BAdapter) Map(_, _, path string, _ int, _ map[string]string) (
+func (a *CustomizedE2BAdapter) Map(req *ParsedRequest) (
 	sandboxID string, sandboxPort int, extraHeaders map[string]string, err error) {
+	path := req.Path
```
AiRanthem, line 16: Request Changes: There is a parsing bug when a private protocol URL has no trailing path. URLs like /kruise/sandbox-foo/8080 fail to parse, while /kruise/sandbox-foo/8080/some-path works correctly. The root cause appears to be the handling of the trailing / delimiter — the parser assumes a path component always exists. Please fix the URL parsing logic to correctly handle the case where no path is present.

pkg/servers/e2b/adapters/native_e2b.go
```diff
@@ -7,25 +7,100 @@ import (
...
+func (a *NativeE2BAdapter) Map(req *ParsedRequest) (
 	sandboxID string, sandboxPort int, extraHeaders map[string]string, err error) {
-	matches := hostRegex.FindStringSubmatch(authority)
-	if len(matches) != 3 {
-		err = fmt.Errorf("invalid authority format: %s", authority)
-		return
+	authority := req.Authority
+	headers := req.Headers
+
+	// Step 1: Header-based extraction (primary)
```
AiRanthem, line 63: Suggestion: Consider splitting the current logic into two separate functions — one for parsing the header and one for parsing the host. When the header parsing returns an error, the code can fall back to the host parsing logic. This separation would make the flow clearer and easier to maintain going forward.

PR #279, opened 2026-04-15
bugfix for a design flaw, the sandbox-controller inconsistently injec…

### LINKED ISSUE #286: [BUG] function setMainContainerConfigWhenInjectRuntimeSidecar
**What happened**:
There exists a category of cases where users wish to set their own postStartHook. However, due to the current injection logic design: if a postStartHook already exists, injection is skipped by default, and the associated environment variables and volumeMounts are also not injected. Nevertheless, the volume itself will still be injected, as indicated in the following code snippet:

Image

Image

Under this context, the envd process within the user's sandbox still fails to start properly, leading to execution failure. Given this situation, we propose adjusting the injection logic such that when a postStartHook exists, its injection remains disabled, while all other associated injection items (such as environment variables and volumeMounts) should continue to be injected.

**Environment**:
- OpenKruiseAgent version:
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Others:

BH4AWS:
…ts volumes without injecting the corresponding postStart hook when one is already defined, potentially causing volumeMount errors due to missing volume configurations.

There exists a category of cases where users wish to define their own postStart hook. The original code implementation dictates that if a postStart hook is already configured, the sandbox-controller by default disables its injection capability. However, due to an inconsistency in the design, although the postStart hook is not injected, the volume is still injected. This results in a fragmented behavior. In scenarios where users manually configure injection requirements but still expect volumes to be injected, this leads to errors upon initial creation—specifically, errors indicating volumeMounts referencing a non-existent volume configuration.

fixes #286

### DIFF
--- a/pkg/utils/sidecarutils/sidecar_config_inject.go
+++ b/pkg/utils/sidecarutils/sidecar_config_inject.go
@@ -174,7 +174,6 @@ func setMainContainerConfigWhenInjectRuntimeSidecar(ctx context.Context, mainCon
 			log.V(consts.DebugLogLevel).Info("conflicting postStart hooks detected, main container already has a postStart hook defined",
 				"existingHook", mainContainer.Lifecycle.PostStart,
 				"injectedHook", config.MainContainer.Lifecycle.PostStart)
-			return
 		}
 	} else {
 		// Main container doesn't have valid postStart, apply config if available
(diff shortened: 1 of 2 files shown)


furykerry (approved):
/lgtm
/approve

PR #302, opened 2026-04-23
sandbox support recreate upgradeStrategy and lifecycle

zmberg:


pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -348,7 +351,22 @@ func calculateStatus(args core.EnsureFuncArgs) (*agentsv1alpha1.SandboxStatus, b
...
-			logger.Info("sandbox pause not completed, cannot enter resume state temporarily")
+			klog.InfoS("sandbox pause not completed, cannot enter resume state temporarily", "sandbox", klog.KObj(box))
+		}
+
+	case agentsv1alpha1.SandboxUpgrading:
+		// This indicates the podTemplate has changed again during an ongoing upgrade.
+		// Therefore, the Upgrading condition reason must be reset to PreUpgrade to restart the upgrade lifecycle.
+		if newStatus.UpdateRevision != box.Status.UpdateRevision {
+			upgradeCond := utils.GetSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionUpgrading))
+			klog.InfoS("podTemplate changed during upgrade, resetting condition Upgrading reason to PreUpgrade",
```
furykerry: In the SandboxUpgrading case of calculateStatus, upgradeCond is retrieved without a nil guard:

pkg/controller/sandbox/core/common_control.go
```diff
@@ -236,34 +242,151 @@ func (r *commonControl) EnsureSandboxResumed(ctx context.Context, args EnsureFun
...
+		// Execute postUpgrade if configured
+		if hasUpgradeAction(box, false) {
+			var action *agentsv1alpha1.UpgradeAction
+			if box.Spec.Lifecycle != nil {
+				action = box.Spec.Lifecycle.PostUpgrade
+			}
+			result := r.executeUpgradeAction(ctx, pod, box, action)
+			klog.InfoS("postUpgrade result", "sandbox", klog.KObj(box), "succeeded", result.Succeeded, "message", result.Message)
+			if !result.Succeeded {
+				// Record preUpgrade action failed
```
furykerry: preUpgrade -> postUpgrade

pkg/utils/runtime/runtime.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+type RunCmdFuncArgs struct {
+	Sbx           *agentsv1alpha1.Sandbox
+	ProcessConfig *process.ProcessConfig
+	Timeout       time.Duration
+}
+
+// sidecar runtime 提供的 run command 能力
+func RunCommandWithRuntime(ctx context.Context, args RunCmdFuncArgs) (RunCommandResult, error) {
```
furykerry, line 76: RunCommandWithRuntime in pkg/utils/runtime/runtime.go and runCommandWithRuntime in the sandbox-cr package share most of the same logic: create a gRPC connection client, make the call, handle errors. Duplication makes maintenance harder and increases risk of divergence. 

Consider consolidate into a single shared implementation. Have the sandbox-cr package call the utility function instead of duplicating it.

api/v1alpha1/sandbox_types.go
```diff
@@ -231,19 +288,35 @@ const (
 
 	// SandboxConditionInplaceUpdate means inplace update state.
 	SandboxConditionInplaceUpdate SandboxConditionType = "InplaceUpdate"
+
+	// SandboxConditionUpgrading means upgrade state.
+	SandboxConditionUpgrading SandboxConditionType = "Upgrading"
+
+	// SandboxConditionUpgradeAction means the pre-upgrade,post-upgrade execution state.
+	SandboxConditionUpgradeAction SandboxConditionType = "UpgradeAction"
```
furykerry: consider  remove unused UpgradeAction declaration

pkg/controller/sandbox/core/common_control.go
```diff
@@ -297,3 +405,98 @@ func (r *commonControl) handleInplaceUpdateSandbox(ctx context.Context, args Ens
...
+	pod, box, newStatus := args.Pod, args.Box, args.NewStatus
+
+	// Step 1: Delete old Pod (has old template hash)
+	if pod != nil && pod.Labels[agentsv1alpha1.PodLabelTemplateHash] != newStatus.UpdateRevision {
+		if !pod.DeletionTimestamp.IsZero() {
+			klog.InfoS("Waiting for pod deletion to complete", "sandbox", klog.KObj(box))
+			return false, nil
+		}
+		// Delete pod
+		if err := client.IgnoreNotFound(r.Delete(ctx, pod)); err != nil {
```
furykerry: shall we update expectation here ?

pkg/controller/sandbox/core/common_inplace_update_handler.go
```diff
@@ -43,17 +37,14 @@ func handleInPlaceUpdateCommon(
...
 	_, hashImmutablePart := HashSandbox(box)
-
 	// old Pod do not include Labels[pod-template-hash] and do not support inplace update.
 	// Check if inplace update is supported
 	if pod.Labels[agentsv1alpha1.PodLabelTemplateHash] == "" {
 		return true, nil
 		// todo, update inplaceupdate condition
 	} else if box.Annotations[agentsv1alpha1.SandboxHashImmutablePart] != hashImmutablePart {
-		logger.Info("sandbox hash-immutable-part changed, and does not permit in-place upgrades",
+		klog.InfoS("sandbox hash-immutable-part changed, and does not permit in-place upgrades", "sandbox", klog.KObj(box),
```
furykerry, line 47: why change the logger ? previous logger will always log sandbox information

api/v1alpha1/sandbox_types.go
```diff
@@ -47,6 +47,10 @@ const (
 	// RuntimeConfigForInjectAgentRuntime is a valid value for RuntimeConfig.Name.
 	// When set, enables agent runtime sidecar injection for the sandbox.
 	RuntimeConfigForInjectAgentRuntime = "agent-runtime"
+
+	// SandboxAnnotationEnableInplaceUpdate 标记是否启用原地升级能力。
+	// 支持场景：sandbox claim 时启用 InplaceUpdate 能力，实现 cpu, memory, image 等资源的原地升级
+	SandboxAnnotationEnableInplaceUpdate = "sandbox.agents.kruise.io/enable-inplace-update"
```
furykerry: plz remove Chinese comments

pkg/controller/sandbox/metrics.go
```diff
@@ -230,7 +230,7 @@ func recordSandboxMetrics(sandbox *agentsv1alpha1.Sandbox) {
 				sandboxStatusReadyTime.WithLabelValues(namespace, name).Set(float64(condition.LastTransitionTime.Unix()))
 			}
 
-		case agentsv1alpha1.SandboxConditionInplaceUpdate:
+		case agentsv1alpha1.SandboxConditionUpgrading:
```
furykerry: File: pkg/controller/sandbox/metrics.go:144-152

The allPhases list does not include agentsv1alpha1.SandboxUpgrading:

pkg/controller/sandbox/core/common_control.go
```diff
@@ -236,34 +227,151 @@ func (r *commonControl) EnsureSandboxResumed(ctx context.Context, args EnsureFun
...
+		utils.SetSandboxCondition(newStatus, *upgradeCond)
+		fallthrough
+	case agentsv1alpha1.SandboxUpgradingReasonPostUpgrade, agentsv1alpha1.SandboxUpgradingReasonPostUpgradeFailed:
+		// Execute postUpgrade if configured
+		if hasUpgradeAction(box, false) {
+			var action *agentsv1alpha1.UpgradeAction
+			if box.Spec.Lifecycle != nil {
+				action = box.Spec.Lifecycle.PostUpgrade
+			}
+			result := r.executeUpgradeAction(ctx, pod, box, action)
```
furykerry, line 329: During performRecreateUpgrade, the pod is deleted and a new one is created, but args.Pod still refers to the old pod object. When executeUpgradeAction is called in the PostUpgrade case, it passes the old pod to lifecycleHookFunc, which uses it to determine the runtime URL. If the old pod's IP has changed, the post-upgrade hook will execute against the wrong pod or fail.

Suggestion: Re-fetch the pod or update args.Pod with the new pod info after performRecreateUpgrade succeeds. Still present on HEAD.

pkg/controller/sandbox/metrics.go
furykerry: can we add another metric, otherwise the metric name does not match the upgrading condition

test/e2e/sandbox_upgrade_test.go
```diff
@@ -0,0 +1,750 @@
...
+		Expect(updateSandboxSpec(ctx, latest)).To(Succeed())
+
+		By("Waiting for sandbox to return to Running phase after rollback")
+		Eventually(func() agentsv1alpha1.SandboxPhase {
+			_ = k8sClient.Get(ctx, types.NamespacedName{
+				Name:      sandbox.Name,
+				Namespace: sandbox.Namespace,
+			}, sandbox)
+			return sandbox.Status.Phase
+		}, time.Hour*3, time.Millisecond*500).Should(Equal(agentsv1alpha1.SandboxRunning))
```
furykerry: too long timeout value, consider reduce it

PR #303, opened 2026-04-24
fix(e2b): enforce connect timeout extension semantics

### LINKED ISSUE #304: [BUG] E2B connect API will decrease the timeout value of sandboxes
## Background
PR #303 introduced stricter timeout behavior for E2B `ConnectSandbox`:
- when sandbox is already `running`, timeout updates should be **extend-only**;
- no deadline rewrite when new timeout would shorten or keep the same deadline;
- `paused -> resume(connect)` should still apply requested timeout directly;
- regression tests were added in Go and Python.

## Why this issue
The PR already contains the core fix. This issue tracks the behavior as a documented product rule so future changes do not regress timeout semantics across routes/SDK usage.

## Scope
- Keep `ConnectSandbox` timeout semantics consistent with PR #303.
- Preserve current `SetTimeout` behavior and ensure boundaries between `ConnectSandbox` and `SetTimeout` remain clear.
- Keep route/SDK compatibility documentation aligned with implementation.

## Acceptance Criteria
- [ ] `ConnectSandbox` on `running` sandbox only extends timeout and never shortens/equal-rewrites deadline.
- [ ] `ConnectSandbox` on `paused` sandbox still applies requested timeout on resume.
- [ ] Regression tests cover both Go and Python paths and stay passing.
- [ ] Docs clearly describe dual route style and timeout compatibility expectations.

## Reference
- PR: https://github.com/openkruise/agents/pull/303

AiRanthem:
fixes #304 

## Summary
- Enforce **extend-only** timeout updates in `ConnectSandbox` when the sandbox is already `running` (no shorten / no equal-deadline rewrite).
- Keep `paused -> resume(connect)` behavior unchanged: apply requested timeout directly (no guard).
- Add regression coverage in Go and Python for connect-timeout semantics.
- Clarify SDK compatibility and dual route style docs in `pause_resume.go` and `pkg/servers/e2b/AGENTS.md`.

## Test Plan
- `go test -count=1 -run 'TestShouldSkipRunningSandboxTimeoutUpdate|TestConnectSandboxRunningTimeoutGuard|TestSetSandboxTimeoutStillShortensRunningSandbox' ./pkg/servers/e2b/...`

### DIFF
--- a/pkg/servers/e2b/pause_resume.go
+++ b/pkg/servers/e2b/pause_resume.go
@@ -134,7 +134,6 @@ func (sc *Controller) ConnectSandbox(r *http.Request) (web.ApiResponse[*models.S
 	// We only enforce the extend-only guard for sandboxes that were already running when Connect was called.
 	// Paused->resume requests should always apply the requested timeout directly.
 	state, pauseResumeReason := sbx.GetState()
-	autoPause, currentEndAt := ParseTimeout(sbx)
 
 	// Step 1: Resuming the sandbox if it is paused
 	statusCode := http.StatusOK
@@ -154,6 +153,7 @@ func (sc *Controller) ConnectSandbox(r *http.Request) (web.ApiResponse[*models.S
 
 	// Step 2: Update the sandbox timeout
 	log.Info("updating sandbox timeout")
+	autoPause, currentEndAt := ParseTimeout(sbx)
 	if err := sc.updateConnectTimeout(ctx, sbx, request.TimeoutSeconds, state, autoPause, currentEndAt); err != nil {
 		log.Error(err, "failed to update sandbox timeout")
 		return web.ApiResponse[*models.Sandbox]{}, err
--- a/pkg/servers/e2b/pause_resume.go
+++ b/pkg/servers/e2b/pause_resume.go
@@ -1,17 +1,18 @@
 package e2b
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"time"
 
-	"github.com/openkruise/agents/pkg/utils/runtime"
 	"k8s.io/klog/v2"
 
 	"github.com/openkruise/agents/api/v1alpha1"
 	"github.com/openkruise/agents/pkg/sandbox-manager/infra"
 	"github.com/openkruise/agents/pkg/servers/e2b/models"
 	"github.com/openkruise/agents/pkg/servers/web"
+	"github.com/openkruise/agents/pkg/utils/runtime"
 )
 
 func (sc *Controller) PauseSandbox(r *http.Request) (web.ApiResponse[struct{}], *web.ApiError) {
@@ -57,17 +58,6 @@ func (sc *Controller) buildPauseTimeoutOptions(sbx infra.Sandbox, now time.Time)
 	return opts
 }
 
-// shouldSkipRunningSandboxTimeoutUpdate implements the E2B rule for already-running sandboxes:
-// only persist a new timeout when it extends the current effective deadline (strictly later than currentEndAt).
-// currentEndAt is the effective EndAt from ParseTimeout (PauseTime when auto-pause, else ShutdownTime).
-func shouldSkipRunningSandboxTimeoutUpdate(currentEndAt, now time.Time, requestTimeoutSeconds int) (skip bool, requestedEndAt time.Time) {
-	requestedEndAt = TimeAfterSeconds(now, requestTimeoutSeconds)
-	if currentEndAt.IsZero() {
-		return false, requestedEndAt
-	}
-	return !requestedEndAt.After(currentEndAt), requestedEndAt
-}
-
 // ResumeSandbox is DEPRECATED and kept only for old SDK compatibility.
 //
 // E2B exposes one "connect" behavior, but different SDK versions call different endpoints:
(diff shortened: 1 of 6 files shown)


pkg/servers/e2b/pause_resume.go
```diff
@@ -132,19 +157,34 @@ func (sc *Controller) ConnectSandbox(r *http.Request) (web.ApiResponse[*models.S
...
 
-	// Only set timeout if the sandbox has a timeout configured (not never-timeout)
-	if !timeout.IsZero() {
-		opts := sc.buildSetTimeoutOptions(autoPause, time.Now(), request.TimeoutSeconds)
+	// Only set timeout if the sandbox has a timeout configured (not never-timeout).
+	// Running: do not shorten or keep-equal TTL on connect (see shouldSkipRunningSandboxTimeoutUpdate).
+	// Paused→resumed: use the requested timeout directly (no extend-only merge).
+	timeoutEnabled := !currentEndAt.IsZero()
+	now := time.Now()
+	if timeoutEnabled && state == v1alpha1.SandboxStateRunning {
```
furykerry: the timeoutEnabled flag mutation makes the flow hard to follow. It's set, then conditionally flipped to false, then branched on again with an extra else if that's just a logging guard.A cleaner approach is to consolidate the timeout logic into a separate updateConnectTimeout func  and move ParseTimeout closer to its usage, e.g. 

```

func (sc *Controller) updateConnectTimeout(log klog.Logger, ctx context.Context, sbx infra.Sandbox, preConnectState v1alpha1.SandboxState, timeoutSeconds int) *web.ApiError {
	autoPause, currentEndAt := ParseTimeout(sbx)
	if currentEndAt.IsZero() {
		log.Info("skip resetting timeout for never-timeout sandbox")
		return nil
	}

	now := time.Now()
	if preConnectState == v1alpha1.SandboxStateRunning {
		skip, requestedEndAt := shouldSkipRunningSandboxTimeoutUpdate(currentEndAt, now, timeoutSeconds)
		if skip {
			log.Info("skip resetting timeout for running sandbox",
				"currentEndAt", currentEndAt,
				"requestedEndAt", requestedEndAt,
				"requestedTimeoutSeconds", timeoutSeconds)
			return nil
		}
	}

	opts := sc.buildSetTimeoutOptions(autoPause, now, timeoutSeconds)
    ...
```

pkg/servers/e2b/pause_resume.go
```diff
@@ -56,6 +56,25 @@ func (sc *Controller) buildPauseTimeoutOptions(sbx infra.Sandbox, now time.Time)
 	return opts
 }
 
+// shouldSkipRunningSandboxTimeoutUpdate implements the E2B rule for already-running sandboxes:
+// only persist a new timeout when it extends the current effective deadline (strictly later than currentEndAt).
+// currentEndAt is the effective EndAt from ParseTimeout (PauseTime when auto-pause, else ShutdownTime).
+func shouldSkipRunningSandboxTimeoutUpdate(currentEndAt, now time.Time, requestTimeoutSeconds int) (skip bool, requestedEndAt time.Time) {
```
furykerry: if we extract updateConnectTimeout func, then this functions seems un-necessary since the caller can ensure currentEndAt is not zero, we can just cancel the func and put the logic directly in updateConnectTimeout

pkg/servers/e2b/pause_resume_test.go
```diff
@@ -180,7 +182,7 @@ func TestConnectSandbox(t *testing.T) {
 				"sandboxID": tt.sandboxID,
 			}, user))
 
-			if tt.expectStatus >= 300 {
+			if tt.expectStatus >= DefaultTimeoutSeconds {
```
furykerry: The variable DefaultTimeoutSeconds (value 300) is a timeout duration, but it's being compared against tt.expectStatus which is an HTTP status code.

test/e2b/test_sandbox.py
```diff
@@ -138,6 +138,27 @@ def test_timeout(sandbox_context):
...
+
+    info_before = sandbox.get_info()
+    assert info_before.state == SandboxState.RUNNING
+
+    connect_sandbox(sandbox, timeout=300)
+    info_after = sandbox.get_info()
+    assert info_after.state == SandboxState.RUNNING
+
+    # For running sandboxes, connect(timeout=<shorter>) must not shorten endAt.
+    assert info_after.end_at == info_before.end_at
```
furykerry, line 159: This uses strict string equality on end_at. If the E2B SDK ever returns timestamps with sub-second precision or different formatting, this could flake. Consider using a time-comparison with a tolerance, or at minimum a comment explaining why strict equality is safe here.

pkg/servers/e2b/pause_resume.go
```diff
@@ -119,11 +130,16 @@ func (sc *Controller) ConnectSandbox(r *http.Request) (web.ApiResponse[*models.S
 	if apiErr != nil {
 		return web.ApiResponse[*models.Sandbox]{}, apiErr
 	}
-	autoPause, timeout := ParseTimeout(sbx)
+	// `state` is intentionally the pre-connect snapshot.
+	// We only enforce the extend-only guard for sandboxes that were already running when Connect was called.
+	// Paused->resume requests should always apply the requested timeout directly.
+	state, pauseResumeReason := sbx.GetState()
+	autoPause, currentEndAt := ParseTimeout(sbx)
```
furykerry: autoPause, currentEndAt  is not used in the resume sandbox block, consider mov it just before calling updateConnectTimeout

test/e2b/test_sandbox.py
AiRanthem: end_at is a string-type field which is produced by the following code:
```go
	_, endAt := ParseTimeout(sbx)
	sandbox.EndAt = endAt.Format(time.RFC3339)
```
If the timeout is not updated via `connect`, the end_at of info_after and info_before should be completely same

furykerry (approved):
/lgtm
/approve

PR #309, opened 2026-04-27
feat(keys): add secret-to-mysql API key migration script

AiRanthem:
Provide a Python utility to export e2b-key-store entries into MySQL DDL and upsert SQL with HMAC-SHA256 key hashing, and fail fast on malformed entries to avoid partial credential migration.

### DIFF
--- a/hack/migrate_secret_keys_to_mysql.py
+++ b/hack/migrate_secret_keys_to_mysql.py
@@ -1,0 +1,328 @@
+#!/usr/bin/env python3
+#
+# Copyright 2026.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


furykerry:
## Suggestions (Consider Improving)

Review from PR #309 code review — the following are non-blocking improvements for `hack/migrate_secret_keys_to_mysql.py`:

### 1. Add `--dry-run` mode

A `--dry-run` flag that validates all entries and prints the count without writing the SQL file would be useful for pre-migration validation:

```python
parser.add_argument(
    "--dry-run",
    action="store_true",
    help="Validate entries without writing SQL output",
)
```

### 2. Consider writing SQL to stdout when `--output -` is specified

Allowing `--output -` to write to stdout would enable piping directly to `mysql` for operators who prefer one-step migration:

```python
if args.output == "-":
    sys.stdout.write(sql)
else:
    with open(args.output, "w", encoding="utf-8") as fp:
        fp.write(sql)
```

### 3. Add a brief usage example in the docstring

The module docstring describes what the script does but doesn't show how to invoke it. Adding a usage example would help operators:

```python
"""...
Usage:
    python migrate_secret_keys_to_mysql.py --namespace default --pepper mysecret
    E2B_KEY_HASH_PEPPER=mysecret python migrate_secret_keys_to_mysql.py -n default
"""
```

### 4. Consider adding a `--quiet` flag

The script always prints two lines to stdout on success. A `--quiet` flag would be useful when the script is called from automation.

---

*Posted by code-reviewer agent. Full review (including Critical Issues, Warnings, and Positive Findings) available in the local review session.*

PR #313, opened 2026-04-28
fix(sandbox-manager): resolve post-claim 502s and add peer reconciliation for split-brain recovery

AiRanthem:
## Summary
- Fix a regression that could return 502 after sandbox claiming by correcting proxy routing/ext-proc behavior and sandbox manager infra interaction.
- Add peer reconciliation capability to recover from split-brain scenarios and improve cluster peer consistency.

### 1) Fix 502 errors after sandbox claiming
- Update proxy request handling and route resolution logic:
  - `pkg/proxy/ext_proc.go`
  - `pkg/proxy/routes.go`
- Add/adjust regression tests for proxy behavior:
  - `pkg/proxy/ext_proc_test.go`
- Align sandbox manager infra behavior with updated routing expectations:
  - `pkg/sandbox-manager/infra/sandboxcr/infra.go`

### 2) Peer reconciliation for split-brain healing
- Introduce peer reconciliation implementation:
  - `pkg/peers/reconcile.go`
- Add dedicated tests for reconciliation behavior:
  - `pkg/peers/reconcile_test.go`
- Integrate reconciliation with existing peer/memberlist flow:
  - `pkg/peers/memberlist.go`
  - `pkg/peers/peers.go`
- Update related tests and wire-up points:
  - `pkg/peers/memberlist_test.go`
  - `pkg/sandbox-gateway/server/server.go`
  - `pkg/sandbox-manager/core.go`

## Test Plan
- [x] Updated proxy unit tests for post-claim routing/ext-proc behavior.
- [x] Added peer reconciliation unit tests, including split-brain recovery paths.
- [ ] Run full upstream CI for integration-level verification.

### DIFF
--- a/pkg/peers/memberlist.go
+++ b/pkg/peers/memberlist.go
@@ -42,6 +42,8 @@ const (
 	DefaultSuspicionMult = 4
 	// DefaultRetransmitMult is the multiplier for the number of retransmissions
 	DefaultRetransmitMult = 4
+	// DefaultReconcileInterval is the interval between peer reconciliation checks
+	DefaultReconcileInterval = 60 * time.Second
 )
 
 type MemberlistPeers struct {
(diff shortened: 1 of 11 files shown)


AiRanthem:
Closing this PR as superseded by the current master implementation.

The post-claim routing/cache-lag problem has since been addressed through the APIReader fallback and the newer event-driven and periodic route reconciliation paths. Rebasing this part would risk reintroducing assumptions that no longer match the current cache and route contracts.

Periodic memberlist reconciliation may still be useful because the current implementation performs only an initial Join and does not retry after a transient Join failure or a recovered network partition. However, that work should be proposed separately against the current `MemberlistPeers` abstraction, reusing its existing APIReader, namespace/selector ownership, and memberlist address resolution rules.

Keeping this PR open would mix a superseded routing fix with a peer-reconciliation implementation that now requires a substantial rewrite, so a focused follow-up issue or PR would be easier to review and maintain.

PR #256, opened 2026-04-06
feat(sandbox-controller): support rolling update of sandboxset

BITLiutianyang:
### Ⅰ. Describe what this PR does
Support rolliing update of SandboxSet

### Ⅱ. Does this pull request fix one issue?
fixes #76 .

### Ⅲ. Describe how to verify it
1. Create a SandboxSet.
2. Update its spec.template, and set spec.updateStrategy as needed. (For now, MaxSurge and MaxUnavailable is supported. They can be an absolute number (ex: 5) or a percentage of desired pods (ex: 10%), and default is 20%. MaxSurge is the maximum number or percentage of pods that can be created above the desired number of pods. MaxUnavailable is the maximum number or percentage of pods that can be unavailable during the update.)
3. The Sandboxes managed by the SandboxSet should be rolling updated according to the update config.

### Ⅳ. Special notes for reviews
This PR does not resolve all features mentioned in the issue. It is the first in a series of PRs.

PR #317, opened 2026-04-29
feat(controller): consider SandboxTemplate for SandboxSet rolling update

BITLiutianyang:

Previously a SandboxSet only rolled its managed Sandboxes when its own spec changed; mutations to a referenced SandboxTemplate (via spec.templateRef) were silently ignored.

This PR wires that path up end-to-end: getPatch now folds the resolved SandboxTemplate.Spec into the revision hash, so any template change flips UpdateRevision and drives the existing rolling update logic.

Related to #256.

1. Apply a SandboxTemplate + a SandboxSet referencing it with replicas and maxUnavailable;
2. Wait until available, then kubectl edit the template;
3. Old sandboxes managed by SandboxSet are gradually rolling updated.

### DIFF
--- a/pkg/controller/sandboxset/revision.go
+++ b/pkg/controller/sandboxset/revision.go
@@ -19,45 +19,72 @@ limitations under the License.
 package sandboxset
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"hash"
 	"hash/fnv"
 	"strconv"
 
+	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
 	apps "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/dump"
 	"k8s.io/apimachinery/pkg/util/rand"
-
-	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// getPatch returns a strategic merge patch that can be applied to restore a StatefulSet to a
-// previous version. If the returned error is nil the patch is valid. The current state that we save is just the
-// PodSpecTemplate. We can modify this later to encompass more state (or less) and remain compatible with previously
-// recorded patches.
-func (r *Reconciler) getPatch(set *agentsv1alpha1.SandboxSet) ([]byte, error) {
-	str, err := runtime.Encode(r.Codec, set)
-	if err != nil {
-		return nil, err
-	}
-	var raw map[string]interface{}
-	err = json.Unmarshal(str, &raw)
-	if err != nil {
-		return nil, err
-	}
-	objCopy := make(map[string]interface{})
+// getPatch serialises the subset of the SandboxSet spec that determines the
+// template revision hash. Both spec.template (inline) and spec.templateRef are
+// normalised to the same `spec.template` key so that they yield identical
+// hashes whenever they describe the same pod template. This also guarantees
+// that switching between inline template and templateRef (or between two
+// templateRefs) only triggers a rolling update when the underlying pod
+// template actually differs.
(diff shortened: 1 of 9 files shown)


pkg/controller/sandboxset/revision.go
```diff
@@ -19,45 +19,64 @@ limitations under the License.
...
-	objCopy["spec"] = specCopy
-	patch, err := json.Marshal(objCopy)
-	return patch, err
+	if spec != nil {
+		if template, ok := spec["template"].(map[string]interface{}); ok {
+			template["$patch"] = "replace"
+			specCopy["template"] = template
+		}
+	}
+	if set.Spec.TemplateRef != nil {
```
furykerry, line 50: change `if` to `else if`, one can not use both Template and TemplateRef

pkg/controller/sandboxset/revision.go
```diff
@@ -19,45 +19,64 @@ limitations under the License.
...
+		tplSpecRaw, err := json.Marshal(tpl.Spec)
+		if err != nil {
+			return nil, err
+		}
+		specCopy["templateRef"] = map[string]interface{}{
+			"name": set.Spec.TemplateRef.Name,
+			"spec": json.RawMessage(tplSpecRaw),
+		}
+	}
+	return json.Marshal(map[string]interface{}{"spec": specCopy})
```
furykerry, line 87: i think we should create a SandboxTemplate CR in NewControllerRevision and cache the calculated hash in sandboxtemplate

pkg/controller/sandboxset/revision.go
```diff
@@ -19,45 +19,64 @@ limitations under the License.
...
+			return nil, fmt.Errorf("failed to resolve sandbox template %s/%s: %w",
+				set.Namespace, set.Spec.TemplateRef.Name, err)
+		}
+		tplSpecRaw, err := json.Marshal(tpl.Spec)
+		if err != nil {
+			return nil, err
+		}
+		specCopy["templateRef"] = map[string]interface{}{
+			"name": set.Spec.TemplateRef.Name,
+			"spec": json.RawMessage(tplSpecRaw),
```
furykerry: consider add "template" (not templateRef) in specCopy just like embedded sandbox template case

pkg/controller/sandboxset/sandboxtemplate_event_handler.go
```diff
@@ -0,0 +1,97 @@
...
+var _ handler.EventHandler = &SandboxTemplateEventHandler{}
+
+func NewSandboxTemplateEventHandler(c client.Reader) *SandboxTemplateEventHandler {
+	return &SandboxTemplateEventHandler{client: c}
+}
+
+func (h *SandboxTemplateEventHandler) Create(context.Context, event.TypedCreateEvent[client.Object], workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+}
+
+func (h *SandboxTemplateEventHandler) Update(ctx context.Context, evt event.TypedUpdateEvent[client.Object], w workqueue.TypedRateLimitingInterface[reconcile.Request]) {
```
furykerry: sandbox template is an immutable object, so it is not necessary to handle sandbox template mutation events, we can just handle the case where sandboxTemplateRef changes (from template a to template b)

pkg/controller/sandboxset/utils.go
```diff
@@ -174,13 +183,20 @@ func scaleExpectationSatisfied(ctx context.Context, scaleExpectation expectation
...
 	generateName := fmt.Sprintf("%s-", sbs.Name)
-	template := sbs.Spec.Template.DeepCopy()
+	// spec.template is optional when spec.templateRef is used; guard the copy.
+	var template *corev1.PodTemplateSpec
+	var inheritedLabels, inheritedAnnotations map[string]string
+	if sbs.Spec.Template != nil {
+		template = sbs.Spec.Template.DeepCopy()
+		inheritedLabels = template.Labels
+		inheritedAnnotations = template.Annotations
+	}
```
furykerry, line 204: if one use sandboxTemplateRef, we can assign spec.template.labels and spec.template.annotations to inheritedLabels and inheritedAnnotations

furykerry (approved):
/lgtm
/approve

PR #325, opened 2026-05-02
feat: add team-based namespace isolation and team-scoped API key authorization

AiRanthem:

Introduces a **Team** concept with namespace isolation across the entire sandbox lifecycle, and migrates API key authorization from a flat admin/non-admin model to team-based scoping.

**What changed:**

1. **Team model & API keys** — Added `Team`/`ListedTeam` types with a reserved `admin` team (ID `550e8400-...`). API keys are now scoped to teams via `teamName`. Non-admin callers can only create keys for their own team; admins can create keys across teams. Added `GET /teams` for upstream-compatible team listing.

2. **Namespace isolation** — All cache, infra, and manager APIs migrated from flat `(ctx, user/id)` signatures to options structs carrying `Namespace`. Sandboxes, checkpoints, sandbox sets, and templates are now looked up within their namespace, preventing cross-team resource access.

3. **Permission middleware** — Replaced the catch-all `CheckAdminKey` middleware with `CheckCreateAPIKeyPermission` (validates team ownership + namespace existence) and `CheckDeleteAPIKeyPermission` (team-scoped delete authorization).

4. **Key storage layer** — Both MySQL and Secret backends refactored for team-scoped operations (`FindTeamByName`, `ListTeams`, team-aware CRUD).

5. **RBAC** — Sandbox-manager now requires namespace `get/list/watch` permissions for namespace validation.

NONE

1. **Unit tests**: `go test ./pkg/... -count=1`
2. **E2B tests**: `cd test/e2b && pytest test_apikey.py -v`
3. **Manual verification**:
   - Create a team and its API key → verify sandbox operations are namespace-isolated
   - Try creating an API key for another team as a non-admin → expect 403
   - List sandboxes with a team API key → only see sandboxes in the team's namespace
   - Admin key → can list across namespaces, create keys for any team

- The `pkg/cache/Provider` interface breaking change propagates through all consumers. The options struct pattern (`GetClaimedSandboxOptions`, etc.) is designed to be extensible for future filters.
- Empty `opts.Namespace` means "no explicit namespace filter" — the effective scope depends on the cache's pre-filtering via `SandboxManagerOptions.SandboxNamespace`. See `pkg/cache/AGENTS.md` for namespace semantics.
- `AdminTeamID` is a hardcoded UUID (`550e8400-e29b-41d4-a716-446655449999`) — admin users use this as their team ID.
- The `GET /teams` endpoint is upstream-compatible with the E2B API, returning `ListedTeam` structs.

---

The following changes are side modifications not directly related to the main feature:

- **Qoder skill file** (`.qoder/skills/create-upstream-issue/SKILL.md`, +188 lines): Internal development tooling — should be excluded from upstream PR.
- **CI: memo-check job** (`.github/workflows/ci.yaml`): Adds a CI step to block `MEMO.md` in the repo root, with `needs` dependency added to all downstream jobs.
- **Trivial .gitignore change** (`.gitignore`): Added a blank line.


.github/workflows/ci.yaml
```diff
@@ -24,9 +24,23 @@ env:
...
 jobs:
+  memo-check:
+    name: Check forbidden files
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout Actions Repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Ensure MEMO.md is absent
+        run: |
+          if [ -f "./MEMO.md" ]; then
```
furykerry: why not put MEMO.md in gitignore?

pkg/servers/e2b/routes.go
```diff
@@ -121,10 +127,74 @@ func (sc *Controller) CheckApiKey(ctx context.Context, r *http.Request) (context
...
+
+	// Validate namespace of target team exists
+	_, found, err := sc.keys.FindTeamByName(ctx, targetTeamName)
+	if err != nil {
+		return ctx, &web.ApiError{
+			Code:    http.StatusInternalServerError,
+			Message: fmt.Sprintf("Failed to find team: %v", err),
+		}
+	}
+	if !found {
```
furykerry: consider validate namespace even if FindTeamByName had found result

pkg/servers/e2b/keys/secret.go
```diff
@@ -246,25 +332,31 @@ func (k *secretKeyStorage) CreateKey(ctx context.Context, user *models.CreatedTe
...
+	k.storeKey(createdKey)
+	return createdKey, nil
 }
 
 func (k *secretKeyStorage) DeleteKey(ctx context.Context, key *models.CreatedTeamAPIKey) error {
 	if key == nil {
 		return nil
 	}
+	// The last key of admin team cannot be deleted
+	if TeamForKey(key).ID == models.AdminTeamID && k.activeKeyCountForTeam(models.AdminTeamID) <= 1 {
```
furykerry: consider move this check inside retryUpdateSecret, multiple sandbox manager may try to delete admin key, so the in-memory check can not ensure the last admin key exists

pkg/servers/e2b/keys/mysql.go
```diff
@@ -309,42 +334,73 @@ func (k *mysqlKeyStorage) CreateKey(ctx context.Context, user *models.CreatedTea
...
 	var hash string
-	if key.Key != "" {
-		hash = k.hashKey(key.Key)
-	} else {
-		var e TeamAPIKeyEntity
-		if err := k.db.WithContext(ctx).Where(&TeamAPIKeyEntity{UID: key.ID.String()}).
-			Select("key_hash").First(&e).Error; err != nil {
-			if !errors.Is(err, gorm.ErrRecordNotFound) {
-				return fmt.Errorf("prefetch key hash: %w", err)
+	err := k.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
```
furykerry: it is too costly to use transaction for key creation, is it possible to ensure AdminKeyID cannot be deleted

pkg/servers/e2b/keys/mysql.go
```diff
@@ -283,21 +297,32 @@ func (k *mysqlKeyStorage) CreateKey(ctx context.Context, user *models.CreatedTea
...
-			Name:         name,
-			Key:          k.hashKey(newKey.String()),
-			TeamID:       teamID,
-			CreatedByUID: &createdBy,
-		}
-		if err := k.db.WithContext(ctx).Create(&entity).Error; err != nil {
-			if errors.Is(err, gorm.ErrDuplicatedKey) {
-				return errUUIDCollision
+		createdBy := key.ID.String()
+		return k.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
```
furykerry: it is not necessary to use transaction here if UID is an unique key

pkg/servers/e2b/api_key.go
```diff
@@ -47,11 +49,15 @@ func (sc *Controller) ListAPIKeys(r *http.Request) (web.ApiResponse[[]*models.Te
 }
 
 func (sc *Controller) CreateAPIKey(r *http.Request) (web.ApiResponse[*models.CreatedTeamAPIKey], *web.ApiError) {
-	var request models.NewTeamAPIKey
-	if err := json.NewDecoder(r.Body).Decode(&request); err != nil {
-		return web.ApiResponse[*models.CreatedTeamAPIKey]{}, &web.ApiError{
-			Message: err.Error(),
+	request, ok := GetNewAPIKeyRequestFromContext(r.Context())
+	if !ok {
+		var decoded models.NewTeamAPIKey
```
furykerry, line 57: why we need decode the body again for the api key? CheckCreateAPIKeyPermission had already decode it .

pkg/sandbox-manager/infra/sandboxcr/sandbox.go
```diff
@@ -108,7 +108,9 @@ func (s *Sandbox) retryUpdate(ctx context.Context, modifier ModifierFunc) error
 	log := klog.FromContext(ctx).WithValues("sandbox", klog.KObj(s.Sandbox))
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		// get the latest sandbox from cache
-		sbx, err := s.Cache.GetClaimedSandbox(ctx, stateutils.GetSandboxID(s.Sandbox))
+		sbx, err := s.Cache.GetClaimedSandbox(ctx, cache.GetClaimedSandboxOptions{
+			SandboxID: stateutils.GetSandboxID(s.Sandbox),
```
furykerry, line 112: add namespace in the option?

pkg/sandbox-manager/api.go
```diff
@@ -141,10 +144,20 @@ func (m *SandboxManager) ListCheckpoints(ctx context.Context, user string, p *ut
...
+	log := klog.FromContext(ctx).WithValues("checkpointID", opts.CheckpointID)
+	cp, err := m.infra.GetCache().GetCheckpoint(ctx, cache.GetCheckpointOptions{
+		Namespace:    opts.Namespace,
+		CheckpointID: opts.CheckpointID,
+	})
+	if err != nil {
+		log.Error(err, "failed to get checkpoint before delete")
+		return errors.NewError(errors.ErrorNotFound, err.Error())
+	}
+	if user != "" && cp.GetAnnotations()[v1alpha1.AnnotationOwner] != user {
```
furykerry: what if user is empty? the logic seems changed from previous version

pkg/servers/e2b/keys/mysql.go
```diff
@@ -65,20 +66,21 @@ var autoMigrateMySQLModels = func(ctx context.Context, db *gorm.DB) error {
 type TeamEntity struct {
 	gorm.Model
 	UID  string `gorm:"column:uid;type:varchar(36);uniqueIndex;not null"`
-	Name string `gorm:"type:varchar(255);index;not null"`
+	Name string `gorm:"type:varchar(255);uniqueIndex;not null"`
```
furykerry, line 71: change of database schema need update the hack/mysql-schema.sql and may need a migration script (checking for duplicates)

pkg/servers/e2b/keys/secret.go
```diff
@@ -216,15 +245,72 @@ func (k *secretKeyStorage) updateSecret(ctx context.Context, id string, apiKey *
...
+}
+
+func findTeamByNameInSecret(secret *corev1.Secret, teamName string) (*models.Team, bool) {
+	for _, bytes := range secret.Data {
+		var apiKey models.CreatedTeamAPIKey
+		if err := json.Unmarshal(bytes, &apiKey); err != nil {
+			continue
+		}
+		team := TeamForKey(&apiKey)
+		if team.Name == teamName {
```
furykerry, line 285: why not use idxByTeam first ?

pkg/sandbox-manager/api.go
```diff
@@ -141,10 +144,20 @@ func (m *SandboxManager) ListCheckpoints(ctx context.Context, user string, p *ut
 	return checkpoints, nextToken, nil
 }
 
-// DeleteCheckpoint deletes a checkpoint and its associated sandbox template
-func (m *SandboxManager) DeleteCheckpoint(ctx context.Context, user string, checkpointID string) error {
-	log := klog.FromContext(ctx).WithValues("checkpointID", checkpointID)
-	if err := m.infra.DeleteCheckpoint(ctx, user, checkpointID); err != nil {
+func (m *SandboxManager) DeleteCheckpoint(ctx context.Context, user string, opts infra.DeleteCheckpointOptions) error {
+	log := klog.FromContext(ctx).WithValues("checkpointID", opts.CheckpointID)
+	cp, err := m.infra.GetCache().GetCheckpoint(ctx, cache.GetCheckpointOptions{
```
furykerry: m.infra.DeleteCheckpoint will also call GetCheckpoint internally, consider delete GetCheckpoint calling in the DeleteCheckpoint

pkg/servers/e2b/routes.go
```diff
@@ -63,9 +67,10 @@ func (sc *Controller) registerRoutes() {
 
 	// API Keys management endpoints
 	if sc.keyCfg != nil {
-		RegisterE2BRoute(sc.mux, http.MethodGet, "/api-keys", sc.ListAPIKeys, sc.CheckApiKey, sc.CheckAdminKey)
-		RegisterE2BRoute(sc.mux, http.MethodPost, "/api-keys", sc.CreateAPIKey, sc.CheckApiKey, sc.CheckAdminKey)
-		RegisterE2BRoute(sc.mux, http.MethodDelete, "/api-keys/{apiKeyID}", sc.DeleteAPIKey, sc.CheckApiKey, sc.CheckAdminKey)
+		RegisterE2BRoute(sc.mux, http.MethodGet, "/teams", sc.ListTeams, sc.CheckApiKey)
```
furykerry, line 70: is there a API for create team ?

.github/workflows/ci.yaml
AiRanthem: I have the Agent use MEMO.md to save work progress, and I hope to manage it with git version control — for example, when rolling back code, the work records should be rolled back together.

pkg/servers/e2b/routes.go
AiRanthem: Yes, the namespace may be deleted after the team is created, the validation is necessary

pkg/servers/e2b/keys/secret.go
AiRanthem: OK

pkg/servers/e2b/keys/mysql.go
AiRanthem: The transaction cost here is minimal — it's just 2~3 row-level operations (resolve team + insert key) under InnoDB row locks. It's necessary because `resolveTeamByName` may create or restore a team atomically with the key insert; without it we'd risk orphan teams or dangling FKs. A bare `INSERT` under autocommit is itself an implicit transaction, so the overhead difference is negligible.

As for making AdminKeyID undeletable — `DeleteKey` already prevents deleting the **last** admin key via `ErrLastAdminKey`. If you want AdminKeyID to be unconditionally protected, we can add an early guard:

```go
if key.ID.String() == AdminKeyID.String() {
    return ErrAdminKeyUndeletable
}
```

AiRanthem: Delete policy changed to "only well-known admin-key cannot be deleted"

pkg/servers/e2b/keys/mysql.go
AiRanthem: Team UID is removed, all codes are simplified

pkg/servers/e2b/api_key.go
AiRanthem: An comment is added

pkg/sandbox-manager/api.go
AiRanthem: `user != ""` is removed

pkg/servers/e2b/keys/secret.go
AiRanthem: findTeamByNameInSecret scans the raw secret data instead of the in-memory cache
(idxByTeam) to guarantee consistency with the exact secret revision being updated.
During retryCreateKey, each retry re-reads the secret from the API server; using
the cache here could miss teams created by other replicas between retries.

pkg/servers/e2b/keys/mysql.go
AiRanthem: A new schema will be generated with skill /generate-mysql-schema after UID is removed

pkg/sandbox-manager/infra/sandboxcr/sandbox.go
AiRanthem: added

pkg/servers/e2b/routes.go
AiRanthem: E2B doesn't have such an API. A new team is created by specifying a non-existent team in the CreateAPIKey method. Currently, only the admin team's API key has the permission to specify a different team when calling CreateAPIKey, meaning only the admin team can create teams.

pkg/sandbox-manager/api.go
AiRanthem: Moved the owner check to DeleteCheckpoint

PR #342, opened 2026-05-07
refactor(controller): remove redundant SandboxUpdateOps concurrency check

### LINKED ISSUE #341: Remove redundant SandboxUpdateOps concurrency check in controller
# Remove redundant SandboxUpdateOps concurrency check in controller

## Description
In `pkg/controller/sandboxupdateops/sandboxupdateops_controller.go`, the controller's `Reconcile` loop manually checks for active operations in the same namespace to prevent concurrent update conflicts. This code block is accompanied by a `TODO` acknowledging it as a short-term workaround and stating that a webhook would be a more robust approach.

Since a ValidatingWebhook (`pkg/webhook/sandboxupdateops/validating/sandboxupdateops_validate.go`) is now fully implemented and successfully rejects concurrent creations at the API admission level, this controller-side check is completely redundant.

## Impact
- **Performance:** The controller performs an unnecessary `List()` API call during every reconciliation loop, which adds load to the Kubernetes API server and slows down the loop.
- **Race Conditions:** Controller-level concurrency checks based on `List()` can be subject to race conditions, unlike the admission webhook which operates atomically.
- **Technical Debt:** The outdated `TODO` and obsolete logic make the codebase confusing to new contributors.

## Proposed Changes
1. Remove the `opsList` conflict-check logic block from `sandboxupdateops_controller.go` (around lines 105-120).
2. Remove the outdated `TODO` comment above it.
3. Remove or update any unit tests in `sandboxupdateops_controller_test.go` (like `TestReconcile_ConcurrentOpsInNamespace`) that explicitly expect the controller to handle this, as the responsibility now strictly belongs to the webhook.

PRAteek-singHWY:
## Summary
This PR removes a redundant concurrency check in the `SandboxUpdateOps` controller, eliminating technical debt and optimizing the reconciliation loop.

## Problem
In `pkg/controller/sandboxupdateops/sandboxupdateops_controller.go`, the controller was manually listing all `SandboxUpdateOps` resources in a namespace to prevent concurrent update operations. This check was accompanied by a `TODO` acknowledging it as a short-term solution and suggesting a validating webhook.

Since a validating webhook (`pkg/webhook/sandboxupdateops/validating/sandboxupdateops_validate.go`) is now fully implemented and successfully rejects concurrent creations at the admission level, the controller's manual check is completely obsolete.

## Solution
Removed the redundant `opsList` conflict-check logic and the outdated `TODO` from the controller's `Reconcile` function. Also removed the corresponding obsolete unit test.

## Impact
- **Performance**: Eliminates an unnecessary `List` API call on every reconciliation loop, reducing load on the Kubernetes API Server.
- **Tech Debt**: Removes outdated comments and redundant race-prone controller logic.

## Checklist
- [x] Removed redundant concurrency check in `sandboxupdateops_controller.go`.
- [x] Removed obsolete unit test for concurrency check.
- [x] Verified ValidatingWebhook already handles this validation.

**fixes #341**


PRAteek-singHWY:
Hello @furykerry, @AiRanthem, and @zmberg

I was reviewing the codebase and noticed an opportunity to clean up some tech debt in the `SandboxUpdateOps` controller. I found a redundant `List()` check that was manually preventing concurrent operations, accompanied by a `TODO` pointing to a webhook solution.

Since the `sandboxupdateops_validate.go` webhook is now fully implemented and successfully handles this at the admission layer, I’ve removed the obsolete loop and its associated test. This should optimize the reconciliation cycle by saving an unnecessary API server call. 

Let me know if you need any adjustments or further testing.
Happy to iterate based on ur suggestions.
Thank You.

furykerry:
@PRAteek-singHWY thanks for your patch, but concurrent update ops for a single sandbox should be avoided, so simply remove this concurrent check is not an option, we should comes up with some more flexible validation rules

PRAteek-singHWY:
> @PRAteek-singHWY thanks for your patch, but concurrent update ops for a single sandbox should be avoided, so simply remove this concurrent check is not an option, we should comes up with some more flexible validation rules

Hey @furykerry 
Thanks for the feedback. 

That makes sense, we definitely want to avoid race conditions on a single sandbox.

Instead of the strict 'one-per-namespace' rule, I'll update this PR to implement a more granular check: I'll make the webhook verify that the selector of the new SandboxUpdateOps doesn't overlap with any currently active ones. This way, users can run concurrent updates across the namespace as long as they target different sets of sandboxes.

I'll push the updated logic shortly.

Thank you.

PRAteek-singHWY:
Hey @furykerry, your point about avoiding race conditions on individual sandboxes makes total sense. I've updated the logic to block updates only if their selectors overlap, this keeps things safe for single sandboxes while allowing parallel updates across the rest of the namespace.

I've also included a small fix in the `Makefile` to add an `envtest` dependency for the `test` target. I noticed the unit-test CI was failing on fresh runners because the `setup-envtest` binary wasn't being downloaded automatically. This ensures the PR (and future ones) can pass CI consistently.

 Happy to iterate further based on ur suggestions.
 Thank you.

pkg/utils/sandboxutils/selector_utils.go
```diff
@@ -0,0 +1,80 @@
...
+package sandboxutils
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SelectorsOverlap checks if two label selectors could potentially match the same resources.
+// It returns true if there is a possible overlap.
+// Note: This is a conservative check. If it's unsure, it returns true.
+func SelectorsOverlap(s1, s2 *metav1.LabelSelector) bool {
```
furykerry: plz copy the utility in kruise repo: https://github.com/openkruise/kruise/blob/master/pkg/util/selector.go

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -102,23 +102,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	// 2. Check if another SandboxUpdateOps in the same namespace is already Updating
```
furykerry, line 105: we may need double checking for concurrent sandboxupdateops in the controller, since the informer of webhook may not updated in realtime

furykerry:
@zmberg the informer cache sync delay problem always exists in both webhook and controller , i think we need a final check in the sandbox controller. The sandbox controller can lock the sandbox in pre-upgrade phase so that no other sandboxupdateops can operate in the same sandbox

pkg/utils/sandboxutils/selector_utils.go
PRAteek-singHWY: Thanks @furykerry,  addressed both in two commits on this branch:

  **1. Selector util (43a047e):** replaced the hand-rolled `SelectorsOverlap` with a copy of kruise's `pkg/util/selector.go`
  (`IsSelectorOverlapping` etc.), license header preserved. Swapped the `k8s.io/kubernetes/pkg/util/slice` calls for std
  `slices.Contains` so we skip that transitive dep; rest is verbatim.

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
PRAteek-singHWY: **2. Controller check (f2acf3a):** before `Pending → Updating`, the reconciler now lists siblings and yields if one overlaps
  - `Updating` wins outright, otherwise older `CreationTimestamp` wins (name asc breaks ties). On yield: `Phase=Pending` + 5s
  requeue.

pkg/utils/sandboxutils/selector_utils.go
```diff
@@ -0,0 +1,202 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+// IsSelectorLooseOverlap indicates whether selectors overlap (indicates that selector1, selector2 have same key, and there is a certain intersection）
+//  1. when selector1、selector2 don't have same key, it is considered non-overlap, e.g. selector1(a=b) and selector2(c=d)
+//  2. when selector1、selector2 have same key, and matchLabels & matchExps are intersection, it is considered overlap.
+func IsSelectorLooseOverlap(selector1, selector2 *metav1.LabelSelector) bool {
```
furykerry, line 128: Dead code: IsSelectorLooseOverlap and its helpers are unused， consider remove them

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -102,23 +108,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
```
furykerry, line 105: After removing the old step-2 block, the remaining comments jump from // 1. Get SandboxUpdateOps directly to // 3. Handle deletion. Pre-existing duplicates (two // 3., two // 4., two // 7.) make this worse, but the PR exacerbates the problem by removing the only "step 2" anchor. Now the function reads as step 1 → step 3, which is confusing. Consider correct the step number

PRAteek-singHWY: Hi @furykerry  will look into this.
Thank you.

pkg/utils/sandboxutils/selector_utils.go
PRAteek-singHWY: > Dead code: IsSelectorLooseOverlap and its helpers are unused， consider remove them

HI @furykerry  Done. Dropped `IsSelectorLooseOverlap`, `isMatchExpOverlap`, and `convertSelectorToMatchExpressions` from `pkg/utils/sandboxutils/selector_utils.go`, and removed the `TestIsSelectorLooseOverlap` table from `selector_utils_test.go`. `IsSelectorOverlapping` is the only entry point we actually use (webhook + new controller double-check), so the file is now just that plus its `isDisjoint` / `sliceOverlaps` / `sliceContains` helpers. `go test ./pkg/utils/sandboxutils/...` and `go vet` pass.

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
PRAteek-singHWY: Hi @furykerry  
 Renumbered the steps in `Reconcile` so they read 1 → 10 in order, no gaps and no cross-step duplicates:

1. Get SandboxUpdateOps
2. Handle deletion (was `// 3.`, now fills the old "step 2" slot)
3. Ensure finalizer
4. Terminal state short-circuit
5. Concurrency safeguard (the new sibling re-check — was an unnumbered block, now an explicit step)
6. Convert selector to `labels.Selector` (was the duplicate `// 3.`)
7. List matching sandboxes (was the duplicate `// 4.`)
8. Classify sandbox states
9. Phase state machine
10. Update status

Also removed the duplicate `// 7. Update status` comment on the alt return path so step 10 isn't repeated. No behavior change — pure comment cleanup. Tests still green.

PRAteek-singHWY: Hi @furykerry , 
would really appreciate your review and guidance on this.
If there are any iterations to be made.
Thank you

PR #354, opened 2026-05-08
feat(observability): add /healthz, /readyz, reconcile metrics, and webhook admission metrics

### LINKED ISSUE #353: [Observability] Add operator-facing metrics and health probes across sandbox-manager subsystems
### Description

`sandbox-manager` already exposes good *application-level* sandbox metrics — `sandbox_creation_duration`, `sandbox_inplace_update_duration`, `sandbox_status_phase`, etc. (`pkg/controller/sandbox/metrics.go`, ~15 collectors). Recent work is extending the picture for the E2B API surface (#333, #337, #335).

However, several **operator-facing** subsystems still have **zero observability**, which makes diagnosing prod incidents hard:

| Subsystem | Current state | Operator pain |
|---|---|---|
| Process health | No `/healthz`, no `/readyz` — only `/debug/pprof/*` (and only when `--enable-pprof`) | Liveness/readiness probes can't detect hung controllers, unsynced caches, lost memberlist quorum, or unreachable DB. |
| Reconcile loop | No `sandbox_reconcile_duration` histogram, no per-result counter | Can't alert on slow reconciles or rising error rates. |
| Webhook admission | No latency histogram, no admit/deny counter | Can't observe webhook-induced API server slowness. |
| Memberlist peer state | `pkg/peers/memberlist.go` exports nothing to Prometheus | Silent partitions; operators only notice when traffic drops. |
| API key cache | `mysqlKeyStorage` / `secretKeyStorage` have no hit/miss counters | Cannot tell if cache layering is effective; auth latency tail is invisible. |
| Critical-path logging | `sandbox_controller.go:285-304` logs full JSON via `DumpJson` per status update; structured fields missing on other hot paths | Logs aren't structured-queryable; reconcile path allocates GC pressure. |

This issue tracks closing all six gaps as a coherent observability suite.

### Scope (in 2 PRs)

I'll deliver this in two XL PRs to keep each review focused and shippable:

**PR 1 — K8s-facing observability** _(in progress)_

- (a) `cmd/sandbox-manager/main.go`, `pkg/cache/cache.go` — enable controller-runtime's built-in `/healthz` and `/readyz` via a new `--health-probe-bind-address` flag.
- (b) `pkg/controller/sandbox/sandbox_controller.go`, `pkg/controller/sandbox/metrics.go` — `sandbox_reconcile_duration_seconds{namespace,result}` histogram + drop the inline `DumpJson` allocation from the hot status-update path.
- (c) `pkg/webhook/pod/validating/`, `pkg/webhook/sandboxset/validating/` — `sandbox_admission_duration_seconds{webhook,operation,allowed}` histogram and `sandbox_admission_total{webhook,operation,allowed}` counter, wrapped around existing handlers (no behavior change).

**PR 2 — Distributed observability + log hygiene** _(follow-up)_

- (d) `pkg/peers/memberlist.go` — `sandbox_peer_state{node,state}` gauge (alive/suspect/dead) + `sandbox_peer_join_duration_seconds` histogram.
- (e) `pkg/servers/e2b/keys/mysql.go`, `pkg/servers/e2b/keys/secret.go` — `e2b_key_cache_hits_total{storage,key_type}` and `e2b_key_cache_misses_total{storage,key_type}` counters.
- (f) misc critical paths — replace `fmt.Sprintf`-based log lines with structured fields where context (sandbox ID, namespace, operation) is currently missing.

### Why bundle into one tracking issue

These six pieces share design questions (metric naming, label cardinality, port conventions). Settling those once here avoids re-litigating in each sub-PR. The PRs themselves stay independently mergeable.

### Open questions for maintainers

1. Default `--health-probe-bind-address` of `:8081` (kubebuilder convention) — or do you prefer a different default?
2. Cardinality concerns with `namespace` as a label given multi-tenant deployments? `pkg/controller/sandbox/metrics.go` already uses `namespace` extensively, so I'm assuming this is acceptable — flag if not.
3. Any preference on metric name prefix? Existing metrics use `sandbox_*` for controller metrics and `e2b_*` for the API surface; I'm following that split.

I have **#296** also active, but happy to drive both in parallel.

PRAteek-singHWY:
## Summary

This is the **first of two PRs** for #353. It closes the K8s-facing gaps in `sandbox-manager` so operators can actually tell what's going on:

- **(a)** Adds `/healthz` and `/readyz` HTTP endpoints so Kubernetes liveness/readiness probes work properly.
- **(b)** Adds a Prometheus histogram for how long each `Sandbox` reconcile takes, and removes a hot-path JSON dump that allocated on every status update.
- **(c)** Adds Prometheus latency + count metrics for every admission webhook, by wrapping each handler in a small decorator. No webhook changes its behavior.

The other three sub-pieces — peer-state metrics, key-cache hit/miss counters, and log-message cleanup — will land in a follow-up PR against the same issue.

## Why this matters

`sandbox-manager` already publishes a lot of *sandbox-level* metrics (`sandbox_creation_duration`, `sandbox_status_phase`, etc.) — but operators have no way to answer simpler questions like:

- "Is this pod actually healthy, or just bound to a port?"
- "Why are reconciles slow today?"
- "Which webhook is making my API server feel slow?"

That's the gap this PR closes.

### (a) Health probes

The controller-runtime manager already supports `/healthz` and `/readyz` — the option was just disabled in `pkg/cache/cache.go` (`HealthProbeBindAddress: ""`). This PR enables it via a new `--health-probe-bind-address` flag.

- Default: `:8081` (kubebuilder convention).
- Disable: pass `--health-probe-bind-address=""`.
- When enabled, both endpoints respond `200 OK` once the manager has started.

### (b) Reconcile metrics

New histogram: **`sandbox_reconcile_duration_seconds{namespace,result}`**, where `result` is `success` / `requeue` / `error`.

Also drops a `utils.DumpJson(newStatus)` call from the success log line in `updateSandboxStatus`. That JSON marshal happened on **every** status update; replaced with three structured fields (`phase`, `observedGeneration`, `updateRevision`) which are cheaper to allocate and easier to query in log aggregators.

### (c) Webhook admission metrics

Two new metrics:

- **`sandbox_admission_duration_seconds{webhook,operation,allowed}`** — how long each handler takes.
- **`sandbox_admission_total{webhook,operation,allowed}`** — admit/deny counts.

Implementation is a tiny `instrumentedHandler` decorator wrapped around every existing handler at registration. No handler logic changes — `Handle()` returns whatever the inner handler returns.

## Backward compatibility

- Default port `:8081` is the kubebuilder convention; operators who don't want the new port can disable it with an empty string.
- The new metrics are purely additive — existing metrics are untouched.
- The webhook decorator is transparent: every existing admission decision passes through unchanged.
- No new dependencies.

## Tests

- Round-trip test for the new `HealthProbeBindAddress` field.
- 4 cases for the result-label mapping (`success` / `requeue` / `error`).
- Histogram observation count assertion for reconcile duration.
- 5 cases for the webhook decorator: allowed, denied, errored, response pass-through, and duration recorded.

All affected packages green: `pkg/sandbox-manager/...`, `pkg/cache/...`, `pkg/servers/e2b/...`, `pkg/webhook/...`, `pkg/controller/...`, `cmd/...`.

> Note: end-to-end probe testing would need an envtest-backed `*rest.Config` (controller-runtime's manager dials the API server during construction). This repo doesn't have envtest set up yet — happy to file a separate issue for that if it's useful.

## Checklist

- [x] Build + vet pass
- [x] All affected-package tests pass
- [x] No new dependencies

**fixes #353 (sub-pieces a, b, c). Sub-pieces (d) memberlist peer-state, (e) key-cache hit/miss, and (f) structured-field logging on hot paths will land in a follow-up PR against the same issue.**

### DIFF
--- a/cmd/sandbox-manager/main.go
+++ b/cmd/sandbox-manager/main.go
@@ -83,6 +83,7 @@ func main() {
 	var kubeClientQPS float64
 	var kubeClientBurst int
 	var memberlistBindPort int
+	var healthProbeBindAddress string
 	var e2bKeyStorage string
 	var e2bKeyStorageDisableAutoMigrate bool
 
@@ -111,6 +112,8 @@ func main() {
 	pflag.Float64Var(&kubeClientQPS, "kube-client-qps", 500, "QPS for Kubernetes client")
 	pflag.IntVar(&kubeClientBurst, "kube-client-burst", 1000, "Burst for Kubernetes client")
 	pflag.IntVar(&memberlistBindPort, "memberlist-bind-port", 7946, "Port for memberlist gossip (default 7946)")
+	pflag.StringVar(&healthProbeBindAddress, "health-probe-bind-address", ":8081",
+		"The address the /healthz and /readyz endpoints bind to. Set to empty string to disable.")
 	pflag.StringVar(&e2bKeyStorage, "e2b-key-storage", "secret",
 		"Storage backend for E2B API keys. Valid values: 'secret' (K8s Secret, default), 'mysql' (MySQL via GORM). "+
 			"When --e2b-key-storage=mysql and auth is enabled, set MySQL DSN via environment variable "+E2BKeyStorageDSNEnvVar)
@@ -221,7 +224,7 @@ func main() {
 	}
 
 	sandboxController := e2b.NewController(domain, sysNs, peerSelector, sandboxNamespace, sandboxLabelSelector, e2bMaxTimeout, e2bMinResumeTimeout, maxClaimWorkers, maxCreateQPS, uint32(extProcMaxConcurrency),
-		port, memberlistBindPort, keyCfg, clientConfig)
+		port, memberlistBindPort, healthProbeBindAddress, keyCfg, clientConfig)
 
 	if err := sandboxController.Init(); err != nil {
 		klog.Fatalf("Failed to initialize sandbox controller: %v", err)
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -151,20 +151,31 @@ func NewControllerManager(cfg *rest.Config, opts config.SandboxManagerOptions) (
 		return nil, fmt.Errorf("failed to create controller manager: %w", err)
 	}
 
-	// Register basic /healthz and /readyz checks when the probe server is enabled.
-	// Empty HealthProbeBindAddress disables the probe server entirely, in which case
-	// AddHealthzCheck / AddReadyzCheck are no-ops semantically — we still register
-	// to fail fast if controller-runtime's contract changes.
-	if opts.HealthProbeBindAddress != "" {
-		if err := mgr.AddHealthzCheck("ping", healthz.Ping); err != nil {
-			return nil, fmt.Errorf("failed to add healthz check: %w", err)
-		}
-		if err := mgr.AddReadyzCheck("ping", healthz.Ping); err != nil {
-			return nil, fmt.Errorf("failed to add readyz check: %w", err)
-		}
+	if err := registerProbeChecks(mgr, opts.HealthProbeBindAddress); err != nil {
+		return nil, err
 	}
 
 	return mgr, nil
+}
+
+// registerProbeChecks registers basic /healthz and /readyz checks on mgr when
+// the probe server is enabled (i.e. addr is non-empty). When addr is empty the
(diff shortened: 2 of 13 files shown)


PRAteek-singHWY:
Hi @furykerry, @AiRanthem, and @zmberg
This PR has been rebased and conflicts resolved.

PR #358, opened 2026-05-09
fix(sandbox): make pause/resume concurrency-safe under parallel requests

### LINKED ISSUE #365: bug: paused sandbox timeout coordination is unstable under concurrent connect
### Background
Paused sandbox flows currently update timeout state across multiple non-atomic steps: pause can rewrite timeout, resume clears timeout before post-resume work completes, and E2B connect/resume updates timeout again after the sandbox has been resumed. Under concurrent or rapid repeated requests, this can leave timeout state inconsistent with the actual pause/resume cycle.

The same flow also relies on shared wait hooks for pause/resume completion. Concurrent waiters for the same sandbox/action can reuse one hook, but the current wait-hook cleanup deletes the hook when any waiter exits, so another waiter can lose the shared hook and wait until timeout or miss the intended deduplication behavior.

### Issue Type
bug

### Relevant Code
- `pkg/sandbox-manager/infra/sandboxcr/sandbox.go`: `Resume` flips `spec.paused=false` and clears timeout with `setTimeout(...{})` before waiting for resume and running post-resume initialization.
- `pkg/servers/e2b/pause_resume.go`: `ResumeSandbox` and `ConnectSandbox` read the pre-resume timeout, resume the sandbox, then call `SaveTimeout` afterward.
- `pkg/servers/e2b/pause_resume.go`: running sandboxes have extend-only timeout behavior, while paused-to-running flows apply timeout after resume, which leaves concurrent paused connect/resume requests without a stable pause-cycle boundary.
- `pkg/cache/utils/wait.go`: `WaitForObjectSatisfied` uses one wait hook per object key, reuses same-action waiters, and deletes the hook unconditionally in defer.

### Notes
- Related to #296, but this issue focuses on timeout persistence and wait-hook lifecycle rather than the broader pause/resume state race.
- Related to #304, but that issue tracks running `ConnectSandbox` shortening timeout; this issue covers paused-to-running timeout coordination.
- Affected component: `sandbox-manager` E2B-compatible pause/resume/connect APIs and cache wait utilities.

AiRanthem:
## Summary
Resolve concurrent Pause/Resume race conditions introduced by overlapping pause/resume requests on the same sandbox.

fixes #365

## What changed
- Fix pause/resume operations to always refresh the latest sandbox state from API reader before precondition checks and updates.
- Introduce timeout snapshot/policy-aware timeout update flow so paused-timeout and shutdown-time settings are updated safely under concurrent operations:
  - `pkg/sandbox-manager/infra/sandboxcr/sandbox.go`
  - `pkg/utils/timeout/timeout.go`
- Make `retryUpdate` refresh latest CR first and return whether the caller won the write lock update, avoiding duplicate writes.
- Make `Pause` / `Resume` first-writer-wins for `spec.paused` flips:
  - Skip timeout/snapshot side effects when another request already changed state.
  - Return early when resume loses update race before post-resume initialization.
- Add test coverage for stale wrapper/snapshot behavior and concurrent update races.
- Keep `Pause`/`Resume` behavior consistent across timeout, cache/controller wait, and e2b paths.

## Why this fixes concurrent Pause/Resume
When concurrent clients issue Pause/Resume, stale in-memory wrappers could still apply side effects after another request already changed the latest state. The new refresh-first + conditional update path ensures only one request performs state transitions and timeout/snapshot side effects, while others become no-op and return without redoing post-resume work.

## Testing
- `go test ./pkg/sandbox-manager/infra/sandboxcr -run TestSandbox_ -count=1`
- `go test ./pkg/cache/... -run Test -count=1`
- `go test ./pkg/servers/e2b -run Test -count=1`
- `go test ./pkg/sandbox-manager/... -run Test -count=1`
- `go test ./pkg/controller/sandbox -run Test -count=1`
- `go test ./pkg/controller/sandboxclaim -run Test -count=1`

## Notes
Branch: `feature/stable-timeout-pause-resume-260506`
Target base: `openkruise/agents:master`


pkg/sandbox-manager/infra/sandboxcr/sandbox.go
```diff
@@ -104,33 +110,57 @@ func (s *Sandbox) refreshFunc() runtime.RefreshFunc {
...
+	updated := false
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		// get the latest sandbox from cache
-		sbx, err := s.Cache.GetClaimedSandbox(ctx, cache.GetClaimedSandboxOptions{
-			SandboxID: stateutils.GetSandboxID(s.Sandbox),
-			Namespace: s.Namespace,
-		})
-		if err != nil {
+		latest := &agentsv1alpha1.Sandbox{}
+		if err := s.Cache.GetAPIReader().Get(ctx, client.ObjectKeyFromObject(s.Sandbox), latest); err != nil {
```
furykerry: consider pass the initial sbx value to the retryUpdate, and access apiserver directly only after update conflicts to reduce apiserver load. After all, in Resume and Pause, we had already refresh the sbx value directly from the apiserver

pkg/utils/timeout/timeout.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	"encoding/json"
+	"time"
+
+	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
+	"github.com/openkruise/agents/pkg/sandbox-manager/infra"
+)
+
+var jsonMarshalTimeoutOptions = json.Marshal
```
furykerry: The variable is never reassigned in tests, so it's unnecessary and adds complexity, plz remove it

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -253,6 +255,15 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (cr
 	return ctrl.Result{RequeueAfter: requeueAfter}, r.updateSandboxStatus(ctx, *newStatus, box)
 }
 
+func (r *SandboxReconciler) EnsureSandboxPaused(ctx context.Context, args core.EnsureFuncArgs) error {
+	// When in Paused state, always sync timeout state (shutdownTime, pauseTime) to annotations.
+	// This is an isolated operation and does not affect the main logic of the sandbox controller.
+	if err := r.ensurePauseTimeoutSnapshot(ctx, args.Box); err != nil {
```
furykerry: why introduce timeout snapshots in controller, the controller has a single master, only controller-runtime can ensure only one work is running for a single sandbox, so no concurrent race problem exists in the controller

AiRanthem: It is primarily to make sandbox-manager compatible with waking up from sleep triggered by non-E2B pathways.

pkg/utils/timeout/timeout.go
AiRanthem: It is overwritten in TestSetTimeoutSnapshot_MarshalError to mock an error

pkg/sandbox-manager/infra/sandboxcr/sandbox.go
AiRanthem: DONE

pkg/utils/timeout/timeout.go
```diff
@@ -0,0 +1,76 @@
...
+package timeout
+
+import (
+	"time"
+
+	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
+	"github.com/openkruise/agents/pkg/sandbox-manager/infra"
+)
+
+func GetTimeoutFromSandbox(sbx *agentsv1alpha1.Sandbox) infra.TimeoutOptions {
```
furykerry: consider move infra.TimeoutOptions  definition to this package, or move the timeout.go directly to infra/sanddboxcr package, this can reduce the circular-ish dependency risk

pkg/sandbox-manager/infra/sandboxcr/sandbox.go
```diff
@@ -192,25 +232,88 @@ func (s *Sandbox) GetImage() string {
...
+//   - BaselineAware: the caller provides the timeout data it observed before issuing
+//     this update. When current timeout matches baseline, it is treated as
+//     overwrite-allowed; when current timeout differs from baseline, only extension
+//     is allowed.
+func (s *Sandbox) SaveTimeoutWithPolicy(ctx context.Context, opts infra.TimeoutOptions, policy infra.TimeoutUpdatePolicy) (infra.TimeoutUpdateResult, error) {
+	log := klog.FromContext(ctx).V(consts.DebugLogLevel).WithValues("sandbox", klog.KObj(s.Sandbox), "policy", policy)
+	result := infra.TimeoutUpdateResult{}
+
+	first := true
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
furykerry: SaveTimeoutWithPolicy has its own retry loop that duplicates retryUpdate, consider refactoring SaveTimeoutWithPolicy to call retryUpdate with a ModifierFunc that embeds the policy logic,

furykerry (approved):
/lgtm
/approve

PR #359, opened 2026-05-09
Validate SandboxSet volume claim template mounts

ajatshatru01:
This PR adds validation for `SandboxSet` resources that define `volumeClaimTemplates`.

Previously, the SandboxSet validating webhook skipped pod template validation whenever `volumeClaimTemplates` was non-empty. That meant invalid pod template configurations could pass admission, including cases where a PVC template was defined but never mounted by the pod.

## Changes

- Continue validating the `PodTemplateSpec` even when `spec.volumeClaimTemplates` is present.
- Build the effective pod template by adding generated PVC-backed volumes for each `VolumeClaimTemplate`.
- Validate that every named `VolumeClaimTemplate` is mounted by at least one:
  - container
  - init container
  - ephemeral container
- Add webhook tests for:
  - valid SandboxSet with a mounted `VolumeClaimTemplate`
  - invalid SandboxSet with an unmounted `VolumeClaimTemplate`
  - invalid SandboxSet with a mismatched volume mount name
  
## Behavior

A `SandboxSet` with:

volumeClaimTemplates:
  - metadata:
      name: data

must include a matching volume mount such as:

containers:
  - name: app
    volumeMounts:
      - name: data
        mountPath: /data

If the claim template is not mounted, admission returns a validation error.

"NONE"

But solves the TODO at sandboxset_create_update.go (line 129) "//TODO: validate the podtemplate if VolumeClaimTemplates is not empty
"

go test ./pkg/webhook/sandboxset/validating

I have ran this test on my local machine 
image

### DIFF
--- a/pkg/webhook/sandboxset/validating/sandboxset_create_update.go
+++ b/pkg/webhook/sandboxset/validating/sandboxset_create_update.go
@@ -23,12 +23,13 @@ import (
 	"net/http"
 	"strings"
 
+	apicorev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	intstrutil "k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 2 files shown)


furykerry (approved):
/lgtm
/approve

PR #368, opened 2026-05-13
fix: include initContainers in resize body to prevent inplace VPA fai…

zmberg:
…lure


pkg/utils/inplaceupdate/inplace_update.go
```diff
@@ -472,10 +462,16 @@ func (c *InPlaceUpdateControl) Update(ctx context.Context, opts InPlaceUpdateOpt
...
 			return err
 		}); err != nil {
-			return progressed, err
+			return false, err
 		}
-		if opts.OnProgress != nil {
-			opts.OnProgress()
+	}
+
+	// perform patch
```
furykerry: plz explain why the order matters, and what problem may occurs.

PR #380, opened 2026-05-14
fix: include stdout/stderr in lifecycle hook failure messages

zmberg:
# Background
When a lifecycle hook (preUpgrade/postUpgrade) fails, the controller only logs:
`"hook execution error: process error: exit status 1"`
Users have no way to know why the hook failed without SSH-ing into the sandbox pod — the command's stdout/stderr is silently discarded.

# Changes

- Fix stdout/stderr loss on error path — ExecuteLifecycleHook now returns already-captured stdout/stderr even when RunCommandWithRuntime returns an error.
- Enrich error messages — executeUpgradeAction appends stderr and stdout to the Condition message, giving users immediate visibility into hook failures.
- Configurable message truncation — MaxConditionMessageLen is now a variable (default 1024) configurable via MAX_CONDITION_MESSAGE_LEN env var, up from the previous hardcoded 256.


furykerry (approved):
/lgtm
/approve

PR #385, opened 2026-05-14
feat(e2b): support naming cloned sandbox via metadata extensions

AiRanthem:
## Summary

- Adds two new metadata keys for `POST /sandboxes` on the clone path:
  - `e2b.agents.kruise.io/sandbox-name` — sets `ObjectMeta.Name` (mutually exclusive with the next).
  - `e2b.agents.kruise.io/sandbox-generate-name` — sets `ObjectMeta.GenerateName` prefix.
- Default behavior is unchanged when neither key is provided (`GenerateName: <checkpointID>-`).
- Validates inputs as DNS-1123 label (`apivalidation.NameIsDNSLabel`); the canonical K8s validator with `prefix=true` masks the trailing `-` for `GenerateName` prefixes. Both keys are stripped from `Metadata` so they don't leak onto sandbox annotations.
- AlreadyExists collision (only reachable when `sandbox-name` is set) maps to `409 Conflict` via the manager's typed `ErrorConflict`.
- Naming is a first-class field on `infra.CloneSandboxOptions`; `newSandboxFromTemplate` uses an explicit `switch` so the Name/GenerateName precedence is visible at the call site rather than relying on K8s semantics.

### Public-contract note (beyond the design spec)

The claim path actively rejects these keys with `400 Bad Request` (`sandbox-name and sandbox-generate-name are only supported for clone`). Caller-controlled naming on the claim path is owned by `SandboxSet` and is out of scope; failing fast is preferable to silently dropping the request.

### Design

`docs/superpowers/specs/2026-05-12-clone-sandbox-name-design.md` (in the design branch; not included in this PR).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./pkg/servers/e2b/... ./pkg/sandbox-manager/...`
- [x] `go test ./pkg/servers/e2b/... ./pkg/sandbox-manager/...`
- [x] `parseExtensionSandboxNaming` table covers mutual exclusion, empty values, dot rejection, trailing-dash prefix, and key-stripping.
- [x] `ValidateAndInitCloneOptions` rejects both Name and GenerateName set together.
- [x] `newSandboxFromTemplate` honors Name, GenerateName, and falls back to `<checkpointID>-`.
- [x] `TestCreateSandboxWithClone_Naming` end-to-end: `Extensions.Name` / `Extensions.GenerateName` flow through to the resulting Sandbox `ObjectMeta`.
- [x] `TestSandboxManager_CloneSandbox` "explicit name collision returns conflict" maps to `ErrorConflict`.
- [x] `TestMapCloneSandboxError` confirms only manager-wrapped conflict maps to `409` (raw `apierrors.AlreadyExists` is not specially mapped — manager is expected to wrap).
- [x] `TestCreateSandboxWithClaim_NamingExtensionRejected` confirms claim path returns `400`.

### DIFF
--- a/pkg/servers/e2b/models/extensions.go
+++ b/pkg/servers/e2b/models/extensions.go
@@ -73,6 +73,8 @@ const (
 	ExtensionHeaderWaitSuccessSeconds         = ExtensionHeaderPrefix + "snapshot-wait-success-seconds"
 )
 
+const sandboxGenerateNameValidationSuffix = "abcde"
+
 // Extensions for NewSandboxRequest
 
 func (r *NewSandboxRequest) ParseExtensions() error {
@@ -159,9 +161,9 @@ func (r *NewSandboxRequest) parseExtensionLabels() error {
 //
 // We validate as DNS-1123 label (63 chars, no dots) rather than subdomain:
 // the sandbox name becomes the underlying Pod name and is used as a DNS
-// hostname downstream, where label semantics apply. NameIsDNSLabel with
-// prefix=true is the canonical K8s validator for GenerateName prefixes —
-// it masks the required trailing dash before validating the rest.
+// hostname downstream, where label semantics apply. GenerateName is also
+// checked against a representative generated name so malformed trailing-dash
+// prefixes fail before the request reaches Kubernetes.
 func (r *NewSandboxRequest) parseExtensionSandboxNaming() error {
 	name, hasName := r.Metadata[ExtensionKeySandboxName]
 	gen, hasGen := r.Metadata[ExtensionKeySandboxGenerateName]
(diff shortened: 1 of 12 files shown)


pkg/servers/e2b/models/extensions.go
```diff
@@ -70,6 +73,8 @@ const (
 	ExtensionHeaderWaitSuccessSeconds         = ExtensionHeaderPrefix + "snapshot-wait-success-seconds"
 )
 
+const sandboxGenerateNameValidationSuffix = "abcde"
```
furykerry, line 76: this suffix is weird, if user had specify generate name, i don't think we should inject any suffix

pkg/servers/e2b/models/extensions_test.go
```diff
@@ -748,6 +748,114 @@ func TestParseExtensionInplaceUpdate(t *testing.T) {
...
+			metadata:    map[string]string{v1alpha1.E2BPrefix + "sandbox-generate-name": "a.-"},
+			expectError: "invalid sandbox-generate-name",
+		},
+		{
+			name:               "trailing dash ok",
+			metadata:           map[string]string{v1alpha1.E2BPrefix + "sandbox-generate-name": "good-prefix-"},
+			expectGenerateName: "good-prefix-",
+		},
+		{
+			name:       "keys stripped from metadata",
```
furykerry, line 828: Missing coverage for validateSandboxGenerateName truncation at max prefix length

pkg/servers/e2b/models/extensions.go
AiRanthem: It is a trick to validate the given generate name. The suffix will not be injected.

PR #387, opened 2026-05-14
fix: remove UnsafeDisableDeepCopy in groupAllSandboxes to prevent informer cache corruption

### LINKED ISSUE #386: bug: UnsafeDisableDeepCopy in groupAllSandboxes corrupts informer cache during scale-down
## Summary

groupAllSandboxes in pkg/controller/sandboxset/sandboxset_controller.go lists sandboxes with client.UnsafeDisableDeepCopy, which returns raw informer cache pointers. Downstream functions (scaleDownSandbox, deleteSandboxForUpdate) mutate annotations on these pointers via LockSandbox, silently corrupting the shared informer cache.

## Root Cause

The cache is configured with DefaultUnsafeDisableDeepCopy: true (see pkg/cache/cache.go:144). When List is called with client.UnsafeDisableDeepCopy, the returned sandboxList.Items[i] are aliases into the informer's internal store. Any in-place mutation (e.g., LockSandbox writing AnnotationLock and AnnotationOwner) modifies the cached object directly, causing:

1. **Phantom state**: Other reconcile loops or event handlers see stale/corrupted annotation values.
2. **Non-deterministic scale-down**: The wrong sandboxes may be selected for deletion because their lock annotations appear pre-set.
3. **Expectation mismatch**: ResourceVersion expectations computed from corrupted objects diverge from the API server state, blocking reconciliation for up to 30 seconds.

Additionally, scaleDown at line 249 uses ppend(groups.Creating, groups.Available...), which is a Go slice aliasing trap — when Creating has spare backing-array capacity, Available elements silently overwrite memory belonging to Creating.

## Impact

- **Severity**: Critical
- **Affected component**: SandboxSet controller (warm pool management)
- **Trigger condition**: Concurrent scale-down or rolling update while sandboxes are being created/claimed
- **Blast radius**: Corrupted cache affects all SandboxSet reconcile loops in the same controller process

## Fix

1. Remove client.UnsafeDisableDeepCopy from the groupAllSandboxes List call so returned objects are deep-copied and safe to mutate.
2. Replace ppend(groups.Creating, groups.Available...) with explicit make + two appends to prevent slice aliasing.

## Reproduction

This bug is non-deterministic and only manifests under concurrent load when:
- groups.Creating has spare backing-array capacity (from a previous reconcile with more creating sandboxes)
- Multiple reconcile loops or claim requests run simultaneously against the same SandboxSet

### LINKED ISSUE #388: bug: slice aliasing in scaleDown via append(groups.Creating, groups.Available...) may delete wrong sandboxes
## Summary

scaleDown in pkg/controller/sandboxset/sandboxset_controller.go uses a Go slice aliasing trap that can silently cause the wrong sandboxes to be selected for deletion.

## Root Cause

`go
// Line 249 (before fix)
candidates := append(groups.Creating, groups.Available...)
`

In Go, ppend reuses the backing array of the first slice when it has spare capacity. groups.Creating is a slice of pointers built from sandboxList.Items — its backing array is sized by the runtime and often has spare capacity from the previous reconcile cycle.

When groups.Creating has spare capacity, ppend writes groups.Available elements directly into groups.Creating's backing array, overwriting memory that groups.Creating still logically owns. This means:

1. groups.Creating[len(creating)...] gets silently overwritten with groups.Available elements
2. Any caller that still holds a reference to groups.Creating's underlying array sees corrupted data
3. The candidates slice contains mixed/incorrect entries, so scaleDown may delete **available sandboxes** that should remain, or **skip creating sandboxes** that should be deleted

## Impact

- **Severity**: Critical (same PR as openkruise/agents#387)
- **Affected component**: SandboxSet controller — warm pool scale-down
- **Trigger condition**: Non-deterministic; occurs when len(groups.Creating) < cap(groups.Creating), which happens when the Creating count has recently decreased (e.g. sandboxes becoming Available between reconcile cycles)
- **Consequence**: Wrong sandbox selection for deletion — may delete in-use or available sandboxes while leaving stale creating ones

## Fix

Replace the aliasing append with a safe pre-allocated slice:

`go
// Before (buggy)
candidates := append(groups.Creating, groups.Available...)

// After (safe)
candidates := make([]*agentsv1alpha1.Sandbox, 0, len(groups.Creating)+len(groups.Available))
candidates = append(candidates, groups.Creating...)
candidates = append(candidates, groups.Available...)
`

## Resolution

Fixed in #387 alongside the related UnsafeDisableDeepCopy cache corruption bug.

oindrilakha12-ui:
## What

Remove client.UnsafeDisableDeepCopy from groupAllSandboxes and fix slice aliasing in scaleDown to prevent informer cache corruption during concurrent scale-down and rolling update operations.

## Why

1. **Cache Corruption**: groupAllSandboxes lists sandboxes with client.UnsafeDisableDeepCopy, returning raw informer cache pointers. Downstream code mutates these pointers, silently corrupting the shared informer cache.
2. **Slice Aliasing**: scaleDown used ppend(groups.Creating, groups.Available...) which is a Go slice aliasing trap — when Creating has spare capacity, Available elements silently overwrite its memory.

## Changes

- Removed client.UnsafeDisableDeepCopy from the groupAllSandboxes List call.
- Fixed slice aliasing by replacing the buggy ppend with explicit make + dual appends.

Fixes #386
Fixes #388

### DIFF
--- a/pkg/controller/sandboxset/rolling_update.go
+++ b/pkg/controller/sandboxset/rolling_update.go
@@ -241,7 +241,7 @@ func (r *Reconciler) deleteSandboxForUpdate(ctx context.Context, sbs *agentsv1al
 
 	// Deep copy the sandbox before mutating it to avoid corrupting the informer cache.
 	sbx = sbx.DeepCopy()
-	
+
 	managerutils.LockSandbox(sbx, lock, consts.OwnerManagerScaleDown)
 	if err := r.Update(ctx, sbx); err != nil {
 		return fmt.Errorf("failed to lock sandbox when delete: %s", err)
--- a/pkg/controller/sandboxset/rolling_update.go
+++ b/pkg/controller/sandboxset/rolling_update.go
@@ -239,6 +239,9 @@ func (r *Reconciler) deleteSandboxForUpdate(ctx context.Context, sbs *agentsv1al
 		return errors.New("sandbox to be deleted claimed before performed, skip")
 	}
 
+	// Deep copy the sandbox before mutating it to avoid corrupting the informer cache.
+	sbx = sbx.DeepCopy()
+	
 	managerutils.LockSandbox(sbx, lock, consts.OwnerManagerScaleDown)
 	if err := r.Update(ctx, sbx); err != nil {
 		return fmt.Errorf("failed to lock sandbox when delete: %s", err)
--- a/pkg/controller/sandboxset/sandboxset_controller.go
+++ b/pkg/controller/sandboxset/sandboxset_controller.go
@@ -365,6 +365,8 @@ func (r *Reconciler) scaleDownSandbox(ctx context.Context, sbx *agentsv1alpha1.S
 		log.Info("sandbox to be scaled down claimed before performed, skip")
 		return errors.New("sandbox to be scaled down claimed before performed, skip")
 	}
+	// Deep copy the sandbox before mutating it to avoid corrupting the informer cache.
+	sbx = sbx.DeepCopy()
 	managerutils.LockSandbox(sbx, lock, consts.OwnerManagerScaleDown)
 	if err = r.Update(ctx, sbx); err != nil {
 		return fmt.Errorf("failed to lock sandbox when scaling down: %s", err)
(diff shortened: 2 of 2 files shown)


oindrilakha12-ui:
@furykerry , please check it

furykerry:
@oindrilakha12-ui thanks for your contribution, however deepcopy is a very heavy operation, we need UnsafeDisableDeepCopy, for the operation that may mutate informer cache, I prefer to delay the deepcopy to just before the operation e.g. LockSandbox

oindrilakha12-ui:
@furykerry ,Thanks for the review and the insight.That makes perfect sense. Retrieving the full list without UnsafeDisableDeepCopy is definitely too heavy when we only end up mutating a small subset of the sandboxes during a scale-down or rolling update.
I will update the PR with the following approach:
1. Revert the change in groupAllSandboxes to keep client.UnsafeDisableDeepCopy in the r.List() call for performance.
2. Add a targeted sbx = sbx.DeepCopy() immediately before managerutils.LockSandbox in both scaleDownSandbox and deleteSandboxForUpdate. This will delay the deep copy and isolate it strictly to the objects about to be mutated, successfully preventing the informer cache corruption without the heavy memory overhead.

furykerry:
@oindrilakha12-ui plz fix lint and ut tests.

oindrilakha12-ui:
@furykerry ,please check it . I have fixed the linting issues . All CI checks, including the unit tests, are now passing successfully.

oindrilakha12-ui:
@furykerry ,please check it

PR #396, opened 2026-05-14
feat(controller): auto create sandboxtemplate for sandboxset

BITLiutianyang:
SandboxSet now auto create/materialises its inline spec.template into a first-class SandboxTemplate CR on each revision, so operators can kubectl get sbt to inspect the currently-bound pod spec.

Related to #256.

- Case 1: apply a SandboxSet with inline spec.template, replicas=1; expect status.currentTemplate != "", SBT <sbs>-<hash> exists with controller=true, blockOwnerDeletion=false.
- Case 2: pre-create SBT my-tpl, apply a SandboxSet with spec.templateRef.name=my-tpl; expect status.currentTemplate=="my-tpl" and zero SBTs owned by this SandboxSet.
- Case 3: patch spec.template; expect status.currentTemplate to change to a new SBT, and the old SBT to remain (in-use by the existing sandbox).
- Case 4: with replicas=0 (no in-use SBTs), patch spec.template 11 times; expect owned SBTs ≤10 and the oldest one gone.
- Case 5: delete the SandboxSet; expect every previously-owned SBT to become NotFound.

### DIFF
--- a/config/crd/bases/agents.kruise.io_sandboxes.yaml
+++ b/config/crd/bases/agents.kruise.io_sandboxes.yaml
@@ -262,12 +262,6 @@ spec:
                 x-kubernetes-list-map-keys:
                 - type
                 x-kubernetes-list-type: map
-              currentRevision:
-                description: |-
-                  CurrentRevision mirrors Labels[agents.kruise.io/template-hash] so that the
-                  bound revision hash is visible via status. Purely informational; it does
-                  not drive any reconcile decision.
-                type: string
               message:
                 description: message
                 type: string
--- a/config/crd/bases/agents.kruise.io_sandboxes.yaml
+++ b/config/crd/bases/agents.kruise.io_sandboxes.yaml
@@ -262,6 +262,12 @@ spec:
                 x-kubernetes-list-map-keys:
                 - type
                 x-kubernetes-list-type: map
+              currentRevision:
+                description: |-
+                  CurrentRevision mirrors Labels[agents.kruise.io/template-hash] so that the
+                  bound revision hash is visible via status. Purely informational; it does
+                  not drive any reconcile decision.
+                type: string
               message:
                 description: message
                 type: string
--- a/pkg/controller/sandboxset/revision.go
+++ b/pkg/controller/sandboxset/revision.go
@@ -14,141 +14,64 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Copied From Kubernetes
-
 package sandboxset
 
 import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"hash"
 	"hash/fnv"
-	"strconv"
 
-	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
-	apps "k8s.io/api/apps/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/util/dump"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
 )
 
-// getPatch serialises the subset of the SandboxSet spec that determines the
-// template revision hash. Both spec.template (inline) and spec.templateRef are
-// normalised to the same `spec.template` key so that they yield identical
-// hashes whenever they describe the same pod template. This also guarantees
-// that switching between inline template and templateRef (or between two
-// templateRefs) only triggers a rolling update when the underlying pod
-// template actually differs.
-func (r *Reconciler) getPatch(ctx context.Context, set *agentsv1alpha1.SandboxSet) ([]byte, error) {
-	specCopy := make(map[string]interface{})
-	// spec.template and spec.templateRef are mutually exclusive; prefer
-	// templateRef when set and fall back to the inline template otherwise.
-	if set.Spec.TemplateRef != nil {
+// buildSandboxTemplateSpec constructs the effective SandboxTemplateSpec from a
+// SandboxSet, handling both inline template and templateRef cases.
+// WARNING: the returned spec shares slice fields with sbs.Spec; callers must
+// not mutate VolumeClaimTemplates, PersistentContents or Runtimes.
+func (r *Reconciler) buildSandboxTemplateSpec(ctx context.Context, sbs *agentsv1alpha1.SandboxSet) (*agentsv1alpha1.SandboxTemplateSpec, error) {
+	if sbs.Spec.TemplateRef != nil {
 		tpl := &agentsv1alpha1.SandboxTemplate{}
 		if err := r.Get(ctx, client.ObjectKey{
-			Namespace: set.Namespace,
-			Name:      set.Spec.TemplateRef.Name,
+			Namespace: sbs.Namespace,
+			Name:      sbs.Spec.TemplateRef.Name,
 		}, tpl); err != nil {
 			return nil, fmt.Errorf("failed to resolve sandbox template %s/%s: %w",
-				set.Namespace, set.Spec.TemplateRef.Name, err)
+				sbs.Namespace, sbs.Spec.TemplateRef.Name, err)
 		}
-		if tpl.Spec.Template != nil {
-			templateRaw, err := json.Marshal(tpl.Spec.Template)
-			if err != nil {
-				return nil, err
-			}
-			var templateMap map[string]interface{}
-			if err := json.Unmarshal(templateRaw, &templateMap); err != nil {
-				return nil, err
-			}
-			templateMap["$patch"] = "replace"
-			specCopy["template"] = templateMap
-		}
-	} else if set.Spec.Template != nil {
-		str, err := runtime.Encode(r.Codec, set)
-		if err != nil {
-			return nil, err
-		}
-		var raw map[string]interface{}
-		if err = json.Unmarshal(str, &raw); err != nil {
-			return nil, err
-		}
-		if spec, ok := raw["spec"].(map[string]interface{}); ok {
-			if template, ok := spec["template"].(map[string]interface{}); ok {
-				template["$patch"] = "replace"
-				specCopy["template"] = template
-			}
-		}
+		return &agentsv1alpha1.SandboxTemplateSpec{
+			Template:             tpl.Spec.Template.DeepCopy(),
+			VolumeClaimTemplates: sbs.Spec.VolumeClaimTemplates,
+			PersistentContents:   sbs.Spec.PersistentContents,
+			Runtimes:             sbs.Spec.Runtimes,
+		}, nil
 	}
-	return json.Marshal(map[string]interface{}{"spec": specCopy})
+
+	if sbs.Spec.Template == nil {
+		return nil, fmt.Errorf("sandboxset %s/%s has neither spec.templateRef nor spec.template", sbs.Namespace, sbs.Name)
+	}
+
+	return &agentsv1alpha1.SandboxTemplateSpec{
+		Template:             sbs.Spec.Template.DeepCopy(),
+		VolumeClaimTemplates: sbs.Spec.VolumeClaimTemplates,
+		PersistentContents:   sbs.Spec.PersistentContents,
+		Runtimes:             sbs.Spec.Runtimes,
+	}, nil
 }
 
-// NewControllerRevision returns a ControllerRevision with a ControllerRef pointing to parent and indicating that
-// parent is of parentKind. The ControllerRevision has labels matching template labels, contains Data equal to data, and
-// has a Revision equal to revision. The collisionCount is used when creating the name of the ControllerRevision
-// so the name is likely unique. If the returned error is nil, the returned ControllerRevision is valid. If the
-// returned error is not nil, the returned ControllerRevision is invalid for use.
-func NewControllerRevision(parent metav1.Object,
-	parentKind schema.GroupVersionKind,
-	templateLabels map[string]string,
-	data runtime.RawExtension,
-	revision int64,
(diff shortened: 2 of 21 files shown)


zmberg (commented):
d

config/crd/bases/agents.kruise.io_sandboxes.yaml
```diff
@@ -262,6 +262,12 @@ spec:
                 x-kubernetes-list-map-keys:
                 - type
                 x-kubernetes-list-type: map
+              currentRevision:
```
zmberg: This `currentRevision` field appears in the CRD diff, but there is no corresponding Go field in `api/v1alpha1/sandbox_types.go`. It looks like this was not generated by `make generate manifests` from this PR's code — possibly dirty data carried over from another branch. Please verify: either remove it from the CRD yaml, or add the matching Go type definition in `sandbox_types.go`.

pkg/controller/sandboxset/sandboxset_controller.go
```diff
@@ -213,6 +212,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	log.Info("reconcile done", "totalCost", time.Since(totalStart))
+	r.cleanupOldSandboxTemplates(ctx, sbs)
```
zmberg: `cleanupOldSandboxTemplates` is called before `updateSandboxSetStatus`. If cleanup deletes old SBTs but the subsequent status update fails, the next reconcile will re-run `ensureTemplateRevision` and potentially re-create the SBT that was just deleted. While idempotent and not a bug, it does wasted work. Consider moving cleanup after a successful status update.

zmberg (commented):
Overall the approach is clean — replacing ControllerRevision with auto-materialised SandboxTemplate CRs is a good simplification. A few issues to address below.

pkg/controller/sandboxset/revision.go
```diff
@@ -14,141 +14,65 @@ See the License for the specific language governing permissions and
...
-		}
-		if spec, ok := raw["spec"].(map[string]interface{}); ok {
-			if template, ok := spec["template"].(map[string]interface{}); ok {
-				template["$patch"] = "replace"
-				specCopy["template"] = template
-			}
+				sbs.Namespace, sbs.Spec.TemplateRef.Name, err)
 		}
+		return &agentsv1alpha1.SandboxTemplateSpec{
+			Template:             tpl.Spec.Template.DeepCopy(),
```
zmberg, line 46: Could just `return &tpl.Spec` directly here. The current code takes `tpl.Spec.Template` but combines it with `sbs.Spec.VolumeClaimTemplates/PersistentContents/Runtimes`, which means those fields from the SandboxTemplate itself are silently ignored. Is that intentional? If the SandboxTemplate also defines its own Runtimes, should they be used, or merged with the SandboxSet's?

pkg/controller/sandboxset/revision.go
```diff
@@ -14,141 +14,65 @@ See the License for the specific language governing permissions and
...
-	}
-	if probe != nil {
-		hf.Write([]byte(strconv.FormatInt(int64(*probe), 10)))
-	}
-	return rand.SafeEncodeString(fmt.Sprint(hf.Sum32()))
+	return &agentsv1alpha1.SandboxTemplateSpec{
+		Template:             sbs.Spec.Template.DeepCopy(),
+		VolumeClaimTemplates: deepCopyVCTs(sbs.Spec.VolumeClaimTemplates),
+		PersistentContents:   append([]string(nil), sbs.Spec.PersistentContents...),
+		Runtimes:             append([]agentsv1alpha1.RuntimeConfig(nil), sbs.Spec.Runtimes...),
```
zmberg: `VolumeClaimTemplates` uses a custom `deepCopyVCTs` (per-element DeepCopy), while `Runtimes` and `PersistentContents` only get a shallow copy via `append`. Shallow copy is sufficient here since the output is only used for hash computation and SBT creation (read-only path). Suggest removing `deepCopyVCTs` and using a plain shallow copy for consistency.

pkg/controller/sandboxset/template_materialize.go
```diff
@@ -0,0 +1,209 @@
...
+		return "", nil
+	}
+	name := fmt.Sprintf("%s-%s", sbs.Name, hash)
+	sbt := &agentsv1alpha1.SandboxTemplate{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: sbs.Namespace,
+			Name:      name,
+		},
+		Spec: *spec,
+	}
```
zmberg, line 76: `*spec` is a value copy, but the internal `Template` pointer still points to the same object. Consider using `spec.DeepCopy()` for defensive copying.

pkg/controller/sandboxset/template_materialize.go
```diff
@@ -0,0 +1,209 @@
...
+// auto-materialised SandboxTemplates per SandboxSet. SBTs still referenced
+// by any Sandbox via Labels[LabelTemplateHash] are always preserved; the
+// remaining (orphan) templates are sorted by CreationTimestamp and the
+// oldest entries are deleted until the limit is honoured. Errors are
+// logged and swallowed so that cleanup never blocks the main reconcile.
+func (r *Reconciler) cleanupOldSandboxTemplates(ctx context.Context, sbs *agentsv1alpha1.SandboxSet) {
+	log := logf.FromContext(ctx)
+	sbtList := &agentsv1alpha1.SandboxTemplateList{}
+	if err := r.List(ctx, sbtList, client.InNamespace(sbs.Namespace)); err != nil {
+		log.Error(err, "failed to list sandbox templates for cleanup")
```
zmberg, line 119: Cleanup logic in production can be tricky — SandboxTemplates may be referenced by Checkpoints or other components in the future. Suggest guarding this behind a feature gate (default off) until the full usage patterns of SandboxTemplate are finalized.

pkg/controller/sandboxset/template_materialize.go
```diff
@@ -0,0 +1,209 @@
...
+		return
+	}
+	owned := make([]*agentsv1alpha1.SandboxTemplate, 0, len(sbtList.Items))
+	for i := range sbtList.Items {
+		sbt := &sbtList.Items[i]
+		if !isControlledBy(sbt, sbs) {
+			continue
+		}
```
zmberg: This lists all SandboxTemplates in the namespace and filters in memory. Consider using a field index for better performance:
```go
client.InNamespace(sbs.Namespace),
client.MatchingFields{fieldindex.IndexNameForOwnerRefUID: string(sbs.UID)},
client.UnsafeDisableDeepCopy,
```

pkg/controller/sandboxset/template_materialize.go
```diff
@@ -0,0 +1,209 @@
...
+		log.Error(err, "failed to collect in-use template names")
+		return
+	}
+	// Orphan = owned by the sbs but not referenced by any live sandbox.
+	// Preserved = the rest.
+	var orphans []*agentsv1alpha1.SandboxTemplate
+	preservedCount := 0
+	for _, sbt := range owned {
+		if inUse[sbt.Name] {
+			preservedCount++
```
zmberg: The exact usage patterns of SandboxTemplate are not fully settled yet (e.g. potential references from Checkpoint). Suggest removing `collectInUseTemplateNames` for now and keeping cleanup simple — just enforce `defaultTemplateHistoryLimit` on owned SBTs without the in-use/orphan distinction.

pkg/controller/sandboxset/template_materialize_test.go
```diff
@@ -0,0 +1,390 @@
...
+	"github.com/openkruise/agents/api/v1alpha1"
+)
+
+// newMaterializeReconciler wires a minimal Reconciler backed by a fake client
+// with SandboxTemplate registered for ensure / cleanup tests.
+func newMaterializeReconciler(objs ...client.Object) *Reconciler {
+	c := fake.NewClientBuilder().
+		WithScheme(testScheme).
+		WithObjects(objs...).
+		WithLists(&v1alpha1.SandboxTemplateList{}, &v1alpha1.SandboxList{}).
```
zmberg: `WithLists` is not necessary here — the fake client automatically supports List for any type registered in the scheme. Can be removed to simplify.

pkg/controller/sandboxset/template_materialize.go
BITLiutianyang: The `buildSandboxTemplateSpec` has already called `.DeepCopy()` at lines 47/59. By the time `ensureSandboxTemplate` is reached, `spec.Template` already points to a brand new deep-copied object.

pkg/controller/sandboxset/revision.go
BITLiutianyang: This is intentional design. When using templateRef, SandboxTemplate only provides Pod templates (Template field), while `VolumeClaimTemplates/PersistentContents/Runtimes` always come from the SandboxSet.

This allows the same SandboxTemplate to be reused by different SandboxSets with different operational parameters. Returning `&tpl.Spec` directly would ignore the operational fields of SandboxSet, undermining this design intent.

pkg/controller/sandboxset/template_materialize.go
```diff
@@ -0,0 +1,132 @@
...
+	if sbs.Spec.Template == nil {
+		return "", nil
+	}
+	name := fmt.Sprintf("%s-%s", sbs.Name, hash)
+	sbt := &agentsv1alpha1.SandboxTemplate{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: sbs.Namespace,
+			Name:      name,
+		},
+		Spec: *spec,
```
furykerry, line 75: plz comment why shallow copy is safe here

pkg/controller/sandboxset/template_materialize.go
```diff
@@ -0,0 +1,132 @@
...
+	if len(sbtList.Items) <= defaultTemplateHistoryLimit {
+		return
+	}
+	// Sort oldest first by CreationTimestamp.
+	sort.SliceStable(sbtList.Items, func(i, j int) bool {
+		return sbtList.Items[i].CreationTimestamp.Before(&sbtList.Items[j].CreationTimestamp)
+	})
+	toDelete := sbtList.Items[:len(sbtList.Items)-defaultTemplateHistoryLimit]
+	for i := range toDelete {
+		if err := r.Delete(ctx, &toDelete[i]); err != nil && !apierrors.IsNotFound(err) {
```
furykerry: maybe we can just annotate the template for cleanup candiate, and create an extra template gc controller to check for reference sandbox and checkpoint

pkg/controller/sandboxset/revision.go
```diff
@@ -14,141 +14,67 @@ See the License for the specific language governing permissions and
...
-				template["$patch"] = "replace"
-				specCopy["template"] = template
-			}
+				sbs.Namespace, sbs.Spec.TemplateRef.Name, err)
 		}
+		return &agentsv1alpha1.SandboxTemplateSpec{
+			Template:             tpl.Spec.Template.DeepCopy(),
+			VolumeClaimTemplates: append([]corev1.PersistentVolumeClaim(nil), sbs.Spec.VolumeClaimTemplates...),
+			PersistentContents:   append([]string(nil), sbs.Spec.PersistentContents...),
+			Runtimes:             append([]agentsv1alpha1.RuntimeConfig(nil), sbs.Spec.Runtimes...),
```
furykerry: is it also safe to assign  sbs.Spec.VolumeClaimTemplates etc to VolumeClaimTemplates directly without any shallow copy? operation after buildSandboxTemplateSpec does not mutate template spec at all, we can add a comment to warn about mutate VolumeClaimTemplates,PersistentContents,Runtimes

pkg/controller/sandboxset/template_materialize.go
BITLiutianyang: According to [zmberg's comment](https://github.com/openkruise/agents/pull/396#discussion_r3323001333), The current cleanupOldSandboxTemplates is already gated behind SandboxTemplateCleanup (default off), so no deletion happens until the gate is explicitly enabled. Introducing an annotation-based marking scheme plus a standalone GC controller is a meaningful architectural change — it would require a new reconciler, reference-checking logic for both Sandbox and Checkpoint, and its own RBAC/lifecycle. That scope goes beyond this PR. I'd prefer to land the current conservative approach first and revisit the GC controller design as a follow-up once the SandboxTemplate usage patterns (especially Checkpoint references) are more settled.

pkg/controller/sandboxset/template_materialize.go
```diff
@@ -0,0 +1,135 @@
...
+	// BlockOwnerDeletion so deleting the SandboxSet is never blocked by the
+	// SBT finalizers.
+	for i := range sbt.OwnerReferences {
+		if sbt.OwnerReferences[i].UID == sbs.UID {
+			sbt.OwnerReferences[i].BlockOwnerDeletion = ptr.To(false)
+		}
+	}
+	if err := r.Create(ctx, sbt); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return name, nil
```
furykerry, line 99: When a SBT with the same name already exists, the code assumes it's identical and returns success. However, if someone manually created a SBT with that name but different content, or if a hash collision occurred, the controller would silently use the wrong template.

Recommendation: On IsAlreadyExists, perform a Get and verify the spec matches (or at minimum, verify the owner reference points to this SandboxSet). If mismatched, return an error or log a warning:

pkg/controller/sandboxset/revision.go
```diff
@@ -14,141 +14,64 @@ See the License for the specific language governing permissions and
...
-	}
-	cr := &apps.ControllerRevision{
-		ObjectMeta: metav1.ObjectMeta{
-			Labels:          labelMap,
-			OwnerReferences: []metav1.OwnerReference{*metav1.NewControllerRef(parent, parentKind)},
-		},
-		Data:     data,
-		Revision: revision,
+	if sbs.Spec.Template == nil {
+		return &agentsv1alpha1.SandboxTemplateSpec{}, nil
```
furykerry: return empty SandboxTemplateSpec is meaningless, sandboxset without template and templateRef is not suppose to exist, consider just log an error

PR #397, opened 2026-05-15
add support for egress control injection

Kuromesi:

Add egress-control sidecar injection with automatic health probe rewriting via a new controller-runtime managed Injector, replacing the old inline injection path.

Support strategic merge patch runtime injection, allow us to inject with more flexibility and extensibility.

### DIFF
--- a/pkg/controller/sandbox/core/common_control.go
+++ b/pkg/controller/sandbox/core/common_control.go
@@ -436,8 +436,7 @@ func (r *commonControl) createPod(ctx context.Context, box *agentsv1alpha1.Sandb
 
 	// to avoid the performance issue, using the controller to inject csi containers
 	// fetch the configmap and parse the configuration based on the controller runtime
-	podTemplate := &pod.Spec
-	injectErr := sidecarutils.InjectPodTemplateCSIAndRuntimeSidecar(ctx, box, podTemplate, r.Client)
+	injectErr := sidecarutils.InjectSandboxRuntimes(ctx, box, pod, r.Client)
 	if injectErr != nil {
 		klog.ErrorS(injectErr, "failed to inject pod template with csi sidecar or runtime sidecar", "sandbox", klog.KObj(box))
 		return nil, injectErr
--- a/pkg/utils/sidecarutils/config_types.go
+++ b/pkg/utils/sidecarutils/config_types.go
@@ -21,9 +21,11 @@ import (
 )
 
 const (
-	KEY_CSI_INJECTION_CONFIG     = "csi"
-	KEY_RUNTIME_INJECTION_CONFIG = "agent-runtime"
-	SandboxInjectionConfigName   = "sandbox-injection-config"
+	KEY_CSI_INJECTION_CONFIG            = "csi"
+	KEY_EGRESS_CONTROL_INJECTION_CONFIG = "egress-control"
+	KEY_RUNTIME_INJECTION_CONFIG        = "agent-runtime"
+
+	SandboxInjectionConfigName = "sandbox-injection-config"
 )
 
 type SidecarInjectConfig struct {
@@ -34,4 +36,13 @@ type SidecarInjectConfig struct {
 	Sidecars []corev1.Container `json:"csiSidecar"`
 	// Support injection for volume mount configurations
 	Volumes []corev1.Volume `json:"volume"`
+
+	// Annotations to be added to the pod.
+	Annotations map[string]string `json:"annotations"`
+	// Labels to be added to the pod.
+	Labels map[string]string `json:"labels"`
+	// InitContainers to be appended to the pod spec.
+	InitContainers []corev1.Container `json:"initContainers"`
+	// Containers to be appended to the pod spec.
+	Containers []corev1.Container `json:"containers"`
 }
--- a/pkg/utils/sidecarutils/egress-control/health_probe.go
+++ b/pkg/utils/sidecarutils/egress-control/health_probe.go
@@ -1,0 +1,338 @@
+package egresscontrol
+
+// Package egresscontrol provides Istio sidecar health probe rewriting.
+//
+// When Istio's envoy proxy intercepts all pod traffic, Kubernetes probes (readiness,
+// liveness, startup) must be redirected through the pilot agent on port 15020.
+// This package converts the original probes into HTTP probes pointing at the
+// pilot agent's /app-health/* endpoints, and serializes the original probe
+// definitions into the ISTIO_KUBE_APP_PROBERS env var on the istio-proxy container.
+
+import (
+	"encoding/json"
+	"fmt"
+	"slices"
+
+	"github.com/openkruise/agents/pkg/utils"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	ProxyContainerName           = "istio-proxy"
+	KubeAppProberEnv             = "ISTIO_KUBE_APP_PROBERS"
+	HealthProbeRewriteAnnotation = "networking.agents.kruise.io/health-probe-rewrite"
+	SidecarProxyAnnotation       = "networking.agents.kruise.io/sidecar-proxy"
+)
+
+// findSidecar returns the pointer to the first container whose name matches the "istio-proxy".
+func findSidecar(pod *corev1.Pod) *corev1.Container {
+	proxyContainerName := pod.Annotations[SidecarProxyAnnotation]
+	if proxyContainerName == "" {
+		proxyContainerName = ProxyContainerName
+	}
+	return findContainerFromPod(proxyContainerName, pod)
+}
+
+// findContainerFromPod returns the pointer to the first container whose name matches in init containers or regular containers
+func findContainerFromPod(name string, pod *corev1.Pod) *corev1.Container {
+	if c := utils.FindContainer(name, pod.Spec.Containers); c != nil {
+		return c
+	}
+	return utils.FindContainer(name, pod.Spec.InitContainers)
+}
+
+// convertAppProber returns an overwritten `Probe` for pilot agent to take over.
+func convertAppProber(probe *corev1.Probe, newURL string, statusPort int) *corev1.Probe {
+	if probe == nil {
+		return nil
+	}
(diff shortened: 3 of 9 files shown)


pkg/utils/sidecarutils/injector.go
```diff
@@ -0,0 +1,314 @@
...
+)
+
+type Injector struct {
+	mu      sync.RWMutex
+	configs map[string]*SidecarInjectConfig
+	client  kubernetes.Interface
+}
+
+// NewRuntimeController creates an Injector that watches the sandbox-injection-config
+func NewRuntimeController(cfg *rest.Config) (*Injector, error) {
```
furykerry: plz use informer cache based controller in controller-runtime, not directly with client-go and shared informer. Actually, is it really necessary to use a controller loop ? consider fetch config directly from informer cache, and cache the parsed config values

pkg/utils/sidecarutils/injector.go
```diff
@@ -0,0 +1,314 @@
...
+		lw,
+		&corev1.ConfigMap{},
+		0, // use default resync
+		toolscache.Indexers{toolscache.NamespaceIndex: toolscache.MetaNamespaceIndexFunc},
+	)
+	informer.AddEventHandler(toolscache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			cm := obj.(*corev1.ConfigMap)
+			if cm.Name == SandboxInjectionConfigName && namespace == cm.Namespace {
+				i.onConfigChange(cm)
```
furykerry: it is bad practice to put heavy operation(lock in onConfigChange) in event handler, which will block watch operations

pkg/utils/sidecarutils/config_types.go
```diff
@@ -17,21 +17,35 @@ limitations under the License.
...
 	// Configuration injection for the main container (by convention, one container is designated as the main container)
 	// Injection configuration items for business-specified main containers, such as volumeMount, environment variables, etc. Format: corev1.Container
-	MainContainer corev1.Container `json:"mainContainer"`
+	MainContainer corev1.Container `json:"mainContainer" yaml:"mainContainer"`
 	// Support injection for multiple independent sidecar containers; CSI container plugins are all injected from this
-	Sidecars []corev1.Container `json:"csiSidecar"`
+	Sidecars []corev1.Container `json:"csiSidecar" yaml:"csiSidecar"`
 	// Support injection for volume mount configurations
-	Volumes []corev1.Volume `json:"volume"`
+	Volume []corev1.Volume `json:"volume" yaml:"volume"`
```
furykerry: plz keep the field and json name as Volumes

pkg/utils/sidecarutils/config_types.go
```diff
@@ -17,21 +17,35 @@ limitations under the License.
...
+
+	// Template is a raw Pod spec overlay used by the controller-runtime Injector
+	// for unknown runtime types (e.g. egress-control). It is applied via strategic
+	// merge patch after container conflict checks.
+	Template json.RawMessage `json:"template" yaml:"template"`
```
furykerry: template patch is too free form, consider using the injection logic similar to csi/agent-runtime

pkg/utils/sidecarutils/egress-control/health_probe.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+	return string(b)
+}
+
+func allContainers(pod *corev1.Pod) []corev1.Container {
+	return append(slices.Clone(pod.Spec.InitContainers), pod.Spec.Containers...)
+}
+
+// patchRewriteProbe generates the patch for webhook.
+func patchRewriteProbe(annotations map[string]string, pod *corev1.Pod, defaultPort int32) {
```
furykerry: patchRewriteProbe accepts annotations parameter but never uses it

pkg/utils/sidecarutils/egress-control/health_probe.go
```diff
@@ -0,0 +1,342 @@
...
+			GRPC:           probe.GRPC,
+			TimeoutSeconds: probe.TimeoutSeconds,
+		}
+	}
+
+	return nil
+}
+
+// kubeLifecycleHandlerToInternalProber converts a Kubernetes LifecycleHandler to an Istio internal Prober
+func kubeLifecycleHandlerToInternalProber(lifecycelHandler *corev1.LifecycleHandler) *Prober {
```
furykerry: lifecycelHandler -> lifecycleHandler

pkg/utils/sidecarutils/config_types.go
```diff
@@ -17,21 +17,35 @@ limitations under the License.
...
+
+	// templatePod is the parsed form of Template, set during parsing.
+	templatePod *corev1.Pod
+	// raw caches the original config string for change detection.
+	raw string
```
furykerry: SidecarInjectConfig mixes public API and internal cache, consider using configmap generation for revision detection

pkg/utils/sidecarutils/sidecar_config_inject.go
```diff
@@ -77,7 +39,13 @@ func parseInjectConfig(ctx context.Context, configKey string, configRaw map[stri
 		return sidecarConfig, nil
 	}
 
-	err := json.Unmarshal([]byte(configRaw[configKey]), &sidecarConfig)
+	var err error
+	if strings.HasPrefix(configValue, "{") || strings.HasPrefix(configValue, "[") {
```
furykerry: The heuristic of checking the first character to decide JSON vs YAML is fragile. A YAML file could start with { (flow style). The parseConfigValue function in injector.go always uses yaml.Unmarshal (which also handles JSON), making the JSON-detection logic in parseInjectConfig redundant and inconsistent. Consider unifying both paths to always use json

pkg/utils/sidecarutils/config_types.go
Kuromesi: Json tag is not changed, only the go struct filed is changed, the go struct field and json tag is not consistent now.

pkg/controller/sandbox/core/common_control.go
```diff
@@ -436,8 +436,7 @@ func (r *commonControl) createPod(ctx context.Context, box *agentsv1alpha1.Sandb
 
 	// to avoid the performance issue, using the controller to inject csi containers
 	// fetch the configmap and parse the configuration based on the controller runtime
-	podTemplate := &pod.Spec
-	injectErr := sidecarutils.InjectPodTemplateCSIAndRuntimeSidecar(ctx, box, podTemplate, r.Client)
+	injectErr := sidecarutils.InjectPodTemplateCSIAndRuntimeSidecar(ctx, box, pod, r.Client)
```
furykerry: consider rename InjectPodTemplateCSIAndRuntimeSidecar to InjectRuntimeSidecars, since we have multiple sidecars to inject, not just csi, and agent-runtime

pkg/utils/sidecarutils/config_types.go
```diff
@@ -21,17 +21,28 @@ import (
...
+	KEY_RUNTIME_INJECTION_CONFIG        = "agent-runtime"
+
+	SandboxInjectionConfigName = "sandbox-injection-config"
 )
 
 type SidecarInjectConfig struct {
 	// Configuration injection for the main container (by convention, one container is designated as the main container)
 	// Injection configuration items for business-specified main containers, such as volumeMount, environment variables, etc. Format: corev1.Container
-	MainContainer corev1.Container `json:"mainContainer"`
+	MainContainer corev1.Container `json:"mainContainer" yaml:"mainContainer"`
```
furykerry: it is weird to mix json with yaml, since currently csi and agent-runtime is using json, can egress-control also use json?

pkg/utils/sidecarutils/config_types.go
```diff
@@ -21,17 +21,28 @@ import (
...
+	Volumes []corev1.Volume `json:"volume" yaml:"volume"`
+
+	// Annotations to be added to the pod.
+	Annotations map[string]string `json:"annotations" yaml:"annotations"`
+	// Labels to be added to the pod.
+	Labels map[string]string `json:"labels" yaml:"labels"`
+	// InitContainers to be appended to the pod spec.
+	InitContainers []corev1.Container `json:"initContainers" yaml:"initContainers"`
+	// Containers to be appended to the pod spec.
+	Containers []corev1.Container `json:"containers" yaml:"containers"`
```
furykerry: consider remove  InitContainers and Containers fields,  and use Sidecars fields instead it is basically the same thing

pkg/utils/sidecarutils/utils.go
```diff
@@ -61,9 +62,27 @@ func doSidecarInjection(ctx context.Context, sandbox *agentsv1alpha1.Sandbox, po
...
+
+	// Egress Control may rewrite health probes, so it should be injected after other sidecars.
+	if enableInjectEgressControlConfig(sandbox) {
+		egressControlInjectConfig, err := parseInjectConfig(ctx, KEY_EGRESS_CONTROL_INJECTION_CONFIG, injectConfigMap)
+		if err != nil {
+			logger.Error(err, "failed to parse egress control injection configuration")
+			return err
+		}
+		if !isContainersExists(pod.Spec.InitContainers, egressControlInjectConfig.InitContainers) && !isContainersExists(pod.Spec.Containers, egressControlInjectConfig.Containers) {
+			applyInjectionTemplate(ctx, pod, egressControlInjectConfig)
```
furykerry: the logic to use Sidecars, Volumes, Labels and Annotations are common to all runtimes, consider refactor the code, and extract all similar logics to here. So this injection can be a common injection logic.

pkg/utils/sidecarutils/config_types.go
Kuromesi: But the json tag of sidecar is csiSidecar, I think this may be a little bit confusing. And adding init containers and containers can also help to reduce the redundant code, similar runtimes can share the same injection config and not write code to determine if they want to inject in init containers or containers.

pkg/controller/sandbox/core/common_control.go
Kuromesi: Agree.

furykerry (approved):
/lgtm
/approve

PR #399, opened 2026-05-15
fix(sandbox-manager): prevent ClaimSandbox returning (nil, nil) on context cancel

AiRanthem:
## Summary

- Fixes a panic in `SandboxManager.ClaimSandbox` when the underlying sandbox lock fails with `context.Canceled` / `context.DeadlineExceeded` (or any other `wait.Interrupted`-matched error) on the first retry attempt.
- Root cause: `client-go`'s `retry.OnError` rewrites the final error to `lastErr` whenever `wait.Interrupted(err)` is true. When the first iteration returns an interrupted error and no prior retriable error has been recorded, `lastErr` is still `nil`, so `retry.OnError` returns `nil`. `Infra.ClaimSandbox` then returns `(nil sandbox, nil error)` and the caller in `pkg/sandbox-manager/api.go` panics on `sandbox.GetState()` / `sandbox.GetNamespace()`.
- Fix: detect `wait.Interrupted(claimErr)` inside the retry callback and re-emit the error via `fmt.Errorf(\"%v\", claimErr)` so the error chain no longer matches `errors.Is` checks used by `wait.Interrupted`, preventing the silent rewrite. Original message is preserved so logs and metrics still carry the underlying cause.

## Test plan

- [x] `go test ./pkg/sandbox-manager/infra/sandboxcr/ -count=1` passes locally.
- [x] New table-driven test `TestInfraClaimSandboxReturnsErrorWhenLockContextCanceled` covers:
  - `context.Canceled` wrapped in `url.Error` (typical apiserver client shape)
  - bare `context.Canceled`
  - `context.DeadlineExceeded` wrapped in `url.Error`
  - bare `context.DeadlineExceeded`
  - `wait.ErrorInterrupted(...)` (the third path matched by `wait.Interrupted`)
- [x] All sub-cases assert both the returned `err` and `metrics.LastError` carry the expected message and that the returned sandbox is `nil`.

### DIFF
--- a/pkg/sandbox-manager/infra/sandboxcr/infra.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/infra.go
@@ -168,6 +168,13 @@ func (i *Infra) ClaimSandbox(ctx context.Context, opts infra.ClaimSandboxOptions
 		} else {
 			metrics.RetryCost += tryMetrics.Total
 		}
+		// client-go retry.OnError rewrites interrupted errors to the last
+		// retriable error. Context cancellation is interrupted but non-retriable,
+		// so keep its message while intentionally avoiding %w; wrapping would
+		// still let wait.Interrupted identify and rewrite it.
+		if wait.Interrupted(claimErr) {
+			return fmt.Errorf("%v", claimErr)
+		}
 		return claimErr
 	})
 	return claimedSandbox, metrics, buildClaimError(err, metrics.LastError, metrics.PickSandboxFailures)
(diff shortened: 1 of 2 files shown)


pkg/sandbox-manager/infra/sandboxcr/infra.go
```diff
@@ -168,6 +168,13 @@ func (i *Infra) ClaimSandbox(ctx context.Context, opts infra.ClaimSandboxOptions
 		} else {
 			metrics.RetryCost += tryMetrics.Total
 		}
+		// client-go retry.OnError rewrites interrupted errors to the last
+		// retriable error. Context cancellation is interrupted but non-retriable,
+		// so keep its message while intentionally avoiding %w; wrapping would
+		// still let wait.Interrupted identify and rewrite it.
+		if wait.Interrupted(claimErr) {
```
furykerry, line 175: if context cancelled in the first retry, one may lose the original error (last error), consider using metrics.Retries to determine whether return context cancelled or last error

PR #404, opened 2026-05-15
fix: reject resume during pausing state and return 400 for conflict

AiRanthem:
## Summary
- `IsSandboxResumable` now splits the combined `Running/Resuming` case into separate branches, adding a `spec.paused` check for Running phase to reject the "pausing in progress" intermediate state (`phase=Running + spec.paused=true`).
- `ConnectSandbox` maps `ErrorConflict` from `ResumeSandbox` to HTTP 400 (BadRequest) instead of 500 (InternalServerError), following the E2B spec pattern.

## Test plan
- [x] `TestIsSandboxResumable` — 6 cases covering Running, Running+paused, Resuming, Paused+condition, Paused-no-condition, Succeeded
- [x] `TestConnectSandbox` — "resume sandbox: pausing" case now expects 400
- [x] Full `pkg/utils/sandboxutils/` and `pkg/servers/e2b/` test suites pass
- [x] `pkg/sandbox-manager/...` regression tests pass


furykerry (approved):
/lgtm
/approve

PR #405, opened 2026-05-15
feat: implement sandbox pausing lifecycle

AyushSriv06:
Ⅰ. Describe what this PR does
This PR introduces a new transitional lifecycle phase, Pausing, for the Sandbox resource.
Previously, when a Sandbox was set to Paused: true, it would transition directly from Running to Paused while the underlying Pod deletion was still in progress.

Added SandboxPausing to SandboxPhase in the API.
Updated the controller's state machine:
Running -> Pausing (triggered when Spec.Paused is true).
Pausing -> Paused (only after the SandboxPaused condition confirms the Pod is fully deleted).
Updated Prometheus metrics to track the Pausing phase.
Updated the Rate Limiter to treat Pausing as a non-creating phase, ensuring it doesn't block high-priority Sandbox creation.
Ⅱ. Does this pull request fix one issue?
fixes https://github.com/openkruise/agents/issues/250

Ⅲ. Describe how to verify it
Unit Tests: Updated sandbox_controller_test.go, metrics_test.go, and rateLimiter_test.go to cover the new state transitions.
E2E Tests: Verified the Sandbox lifecycle in a Kind cluster. The controller correctly transitions through Pausing before reaching Paused during hibernation.
Manual Verification: Observed Sandbox status transitions using kubectl get sandbox -w while toggling the spec.paused field.

### DIFF
--- a/api/v1alpha1/sandbox_types.go
+++ b/api/v1alpha1/sandbox_types.go
@@ -243,6 +243,8 @@ const (
 	// SandboxRunning means the pod has been bound to a node and all of the containers have been started.
 	// At least one container is still running or is in the process of being restarted.
 	SandboxRunning SandboxPhase = "Running"
+	// SandboxPausing means the sandbox is pausing and has not completed hibernation yet.
+	SandboxPausing SandboxPhase = "Pausing"
 	// SandboxPaused means the sandbox has entered the paused state.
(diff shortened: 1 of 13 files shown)


AyushSriv06:
@furykerry @zmberg @PersistentJZH
Hello Sir, may this PR get reviewed?

PR #407, opened 2026-05-16
fix(inplaceupdate): fix false-positive resource change detection caus…

zmberg:
### What

Fix incorrect resource comparison in `DefaultGenerateResizeSubresourceBody` that causes unnecessary pod resize operations.

### Why

The original code uses `reflect.DeepEqual` to compare `ResourceRequirements` between Sandbox spec and Pod. This produces false positives in two common scenarios:

| Scenario | Sandbox Spec | Pod Actual | Expected | Actual (before fix) |
|----------|-------------|-----------|----------|-------------------|
| System-injected resources | `cpu: 1, memory: 1Gi` | `cpu: 1, memory: 1Gi, ephemeral-storage: 30Gi` | Equal ✅ | Not equal ❌ |
| Different unit representation | `cpu: 1000m` | `cpu: "1"` | Equal ✅ | Not equal ❌ |

This leads to unnecessary resize API calls on every reconcile, wasting API server resources and potentially causing resize loops.

### How

- Replace `reflect.DeepEqual(origin.Resources, container.Resources)` with `resourcesEqual(desired, actual)`.
- `resourcesEqual` only checks resources **specified in the Sandbox spec**, ignoring extra resources injected by the system (e.g., kubelet adding `ephemeral-storage`).
- Uses `resource.Quantity.Cmp()` for unit-aware comparison instead of string/struct equality.

### Testing

Added 19 unit test cases:
- `TestResourceListContains` (9 cases)
- `TestResourcesEqual` (8 cases)
- `TestDefaultGenerateResizeSubresourceBody_NoResizeWhenExtraOrUnitDiff` (2 integration cases)

```bash
go test ./pkg/utils/inplaceupdate/ -run "TestResourceListContains|TestResourcesEqual|TestDefaultGenerateResizeSubresourceBody_NoResize" -v
# PASS
```

### DIFF
--- a/pkg/utils/inplaceupdate/inplace_update.go
+++ b/pkg/utils/inplaceupdate/inplace_update.go
@@ -234,6 +234,35 @@ func buildContainerResourcesMap(containers []corev1.Container) map[string]corev1
 	return result
 }
 
+// resourcesEqual compares two ResourceRequirements semantically.
+// It only checks resources specified in 'desired' (the sandbox spec), ignoring extra resources
+// in 'actual' (the pod) injected by the system (e.g., ephemeral-storage).
+// It uses Quantity.Cmp() to handle different unit representations (e.g., "1" vs "1000m").
+func resourcesEqual(desired, actual corev1.ResourceRequirements) bool {
+	if !resourceListContains(actual.Limits, desired.Limits) {
+		return false
+	}
+	if !resourceListContains(actual.Requests, desired.Requests) {
+		return false
+	}
+	return true
+}
+
+// resourceListContains checks that for every resource in 'desired', the 'actual' list
+// contains the same resource with an equal quantity value.
+func resourceListContains(actual, desired corev1.ResourceList) bool {
+	for resName, desiredQty := range desired {
+		actualQty, exists := actual[resName]
+		if !exists {
+			return false
+		}
+		if desiredQty.Cmp(actualQty) != 0 {
+			return false
+		}
+	}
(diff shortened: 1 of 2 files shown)


pkg/utils/inplaceupdate/inplace_update.go
```diff
@@ -234,6 +234,35 @@ func buildContainerResourcesMap(containers []corev1.Container) map[string]corev1
...
+	}
+	if !resourceListContains(actual.Requests, desired.Requests) {
+		return false
+	}
+	return true
+}
+
+// resourceListContains checks that for every resource in 'desired', the 'actual' list
+// contains the same resource with an equal quantity value.
+func resourceListContains(actual, desired corev1.ResourceList) bool {
```
furykerry, line 253: reuse isResourceListCovered func?

pkg/utils/inplaceupdate/inplace_update.go
```diff
@@ -234,6 +234,35 @@ func buildContainerResourcesMap(containers []corev1.Container) map[string]corev1
...
+	for resName, desiredQty := range desired {
+		actualQty, exists := actual[resName]
+		if !exists {
+			return false
+		}
+		if desiredQty.Cmp(actualQty) != 0 {
```
furykerry, line 259: to handle the problem of resource normalization of some cloud provider, consider use `actualQty.Cmp(desiredQty) > 0` instead

PR #409, opened 2026-05-16
fix(storages): make CSI NodePublishVolume idempotent via a deterministic VolumeId

### LINKED ISSUE #408: [BUG] CSI volume re-mount on resume is not idempotent and leaves mounts behind
# [BUG] CSI volume re-mount on resume is not idempotent and leaves mounts behind

**What happened**

When a claimed sandbox is resumed, the controller re-attaches its dynamic CSI volumes by calling `Initialize` → `ProcessCSIMounts`. The problem is that this re-mount is not "do it once and only if needed" — it re-runs on every reconcile while the sandbox is still in the `Resuming` phase, and it is not idempotent.

Three things combine to make this a real correctness problem:

- The "only do this once" guard in `EnsureSandboxResumed` only protects the *condition flip*. The actual re-mount call (`r.initializer.Initialize`) sits **outside** that guard, so it runs again on every reconcile until the phase is persisted as `Running`.
- On the resuming branch, when `Initialize` returns an error the new phase is **not persisted** (status is only written back for the upgrading branch on the error path). So one failing volume keeps the sandbox in `Resuming`, and the next reconcile re-mounts *the whole list again* — including the volumes that already mounted successfully.
- Every mount call builds a `NodePublishVolume` request with a **freshly randomized `VolumeId`** against the **same target path**:

  ```go
  // pkg/agent-runtime/storages/volume_mount_provider.go
  VolumeId:   fmt.Sprintf("%v-%s", persistentVolumeObj.Name, generateRandomString(6)),
  TargetPath: containerMountTarget,
  ```

  The CSI spec expects `NodePublishVolume` to be idempotent on `(volume_id, target_path)`. A changing `volume_id` defeats that, so the driver treats every retry as a brand-new publish. And there is **no `NodeUnpublishVolume` / unmount anywhere** in the codebase, so nothing ever cleans the previous mount up.

**Why it matters**

For block-style drivers this is wasteful. For **FUSE-based drivers — JuiceFS and Ceph** — it is genuinely harmful: each publish typically forks a mount-helper process / FUSE session. Re-publishing repeatedly on the same target path stacks bind-mounts and leaks helper processes, eventually corrupting the container's mount table and exhausting node resources. A single transient failure on one volume is enough to trigger the loop.

**How to reproduce**

1. Create a `SandboxClaim` whose sandbox mounts two or more dynamic CSI volumes (e.g. a JuiceFS or Ceph PV).
2. Pause, then resume the sandbox.
3. Make one volume's mount fail transiently (CSI sidecar briefly down, or a bad secret for just that volume).
4. The sandbox stays in `Resuming`; on the next reconcile **all** volumes are re-published with new random `VolumeId`s and no unmount of the prior mounts.

**Proposed fix**

- Track re-mount completion per resume (a dedicated status condition / annotation) and skip when already satisfied — mirror the existing "resumed once" semantics so re-mount runs once per resume.
- Derive a **stable** `VolumeId` from PV name + target path so `NodePublishVolume` is genuinely idempotent and the driver can short-circuit a repeat.
- Add unmount / rollback so a partial failure does not leave orphaned mounts.

**Scope and relation to existing work**

This is a standalone, pre-existing correctness bug in the *generic, driver-agnostic* CSI re-mount path — it is not tied to any roadmap or feature-project deliverable. It affects all CSI drivers; it is simply most visible on FUSE drivers (JuiceFS/Ceph) because those fork a mount helper per publish. For dedup: #305 introduced this resume re-mount path (touching `common_control.go` / `sandbox_initializer.go` / `pkg/utils/runtime/csi.go`) and #290 parallelized it, but neither makes it idempotent/once-per-resume nor adds cleanup — this is an unaddressed gap in that design, not a regression of a later change. #359 validates volume-claim-template mounts (different concern) and #254 (closed, unmerged) was superseded prep that did not address this. Happy to submit a fix.

**Environment**

- Kubernetes version: N/A — code-level defect
- Affects: any sandbox using dynamic CSI volumes across pause/resume or upgrade

PRAteek-singHWY:
## fix(storages): make CSI NodePublishVolume idempotent via a deterministic VolumeId

# fixes #408

`GenerateCSINodePublishVolumeRequest` built the CSI `VolumeId` as `<pv-name>-<random6>`, a **new random id on every call**, against a fixed `TargetPath`. The CSI spec requires `NodePublishVolume` to be idempotent keyed
on `(volume_id, target_path)`. Because the controller re-runs the resume re-mount (`Initialize` → `ProcessCSIMounts`) on every reconcile while the sandbox is in `Resuming` — and on every error-retry, since the resuming branch does not persist status on error — a changing `volume_id` made each retry look like a brand-new publish. For FUSE drivers (JuiceFS/Ceph) that stacks bind-mounts and leaks mount-helper processes; there is no `NodeUnpublish` to clean them up.

This PR makes the id **deterministic**: derived from the PV name plus a hash of the container target path. A spec-compliant driver/sidecar can now short-circuit an already-published volume, so the repeated re-mount becomes a safe no-op instead of a leak. The change is generic and driver-agnostic — no driver-specific code, no behavioural change on the first successful resume.

Also removes the now-dead `generateRandomString`/`charset` helper and its test, and strengthens the existing idempotency test to assert `VolumeId` stability (it previously asserted every field *except* `VolumeId`, precisely because the id was random).

Scope note: a controller-side "run re-mount once per resume" guard and an explicit `NodeUnpublish`/unmount path are reasonable follow-ups, but both are larger and the unmount path needs support in the external mount binary (none exists today). Deterministic, idempotent publish resolves the correctness and leak problem reported in #408 on its own and keeps this PR minimal.

fixes #408

- `go test ./pkg/agent-runtime/storages/...` — all green, including:
  - `TestDeterministicVolumeID` (new): stable for same pv+target; distinct per target path and per PV.
  - `TestMountProvider_GenerateNodePublishVolumeRequest_Idempotency` (strengthened): now asserts `firstResult.VolumeId == secondResult.VolumeId`
    and that a different target path yields a different id.
  - existing `GenerateCSINodePublishVolumeRequest` / edge-case / `IsPureReadOnly` tests unchanged and passing.
- `go build ./...` of dependents (`csiutils`, `runtime`, `controller/sandbox`, `sandbox-manager`) passes; `gofmt`/`go vet` clean.

Self-contained change in `pkg/agent-runtime/storages`. No new driver support, E2E suites, framework, or docs are added — purely makes the existing re-mount idempotent.

### DIFF
--- a/pkg/agent-runtime/storages/utils.go
+++ b/pkg/agent-runtime/storages/utils.go
@@ -17,21 +17,24 @@ limitations under the License.
 package storages
 
 import (
-	"math/rand"
-	"time"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 )
 
-const charset = "abcdefghijklmnopqrstuvwxyz0123456789"
-
-func generateRandomString(length int) string {
-	rand.Seed(time.Now().UnixNano())
-	b := make([]byte, length)
-	for i := range b {
-		b[i] = charset[rand.Intn(len(charset))]
-	}
-	return string(b)
+// deterministicVolumeID returns a stable CSI volume id for a (PV, target path)
+// pair. The CSI spec requires NodePublishVolume to be idempotent keyed on
+// (volume_id, target_path): a previously random id meant every resume re-mount
+// looked like a brand-new publish to the driver, stacking mounts. A stable id
+// lets a spec-compliant driver short-circuit an already-published volume.
+// The target path is hashed (it contains '/'), keeping the id bounded while
+// remaining unique per target so the same PV mounted at different paths does
+// not collide.
+func deterministicVolumeID(pvName, targetPath string) string {
+	sum := sha256.Sum256([]byte(targetPath))
+	return fmt.Sprintf("%s-%s", pvName, hex.EncodeToString(sum[:])[:12])
 }
 
 func IsPureReadOnly(accessModes []corev1.PersistentVolumeAccessMode) bool {
--- a/pkg/agent-runtime/storages/utils_test.go
+++ b/pkg/agent-runtime/storages/utils_test.go
@@ -79,40 +79,26 @@ func TestIsPureReadOnly(t *testing.T) {
 	}
 }
 
-func TestGenerateRandomString(t *testing.T) {
-	tests := []struct {
-		name     string
-		length   int
-		expected int
-	}{
-		{
-			name:     "zero length",
-			length:   0,
-			expected: 0,
-		},
-		{
-			name:     "positive length",
-			length:   5,
-			expected: 5,
-		},
-		{
-			name:     "large length",
-			length:   10,
-			expected: 10,
-		},
-	}
+func TestDeterministicVolumeID(t *testing.T) {
+	// Same (pv, target) must always yield the same id so a resume re-mount is
+	// idempotent per the CSI spec.
+	id1 := deterministicVolumeID("pv-a", "/mnt/data")
+	id2 := deterministicVolumeID("pv-a", "/mnt/data")
+	assert.Equal(t, id1, id2, "same pv+target must produce a stable volume id")
 
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			result := generateRandomString(tt.length)
-			assert.Len(t, result, tt.expected)
+	// The id must contain the PV name (existing callers/tests rely on this).
+	assert.Contains(t, id1, "pv-a")
 
-			if tt.length > 0 {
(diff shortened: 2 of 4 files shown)


PRAteek-singHWY:
Hi @furykerry @AiRanthem @zmberg , #409 fixes #408. On resume the same volume was being re-mounted with a brand-new random ID every time on the same path, so the CSI driver couldn't tell it was the same volume and kept piling up mounts (and leaking mount-helper processes on JuiceFS/Ceph). The fix gives each volume a stable ID from its PV name + target path, so re-mounting the same spot is now a safe no-op, nothing changes on a normal first resume.

Issue #323, opened 2026-05-01
feat: integrate peers with KeyStorage for cross-replica cache invalidation


AiRanthem:
### Background

When `--e2b-key-storage=mysql` is used, the `mysqlKeyStorage` maintains local in-memory TTL caches (`byKey` and `byID`) with a 10-minute TTL. Currently, when `DeleteKey` is called on a key, it only invalidates the cache on the local node. In a multi-replica sandbox-manager deployment (which is the common case given that peer discovery via memberlist is already set up), other replicas will continue to serve the deleted key from their own caches until the TTL expires, leaving a window of up to 10 minutes where a deleted API key can still be used for authentication.

### Issue Type

feature

### Relevant Code

- `pkg/servers/e2b/keys/interface.go` — `KeyStorage` interface (with `DeleteKey` method)
- `pkg/servers/e2b/keys/mysql.go` — `mysqlKeyStorage.DeleteKey()` only clears local `byKey` / `byID` caches (lines 390-393)
- `pkg/peers/peers.go` — Existing `Peers` interface for peer discovery (with `GetPeers()` to list alive peers)
- `pkg/servers/e2b/core.go` — Controller already builds SandboxManager with `WithMemberlistPeers()`

### Notes

- The `secretKeyStorage` backend (K8s Secret-based) avoids this issue via periodic full refresh, but the MySQL backend relies on TTL-based cache expiration.
- The stale cache window equals `keyCacheTTL` (10 minutes).
- Multiple sandbox-manager replicas may share the same MySQL database.
- The sandbox-manager already has peer discovery infrastructure via memberlist, which could potentially be leveraged for broadcasting cache invalidation events.

PR #414, opened 2026-05-17
fix(proxy): bound and pool peer route-sync so one slow replica can't stall the lifecycle path

### LINKED ISSUE #413: [BUG] Peer route sync does an unbounded per-peer fan-out with no connection pooling and a retry budget larger than its own timeout
**What happened**

Every sandbox lifecycle change (claim / pause / resume / delete) propagates the route to peer replicas through `Server.SyncRouteWithPeers`. The way it does that doesn't hold up under load:

- It starts one goroutine per peer per route change, and the caller blocks on `wg.Wait()` until the slowest peer finishes.
- The HTTP client has no connection pooling. It uses the default transport (`MaxIdleConnsPerHost = 2`), so a burst of route changes to many peers keeps opening fresh connections instead of reusing them.
- The retry budget is larger than the per-request timeout, and it retries on every error:

  ```go
  // pkg/proxy/routes.go
  retry.OnError(wait.Backoff{Steps: 10, Duration: 10ms, Factor: 1.0, Jitter: 1.2},
      func(err error) bool { return true },                 // retries even non-retriable errors
      func() error { return requestPeer(http.MethodPost, peerIP, RefreshAPI, body) })
  ```

  `Factor: 1.0` means the backoff never grows: about 10ms x jitter x 10 steps, so roughly 100-120ms of sleeps, plus up to the 100ms client timeout per attempt for a slow or hung peer. So a single slow peer can push one route sync to ~1s, and because `wg.Wait()` blocks the caller, that latency lands on the synchronous sandbox lifecycle path (called from `pkg/sandbox-manager/api.go` in the claim/pause/resume/delete handlers).

**Why it matters**

Route changes scale with the number of sandbox lifecycle events, and the fan-out scales with the number of replicas. At the volume the load-testing roadmap targets (toward 100k sandboxes) this turns into an outbound-connection and goroutine storm (`routeChanges x peers`, with almost no connection reuse). Worse, one slow or unhealthy replica adds ~1s of head-of-line latency to every sandbox lifecycle operation across the fleet, because every caller waits on the slowest peer's full retry budget.

**How to reproduce**

1. Run at least 3 sandbox-manager replicas (memberlist peers).
2. Make one replica slow to answer `POST /refresh` (around 100ms+ latency, or refuse connections).
3. Drive a steady stream of claim/delete operations.
4. Every lifecycle op that triggers a route sync blocks ~1s on `wg.Wait()`; goroutine and outbound-connection counts grow proportionally to `routeChanges x peers`.

**Proposed fix**

- Use a shared, tuned `http.Transport` with `MaxIdleConnsPerHost` / `MaxConnsPerHost` and keep-alives.
- Bound the fan-out with a concurrency semaphore instead of an unbounded goroutine per peer.
- Make the total retry budget strictly smaller than the per-call timeout, with a real growth factor and a predicate that only retries transient errors.
- Optionally make peer sync best-effort / asynchronous off the synchronous request path, with the existing reconcile loop as the correctness backstop.

**Scope and relation to existing work**

This is a standalone, pre-existing scalability/resiliency bug in the existing peer route-sync code. It is not tied to any roadmap or feature-project deliverable, and only surfaces at scale and under replica churn.

It is distinct from #313 and #323. #323 is about *what* state is synced (peers and KeyStorage). The open PR #313 touches `pkg/proxy/routes.go` only for ext-proc/peer-reconcile correctness and a log line; it does not modify the peer HTTP client, the retry/backoff, or the per-peer fan-out, so there is no functional overlap (only a minor same-file adjacency a fix would rebase around). This issue is purely about the transport, retry, and fan-out mechanics of `SyncRouteWithPeers`. Happy to send a fix.

**Environment**

- Kubernetes version: N/A (code-level defect)
- Affects: any multi-replica sandbox-manager / sandbox-gateway deployment with memberlist peering; severity scales with replica count and lifecycle event rate

PRAteek-singHWY:
Fixes #413

## Why

`SyncRouteWithPeers` starts one goroutine per peer on every route change, uses the default HTTP transport (no connection pooling), and has a retry budget larger than its own per-request timeout, all while the caller blocks on `wg.Wait()`. Under load this becomes a goroutine and outbound-connection storm, and a single slow replica can add roughly a second of latency to every sandbox lifecycle operation across the fleet. Full analysis in #413.

## What changed

- Connection pooling: the peer client now uses a shared, tuned `http.Transport` (bounded idle/total conns per host, keep-alives) so syncs reuse connections instead of dialing fresh ones.
- Bounded fan-out: peer requests run under a concurrency semaphore instead of an unbounded goroutine per peer.
- Retry: the inter-attempt sleep budget is now strictly smaller than the per-request timeout, the backoff actually grows, and only transient failures are retried (network errors, 429, 5xx). 4xx and other permanent errors fail fast.
- A context deadline caps the whole per-peer attempt sequence so one hung replica can't stall the lifecycle path. The peer-reconcile loop is still the correctness backstop.

Behaviour is unchanged when all peers are healthy.

## Testing

New unit tests in `pkg/proxy/peer_sync_test.go`:

- `TestIsRetryablePeerError`: rejects context errors and 4xx (including 409), retries 429/5xx and transport errors.
- `TestPeerSyncBackoffBudgetUnderRequestTimeout`: total inter-attempt sleep budget is strictly less than `consts.RequestPeerTimeout`.
- `TestSyncRouteWithPeers_ConcurrencyBounded`: with 200 peers, peak in-flight requests never exceed `peerSyncMaxConcurrency`.

`go test ./pkg/proxy/... ./pkg/sandbox-manager/...` passes (existing `SyncRouteWithPeers` and memberlist tests still green); `go vet` and `gofmt` clean.

Not automated: the 3-replica scenario with one replica slow on `POST /refresh` was reasoned through but not turned into a test.

## Scope

Only `pkg/proxy` route-sync transport, retry, and fan-out change. What gets synced is unchanged, so there is no overlap with #313 or #323. No new framework, tooling, or CI.

### DIFF
--- a/pkg/proxy/peer_sync_test.go
+++ b/pkg/proxy/peer_sync_test.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package proxy
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/openkruise/agents/pkg/peers"
+	"github.com/openkruise/agents/pkg/sandbox-manager/consts"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsRetryablePeerError(t *testing.T) {
+	cases := []struct {
+		name string
+		err  error
+		want bool
+	}{
+		{"nil", nil, false},
+		{"context canceled", context.Canceled, false},
+		{"context deadline", context.DeadlineExceeded, false},
+		{"wrapped context deadline", fmt.Errorf("Post: %w", context.DeadlineExceeded), false},
+		{"http 400", &peerStatusError{ip: "1.2.3.4", code: http.StatusBadRequest}, false},
+		{"http 404", &peerStatusError{ip: "1.2.3.4", code: http.StatusNotFound}, false},
+		{"http 409", &peerStatusError{ip: "1.2.3.4", code: http.StatusConflict}, false},
+		{"http 429", &peerStatusError{ip: "1.2.3.4", code: http.StatusTooManyRequests}, true},
+		{"http 500", &peerStatusError{ip: "1.2.3.4", code: http.StatusInternalServerError}, true},
+		{"http 503", &peerStatusError{ip: "1.2.3.4", code: http.StatusServiceUnavailable}, true},
+		{"transport error", fmt.Errorf("dial tcp 1.2.3.4:7789: connect: connection refused"), true},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.want, isRetryablePeerError(tc.err))
+		})
+	}
+}
+
+// The whole point of the fix: the inter-attempt sleep budget must stay strictly
+// below a single per-request timeout, otherwise retries dominate the latency.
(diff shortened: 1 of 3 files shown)


PRAteek-singHWY:
Hi @furykerry @AiRanthem @zmberg, opened this to fix the peer route-sync path. It only touches the route-sync mechanics: a shared pooled HTTP transport, a bounded fan-out instead of one goroutine per peer, and a retry budget smaller than the per-request timeout so one slow replica can't stall the lifecycle path. What gets synced is unchanged, so it doesn't overlap #313 or #323, and there are new unit tests in pkg/proxy with the existing suites passing.

PR #420, opened 2026-05-19
fix(inplaceupdate): fix false-positive resource change detection caus…

zmberg:
### What

Fix incorrect resource comparison in `DefaultGenerateResizeSubresourceBody` that causes unnecessary pod resize operations.

### Why

The original code uses `reflect.DeepEqual` to compare `ResourceRequirements` between Sandbox spec and Pod. This produces false positives in two common scenarios:

| Scenario | Sandbox Spec | Pod Actual | Expected | Actual (before fix) |
|----------|-------------|-----------|----------|-------------------|
| System-injected resources | `cpu: 1, memory: 1Gi` | `cpu: 1, memory: 1Gi, ephemeral-storage: 30Gi` | Equal ✅ | Not equal ❌ |
| Different unit representation | `cpu: 1000m` | `cpu: "1"` | Equal ✅ | Not equal ❌ |

This leads to unnecessary resize API calls on every reconcile, wasting API server resources and potentially causing resize loops.

### How

- Replace `reflect.DeepEqual(origin.Resources, container.Resources)` with `resourcesEqual(desired, actual)`.
- `resourcesEqual` only checks resources **specified in the Sandbox spec**, ignoring extra resources injected by the system (e.g., kubelet adding `ephemeral-storage`).
- Uses `resource.Quantity.Cmp()` for unit-aware comparison instead of string/struct equality.

### DIFF
--- a/pkg/utils/inplaceupdate/inplace_update.go
+++ b/pkg/utils/inplaceupdate/inplace_update.go
@@ -281,7 +281,7 @@ func DefaultGenerateResizeSubresourceBody(opts InPlaceUpdateOptions) *corev1.Pod
 		if !ok {
 			continue
 		}
-		if reflect.DeepEqual(origin.Resources, container.Resources) {
+		if resourcesEqual(origin.Resources, container.Resources) {
 			continue
 		}
 		container.Resources = origin.Resources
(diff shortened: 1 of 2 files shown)


furykerry (approved):
/lgtm

PR #421, opened 2026-05-19
fix(keys): replace ticker with informer-driven refresh for secret-backed key storage

### LINKED ISSUE #418: bug: secret KeyStorage read path is per-process, causing write-after-read inconsistency across sandbox-manager replicas
### Background

`sandbox-manager` uses the `secret` KeyStorage backend by default (`pkg/servers/e2b/keys/secret.go`), which persists all API keys in a single K8s Secret. When `sandbox-manager` is deployed with ≥2 replicas, after creating an API key on any replica, a subsequent `GET /api-keys` request may fail to return that key for a long time.

The E2E case `test/e2b/test_apikey.py::test_api_keys_lifecycle` reproduces this consistently in a multi-replica environment:

- `POST /api-keys` returns 201 with the new uid.
- `GET /api-keys` returns a list that does not contain the new uid.
- Adding a few-second `time.sleep` does not help; the MySQL backend does not reproduce the issue.

### Issue Type

bug

### Relevant Code

- `pkg/servers/e2b/keys/secret.go`
  - `CreateKey` / `DeleteKey` only update the in-process `idxByKey` / `idxByID` / `idxByTeam` (`sync.Map`) after writing the secret.
  - `ListByOwnerTeam` / `ListTeams` / `LoadByKey` / `LoadByID` / `FindTeamByName` all read exclusively from the in-process `idxBy*` maps.
  - The only cross-process synchronization source is the fixed 10-minute ticker started in `Run()` (`newRefreshTicker = time.NewTicker(10 * time.Minute)`); there is no informer watch handler nor any peer broadcast.
- `config/sandbox-manager/deployment.yaml`: defaults to `replicas: 1`, which implicitly hides the multi-replica inconsistency.

### Notes

- The MySQL backend (`pkg/servers/e2b/keys/mysql.go::ListByOwnerTeam`) queries the DB directly and is not affected.
- Possible direction for discussion (not prescribing an implementation): the `secret` backend already lives inside the controller-runtime ecosystem, which provides an informer cache. Leveraging the informer on the read path could offer near-real-time consistency while balancing latency and API Server pressure across replicas; the write path could keep using `APIReader` if read-modify-write consistency is still required. Concrete design is left open.
- Reproduction: bump `replicas` in `config/sandbox-manager/deployment.yaml` to ≥2, deploy, and run `pytest test/e2b/test_apikey.py::test_api_keys_lifecycle`. The "Created key ... not found in list" assertion is almost always triggered.

AiRanthem:
## Summary

fixes #418 

In multi-replica deployments, the previous secret-backed `KeyStorage` relied on a 10-minute ticker to refresh its in-memory indexes. As a result, a key created or deleted on one replica could take up to 10 minutes to become visible on others. This PR replaces the ticker with an informer-event-driven refresh, bounded by informer push latency (typically sub-second).

### Key changes

- **`pkg/cache`**: expose `ctrlcache.Cache` via `Provider.GetCache()` so downstream components can register informer event handlers on the shared manager cache.
- **`pkg/servers/e2b/keys/secret.go`**:
  - Drop the 10-minute ticker.
  - Register a `Secret` event handler on the shared cache, filtered to the `e2b-key-store` Secret.
  - Coalesce events into a buffered (cap=1) channel and drain them in a single worker goroutine that calls `refresh()`.
  - `Run()` registers the handler and starts the worker; `Stop()` removes the handler and waits for the worker to exit.
- **`pkg/servers/e2b/keys/factory.go`**: `Config.Cache` is now required for secret mode and validated in `NewKeyStorage`. `NewSecretKeyStorage` itself does not validate (allowing unit tests to construct the storage without a cache as long as they do not call `Run()`).
- **`pkg/servers/e2b/core.go`**: wire `cache.Provider.GetCache()` into `keys.Config.Cache`.
- **Docs**: update `pkg/servers/e2b/keys/AGENTS.md` to describe the new watch-driven semantics.

### Why no writer-triggered refresh

An earlier iteration added `triggerRefresh()` calls in `CreateKey` / `DeleteKey` to compress the writer's own propagation latency. Codex review correctly flagged this as a stale-cache race: the worker reads through the controller-runtime cached client, whose local store is populated only by informer watch events. The worker almost always loses the race against its own write and reads a stale Secret revision, briefly removing the just-stored key from the indexes (or re-adding a just-deleted key) until the informer event arrives.

The writer-triggered refresh provided no real benefit — `storeKey()` and `idxBy*.Delete()` already update the auth indexes inline before returning, so the writing replica's authentication path is correct without it. The trigger was removed; cross-replica visibility remains purely informer-driven.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./pkg/...` clean
- [x] `go test ./pkg/servers/e2b/keys/...` all green (existing + new tests for `triggerRefresh` coalescing, `onSecretEvent` filtering/tombstone, `Run`/`Stop` lifecycle, factory cache validation)
- [x] `go test ./pkg/servers/e2b/... ./pkg/cache/...` all green
- [ ] (Reviewer) verify multi-replica behavior in a staging cluster: create a key on one replica, confirm the other replica authenticates with it within informer push latency

### DIFF
--- a/pkg/servers/e2b/keys/secret.go
+++ b/pkg/servers/e2b/keys/secret.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
+	ctrlcache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openkruise/agents/pkg/sandbox-manager/logs"
@@ -60,21 +61,24 @@ type secretKeyStorage struct {
 	Client client.Client
 	// APIReader is used for reading secrets before cache is started (e.g., during Init).
 	APIReader client.Reader
-	stop      chan struct{}
-	done      chan struct{}
-	stopOnce  sync.Once
+	Cache     ctrlcache.Cache
+
+	stop     chan struct{}
+	done     chan struct{}
+	stopOnce sync.Once
 
 	idxByKey  sync.Map
 	idxByID   sync.Map
 	idxByTeam sync.Map // teamName -> *models.Team
 }
 
-func NewSecretKeyStorage(client client.Client, apiReader client.Reader, namespace, adminKey string) KeyStorage {
+func NewSecretKeyStorage(client client.Client, apiReader client.Reader, cache ctrlcache.Cache, namespace, adminKey string) KeyStorage {
 	return &secretKeyStorage{
 		Namespace: namespace,
 		AdminKey:  adminKey,
 		Client:    client,
 		APIReader: apiReader,
+		Cache:     cache,
 		stop:      make(chan struct{}),
 		done:      make(chan struct{}),
 	}
--- a/pkg/servers/e2b/keys/secret.go
+++ b/pkg/servers/e2b/keys/secret.go
@@ -46,7 +46,7 @@ var (
 	marshalAPIKey = json.Marshal
 )
 
-const secretKeyStorageRefreshInterval = 15 * time.Second
+const secretKeyStorageRefreshInterval = 10 * time.Minute
 
 func init() {
 	AdminKeyID = uuid.MustParse("550e8400-e29b-41d4-a716-446655440000") // no means, just a const
@@ -169,6 +169,9 @@ func (k *secretKeyStorage) refresh(ctx context.Context, reader client.Reader) er
 	return nil
 }
 
+// triggerRefresh uses refreshC, which is a buffered channel (cap=1), to coalesce refresh signals.
+// Multiple events between refresh cycles are collapsed into a single refresh,
+// preventing redundant Secret reads when the informer delivers a burst of events.
 func (k *secretKeyStorage) triggerRefresh() {
 	select {
 	case k.refreshC <- struct{}{}:
--- a/pkg/servers/e2b/keys/secret.go
+++ b/pkg/servers/e2b/keys/secret.go
@@ -46,6 +46,8 @@ var (
 	marshalAPIKey = json.Marshal
 )
 
+const secretKeyStorageRefreshInterval = 15 * time.Second
+
 func init() {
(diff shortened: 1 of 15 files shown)


pkg/servers/e2b/keys/secret.go
```diff
@@ -154,32 +159,80 @@ func (k *secretKeyStorage) refresh(ctx context.Context, reader client.Reader) er
 	return nil
 }
 
+func (k *secretKeyStorage) triggerRefresh() {
```
furykerry, line 175: The refreshC chan struct{} with cap=1 coalescing pattern is clever but non-obvious. Multiple informer events between refresh cycles get collapsed into one refresh.

consider add a brief comment explaining the coalescing semantics, e.g. 
```
// refreshC is a buffered channel (cap=1) used for coalescing refresh signals.
// Multiple events between refresh cycles are collapsed into a single refresh,
// preventing redundant Secret reads when the informer delivers a burst of events.
```

pkg/servers/e2b/keys/secret.go
```diff
@@ -38,15 +40,14 @@ import (
...
-	newRefreshTicker = func() *time.Ticker {
-		return time.NewTicker(10 * time.Minute)
-	}
+	KeySecretName = "e2b-key-store"
+	AdminKeyID    uuid.UUID
+	generateUUID  = uuid.New
+	marshalAPIKey = json.Marshal
 )
 
+const secretKeyStorageRefreshInterval = 15 * time.Second
```
furykerry: The old ticker was 10 minutes; the new periodic fallback is 15 seconds. Since the informer events should handle most updates, the periodic refresh is a safety net. Consider making this configurable or increasing the interval (e.g., 10 minutes) to reduce unnecessary refresh operations

pkg/servers/e2b/keys/secret.go
AiRanthem: Added

pkg/servers/e2b/keys/secret.go
AiRanthem: Updated to 10 min

PR #422, opened 2026-05-20
fix(e2b): allow pause sandboxes in state pausing

AiRanthem:

Make the E2B `PauseSandbox` endpoint idempotent for already-paused sandboxes and concurrent pause requests.

- Delegate pause state handling to the sandbox manager/infra layer so an already-paused sandbox returns `204 No Content`.
- Preserve `409 Conflict` for real pause conflicts, including active opposite wait tasks such as an in-flight resume operation.
- Add regression coverage for both repeated pause and pause-vs-resume wait conflict behavior.

NONE

```bash
gofmt -w pkg/servers/e2b/pause_resume.go pkg/servers/e2b/pause_resume_test.go
git diff --check
go test -mod=readonly ./pkg/servers/e2b
```

Please focus review on the HTTP status mapping for pause conflicts. The handler now maps both sandbox-manager domain conflicts and cache wait-task conflicts to `409 Conflict`.

### DIFF
--- a/pkg/servers/e2b/pause_resume.go
+++ b/pkg/servers/e2b/pause_resume.go
@@ -18,11 +18,12 @@ package e2b
 
 import (
 	"context"
-	stderrors "errors"
+	"errors"
 	"fmt"
 	"net/http"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/klog/v2"
 
 	"github.com/openkruise/agents/api/v1alpha1"
@@ -59,8 +60,11 @@ func (sc *Controller) PauseSandbox(r *http.Request) (web.ApiResponse[struct{}], 
 }
 
 func pauseSandboxErrorCode(err error) int {
+	if apierrors.IsNotFound(err) {
+		return http.StatusNotFound
+	}
 	if managererrors.GetErrCode(err) == managererrors.ErrorConflict ||
-		stderrors.Is(err, cacheutils.ErrWaitTaskConflict) {
+		errors.Is(err, cacheutils.ErrWaitTaskConflict) {
 		return http.StatusConflict
 	}
 	return http.StatusInternalServerError
--- a/pkg/servers/e2b/pause_resume.go
+++ b/pkg/servers/e2b/pause_resume.go
@@ -18,6 +18,7 @@ package e2b
 
 import (
 	"context"
+	stderrors "errors"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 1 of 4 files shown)


pkg/servers/e2b/pause_resume.go
```diff
@@ -41,18 +43,12 @@ func (sc *Controller) PauseSandbox(r *http.Request) (web.ApiResponse[struct{}],
 	if apiErr != nil {
 		return web.ApiResponse[struct{}]{}, apiErr
 	}
-	if state, reason := sbx.GetState(); state != v1alpha1.SandboxStateRunning {
-		log.Info("skip pause sandbox: sandbox is not running", "state", state, "reason", reason)
-		return web.ApiResponse[struct{}]{}, &web.ApiError{
-			Code:    http.StatusConflict,
-			Message: fmt.Sprintf("Sandbox %s is not running", id),
-		}
-	}
```
furykerry, line 50: in func `func (s *Sandbox) Pause(ctx context.Context, opts infra.PauseOptions) error`, shall we call NewSandboxPauseTask after L319, so that a paused sandbox can return quickly without initiating a pause task

pkg/servers/e2b/pause_resume.go
```diff
@@ -62,6 +58,14 @@ func (sc *Controller) PauseSandbox(r *http.Request) (web.ApiResponse[struct{}],
 	}, nil
 }
 
+func pauseSandboxErrorCode(err error) int {
+	if managererrors.GetErrCode(err) == managererrors.ErrorConflict ||
```
furykerry, line 66: 1. PauseSandbox can also return a k8s.io/apimachinery not-found error if the sandbox is deleted concurrently, shall we return 404?

pkg/servers/e2b/pause_resume_test.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package e2b
 
 import (
+	stderrors "errors"
```
furykerry: why not uses errors directly? it seems no name collision exists.

pkg/servers/e2b/pause_resume.go
```diff
@@ -18,14 +18,16 @@ package e2b
 
 import (
 	"context"
+	stderrors "errors"
```
furykerry: why not uses errors directly? it seems no name collision exists.

pkg/servers/e2b/pause_resume.go
AiRanthem: OK

pkg/servers/e2b/pause_resume_test.go
AiRanthem: recovered

pkg/servers/e2b/pause_resume.go
AiRanthem: recovered

pkg/servers/e2b/pause_resume.go
AiRanthem: OK

PR #426, opened 2026-05-20
fix(sandboxupdateops): fail fast when failed sandboxes block all maxUnavailable slots

### LINKED ISSUE #425: bug: SandboxUpdateOps stuck in Updating when failed count reaches maxUnavailable
### Background

When upgrading Claimed Sandboxes via SandboxUpdateOps, if the number of failed sandboxes reaches `maxUnavailable` while candidates remain, the SUO gets permanently stuck in `Updating` phase. This blocks creation of any recovery SUO in the same namespace due to the single-active-ops webhook guard, and operators typically don't have delete permissions on SUO resources.

### Issue Type

bug

### Relevant Code

`pkg/controller/sandboxupdateops/sandboxupdateops_controller.go` — the concurrency formula `toUpgrade = maxUnavailable - updating - failed` returns 0 when `failed >= maxUnavailable`, but the terminal state condition `updated+failed == total` is never met because candidates still exist. The SUO stays in `Updating` indefinitely.

### Notes

- Reproduce: create a SUO with `maxUnavailable: 1` selecting 2+ sandboxes, let the first sandbox fail (e.g. bad preUpgrade hook). The SUO will block all remaining sandboxes in `Updating` indefinitely.
- The webhook at `pkg/webhook/sandboxupdateops/validating/sandboxupdateops_validate.go` rejects new SUO creation when a non-terminal SUO exists in the namespace.

zyl1121:

Adds a phase transition case in the SandboxUpdateOps controller: when `failed >= maxUnavailable`, no sandboxes are actively updating, and candidates remain, the SUO transitions directly to `Failed` instead of staying stuck in `Updating`.

Previously, the SUO would block indefinitely because `toUpgrade = maxUnavailable - updating - failed` evaluates to 0 while the terminal condition `updated+failed == total` is never satisfied. This made it impossible to create recovery SUOs in the same namespace without delete permissions.

fixes #425

1. `go test ./pkg/controller/sandboxupdateops/...`
2. New test `TestReconcile_FailedBlocksAllSlotsLeadsToFailed`: maxUnavailable=1, 1 failed + 2 candidates → SUO transitions to Failed
3. New test `TestReconcile_FailedBelowMaxUnavailableContinuesUpdating`: maxUnavailable=2, 1 failed + 1 candidate → SUO stays Updating, candidate gets patched

- The new case is ordered after the existing `updated+failed == total` case so it only triggers when there are still unprocessed candidates
- `updating == 0` guard ensures we don't prematurely fail while in-flight upgrades could still succeed and free up slots

### DIFF
--- a/pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
+++ b/pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
@@ -184,6 +184,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		} else {
 			newStatus.Phase = agentsv1alpha1.SandboxUpdateOpsCompleted
 		}
+
+	case failed > 0 && updating == 0 && len(candidates) > 0 &&
+		int(calculateMaxUnavailable(ops.Spec.UpdateStrategy.MaxUnavailable, total)) <= int(failed):
+		newStatus.Phase = agentsv1alpha1.SandboxUpdateOpsFailed
 	}
 
 	// Only initiate new upgrades when in Updating phase, not paused, and there are candidates
--- a/pkg/controller/sandboxupdateops/sandboxupdateops_controller_test.go
+++ b/pkg/controller/sandboxupdateops/sandboxupdateops_controller_test.go
@@ -949,6 +949,63 @@ func TestHandleDeletion_UpdateFinalizerRemovalError(t *testing.T) {
 	assert.Contains(t, err.Error(), "simulated update error")
 }
 
+func TestReconcile_FailedBlocksAllSlotsLeadsToFailed(t *testing.T) {
+	maxUnavail := intstrutil.FromInt32(1)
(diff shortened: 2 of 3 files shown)


furykerry:
@zyl1121 thanks for your patch, but i think keep the suo at the upgrading phase is desired, at least for default behavior,  since the sandbox may recover from failure through retry or dependent service modification

zyl1121:
> @zyl1121 thanks for your patch, but i think keep the suo at the upgrading phase is desired, at least for default behavior, since the sandbox may recover from failure through retry or dependent service modification

@furykerry I see your point, but I think auto-retry is not that straightforward. Different failure stages require different recovery actions. For example, if UpgradePodFailed happens, the user needs to fix the image and skip preUpgrade since it has already run. A blind retry could overwrite the backup, so this really requires human intervention.

The reason I raised this fix is the current limitation introduced by the webhook guard: once the SUO is stuck in Updating, no new SUO can be created in the same namespace, and operators usually don't have delete permissions, leaving no recovery path.

Are you already planning to add intelligent recovery to the SUO controller? If so, I'm happy to drop this PR and wait for that instead.

Issue #427, opened 2026-05-20
[BUG] Unauthenticated Route Refresh Endpoint Allows Arbitrary Sandbox Traffic Hijacking


rakshaak29:
### Description

**What happened**:
Both the Sandbox Gateway (`pkg/sandbox-gateway/server/server.go`) and the Sandbox Manager (`pkg/proxy/server.go`) expose an HTTP POST endpoint at `/refresh` on port `7789` (`proxy.SystemPort`). 

This endpoint is used for route synchronization between peers, but it completely lacks authentication or authorization middleware. It binds to the system port unconditionally and directly parses JSON into a `proxy.Route` object, which is then passed to `registry.GetRegistry().Update(route.ID, route)`.

Because port `7789` is exposed within the cluster, any pod—including potentially compromised user sandboxes—can send unauthenticated HTTP POST requests containing spoofed route mappings to the Gateway or Manager. This allows an attacker to manipulate the Envoy routing table, hijacking traffic meant for other sandboxes, causing denial of service, or routing sensitive data to malicious endpoints.

**What you expected to happen**:
The `/refresh` peer-to-peer endpoint should require strict authentication (such as a shared secret token via an `X-Peer-Auth-Token` HTTP header or mutual TLS) so that only legitimate `sandbox-manager` and `sandbox-gateway` instances can synchronize routing data.

**How to reproduce it (as minimally and precisely as possible)**:
1. Deploy OpenKruise Agents in a Kubernetes cluster.
2. Exec into any pod within the cluster (such as a Sandbox container) that has network reachability to the Sandbox Gateway or Manager pod IP on port 7789.
3. Send an unauthenticated POST request to the Gateway:
   ```bash
   curl -X POST http://<sandbox-gateway-pod-ip>:7789/refresh \
        -H "Content-Type: application/json" \
        -d '{"ID":"<target-sandbox-id>", "IP":"<malicious-pod-ip>", "State":"Running"}'
   ```
4.Observe that the Gateway blindly accepts the request (`204 No Content`) and updates its Envoy routing registry to redirect all traffic for `<target-sandbox-id>` to `<malicious-pod-ip>`.

**Anything else we need to know?**: 
This represents a critical security vulnerability (CWE-306: Missing Authentication for Critical Function). It breaks tenant isolation as sandboxes can manipulate infrastructure-level traffic routing.

PR #428, opened 2026-05-20
fix: secure unauthenticated route refresh endpoint and fix PauseTime calculation

rakshaak29:
### PR Title
fix: secure unauthenticated route refresh endpoint and fix PauseTime calculation

This PR addresses a critical security vulnerability and a core functional bug in the Sandbox controller:
1. **Unauthenticated Route Hijacking (Security)**: The `/refresh` endpoint on `proxy.SystemPort` (7789) used for peer-to-peer route synchronization was previously unauthenticated. This allowed any pod in the cluster to send spoofed route mappings to the Gateway/Manager, hijacking traffic meant for other sandboxes. This PR injects a lightweight `X-Peer-Auth-Token` header derived from the `PEER_AUTH_TOKEN` environment variable. If configured, the receiver validates the token and returns a `401 Unauthorized` response on mismatch.
2. **Broken `PauseTime` Behavior (Bug)**: The `PauseTime` calculation in the Sandbox controller previously initialized `requeueAfter` to `0` and then used `min(requeueAfter, pauseDuration)`. This incorrectly evaluated to `0`, preventing the controller from ever requeuing based on the `PauseTime`. This PR corrects the logic to properly assign the `pauseDuration` when `requeueAfter` is `0` or less than the `pauseDuration`.

#427 

**Security Fix Verification:**
1. Send an unauthenticated POST request to the Gateway pod:
   ```bash
   curl -X POST http://<gateway-pod-ip>:7789/refresh -H "Content-Type: application/json" -d '{"ID":"test", "IP":"10.0.0.1", "State":"Running"}
   ```
 2.If PEER_AUTH_TOKEN is set, verify that it is rejected with 401 Unauthorized.
 3.Provide the correct header X-Peer-Auth-Token: <token> and verify it returns 204 No Content.

**PauseTime Fix Verification:**
1.Create a Sandbox with a PauseTime set to 5 minutes in the future.
2.Observe the Sandbox transitions into the Paused state exactly at the defined PauseTime without requiring manual intervention.
For Operators: To fully enable the security fix, cluster administrators must ensure that the PEER_AUTH_TOKEN environment variable is identically injected into the Pod specs of both the sandbox-manager and sandbox-gateway deployments. If left unset, the endpoints fallback to the previous unauthenticated behavior to preserve backward compatibility during upgrades.

### DIFF
--- a/pkg/proxy/server.go
+++ b/pkg/proxy/server.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
 	"sync"
 	"time"
 
@@ -156,6 +157,19 @@ func (s *Server) Stop(ctx context.Context) {
 func (s *Server) handleRefresh(r *http.Request) (web.ApiResponse[struct{}], *web.ApiError) {
 	ctx := r.Context()
 	log := klog.FromContext(ctx)
+
+	// Authenticate the request if PEER_AUTH_TOKEN is set
+	if expectedToken := os.Getenv("PEER_AUTH_TOKEN"); expectedToken != "" {
+		token := r.Header.Get("X-Peer-Auth-Token")
+		if token != expectedToken {
+			log.Error(nil, "unauthorized route refresh request", "token", token)
+			return web.ApiResponse[struct{}]{}, &web.ApiError{
+				Code:    http.StatusUnauthorized,
+				Message: "unauthorized",
+			}
+		}
+	}
+
 	var route Route
 	if err := json.NewDecoder(r.Body).Decode(&route); err != nil {
 		return web.ApiResponse[struct{}]{}, &web.ApiError{
--- a/pkg/proxy/server.go
+++ b/pkg/proxy/server.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/sandbox-gateway/server/server.go
```diff
@@ -207,6 +207,16 @@ func (s *Server) handleRefresh(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	log := klog.FromContext(ctx)
 
+	// Authenticate the request if PEER_AUTH_TOKEN is set
+	if expectedToken := os.Getenv("PEER_AUTH_TOKEN"); expectedToken != "" {
+		token := r.Header.Get("X-Peer-Auth-Token")
+		if token != expectedToken {
+			log.Error(nil, "unauthorized route refresh request", "token", token)
+			http.Error(w, "Unauthorized", http.StatusUnauthorized)
```
HARSHRAJ2789, line 215: `token` here is attacker-controlled input on an unauthenticated endpoint, and a credential whenever it is nearly right (a stale or mistyped operator token). `pkg/identity/AGENTS.md` says never to log security metadata values or credentials. Dropping the field loses nothing: the remote address and the rejection itself are what an operator needs.

pkg/sandbox-gateway/server/server.go
```diff
@@ -207,6 +207,16 @@ func (s *Server) handleRefresh(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	log := klog.FromContext(ctx)
 
+	// Authenticate the request if PEER_AUTH_TOKEN is set
+	if expectedToken := os.Getenv("PEER_AUTH_TOKEN"); expectedToken != "" {
+		token := r.Header.Get("X-Peer-Auth-Token")
+		if token != expectedToken {
+			log.Error(nil, "unauthorized route refresh request", "token", token)
```
HARSHRAJ2789, line 214: `!=` short-circuits on the first differing byte, and this is a shared secret an unauthenticated caller can probe freely. The repo already does the equivalent check the other way in the same component: `pkg/sandbox-gateway/filter/filter.go:159` uses `subtle.ConstantTimeCompare`.

pkg/proxy/server.go
```diff
@@ -156,6 +157,17 @@ func (s *Server) Stop(ctx context.Context) {
 func (s *Server) handleRefresh(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	log := klog.FromContext(ctx)
+
+	// Authenticate the request if PEER_AUTH_TOKEN is set
```
HARSHRAJ2789, line 161: When `PEER_AUTH_TOKEN` is unset this returns to the behaviour #427 reports, silently. Reading it once at construction and logging a warning when it is absent would keep the permissive default for existing deployments while making the open state discoverable, and would move the lookup off the request path.

PR #437, opened 2026-05-22
feat(sandbox-manager): retry clone sandbox failures

AiRanthem:
## Summary
- Add clone sandbox retry handling for sandbox creation, readiness, runtime init, and related failed-sandbox cleanup paths.
- Add API/model support for clone retry extension settings and document the orphan trade-off for retried Create calls.
- Expand unit coverage for clone retry behavior, reservation policies, and sandbox manager extension parsing.

## Test Plan
- [x] /Users/sophon/Bin/go test ./pkg/...


pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -275,23 +280,38 @@ func TryClaimSandbox(ctx context.Context, opts infra.ClaimSandboxOptions, pickCa
...
-			log.Info("sandbox deleted")
+	if effective < 0 {
+		log.Info("the failed sandbox is reserved forever for debugging", "reason", err)
+		return
+	}
+
+	cleanupCtx, cancel := context.WithTimeout(context.Background(), DefaultCleanupTimeout)
+	defer cancel()
+
+	if effective > 0 {
```
furykerry, line 299: effective already guarded at L291, consider remove this if judgement

pkg/sandbox-manager/infra/types.go
```diff
@@ -113,6 +119,16 @@ const (
 	LockTypeSpeculate = LockType("speculate")
 )
 
+// Sentinel values for ReserveFailedSandboxFor.
+const (
+	// ReserveFailedSandboxNever instructs the backend to delete a failed
+	// sandbox immediately instead of keeping it around.
+	ReserveFailedSandboxNever = time.Duration(0)
```
furykerry: these variables is only used in infra test, consider move them to test files.

pkg/servers/e2b/models/extensions.go
```diff
@@ -33,6 +33,23 @@ import (
...
 )
 
+// Sentinel string values accepted by the
+// agents.kruise.io/reserve-failed-sandbox-for extension.
+const (
+	ReserveFailedSandboxValueNever   = "never"
+	ReserveFailedSandboxValueForever = "forever"
+)
+
+// Local copies of infra.ReserveFailedSandboxNever / ReserveFailedSandboxForever.
```
furykerry: these variable is not a duplicate, they're actually used in the code. In the contrary, infra.ReserveFailedSandboxNever / ReserveFailedSandboxForever are copies (used only in test files)

pkg/sandbox-manager/infra/sandboxcr/claim.go
AiRanthem: if effective == 0, L311 will call Delete directly, which reduces 1 api-server call

pkg/servers/e2b/models/extensions.go
AiRanthem: refactored

pkg/sandbox-manager/infra/types.go
AiRanthem: duplicated codes are removed, they are now used

PR #355, opened 2026-05-08
  feat(observability): add memberlist peer-state metrics, key-cache hit/miss counters, and structured-log cleanup

PRAteek-singHWY:
## What this PR does                                                                            
                                                                                                                           
  Second and final PR for #353. Adds operator-facing observability for the **distributed parts** of `sandbox-manager` — the
   parts that PR #354 didn't cover.                                                                                        
                                                                                         
  ## What it handles                                                                                                       
   
  | Sub-piece | What it adds |                                                                                             
  |---|---|                                                                              
  | **(d) Memberlist peer-state metrics** | `sandbox_peer_state{node,state}` gauge (alive/dead) +
  `sandbox_peer_join_duration_seconds` histogram. |
  | **(e) Key-cache hit/miss counters** | `e2b_key_cache_hits_total` and `e2b_key_cache_misses_total`, both labelled by
  storage backend (`mysql` / `secret`) and lookup path (`by_key` / `by_id`). |
  | **(f) Structured-log cleanup** | Converts the 6 remaining unstructured `klog.Infof` / `Errorf` / `Warningf` calls that 
  fire on real events to `klog.InfoS` / `ErrorS` with named fields. |
                                                                                                                           
  ## Relation to PR #354                                                                 
                                                                                                                           
  PR #354 closed the **K8s-facing** gaps — health probes (`/healthz`, `/readyz`), reconcile-loop metrics, and webhook
  admission metrics. Together with this PR, all 6 sub-pieces from issue #353 are addressed.                                
                                                                                         
  | | PR #354 (sub-pieces a, b, c) | This PR (sub-pieces d, e, f) |                                                        
  |---|---|---|                   
  | Scope | K8s-facing | Distributed |                                                                                     
  | Surfaces | health probes, reconcile, webhooks | memberlist, auth cache, log hygiene |                                  
                                  
  ## Why it matters                                                                                                        
                                                                                                                           
  Two parts of `sandbox-manager` have been completely silent in Prometheus until now:                                      
                                                                                                                           
  - **Memberlist gossip.** A peer can drop out of the cluster and nothing tells the operator — they only notice when       
  traffic skews.                                                                                                           
  - **API key cache.** Every authenticated request runs through two layers of in-memory cache before hitting the DB, but
  there's no signal at all about whether the cache is doing useful work.                                                   
                                  
  This PR exposes both.                                                                                                    
                                                                                                                           
  ## Implementation notes
                                                                                                                           
  ### (d) Memberlist metrics                                                             
[...]

PR #347, opened 2026-05-07
fix(server): add missing ReadTimeout, WriteTimeout and IdleTimeout to HTTP servers

PRAteek-singHWY:
# fix(server): add missing ReadTimeout, WriteTimeout and IdleTimeout to HTTP servers

## Summary
Adds missing HTTP server timeouts to all three internal HTTP servers to protect against Slowloris-style denial-of-service attacks.

## Problem
All three HTTP servers — `sandbox-manager`, `sandbox-gateway`, and `proxy` — configured `ReadHeaderTimeout` but omitted `ReadTimeout`, `WriteTimeout`, and `IdleTimeout`. A slow client can exploit this by holding a connection open indefinitely, trickling data byte-by-byte, eventually exhausting all available server file descriptors.

## Solution
Added `ReadTimeout`, `WriteTimeout`, and `IdleTimeout` to all three `http.Server` structs. The values are defined as shared constants in `pkg/sandbox-manager/consts/consts.go`:

| Timeout | Value | Purpose |
|---|---|---|
| `ReadTimeout` | 30s | Max time to read the full request body |
| `WriteTimeout` | 60s | Max time to write the full response |
| `IdleTimeout` | 120s | Max idle time for keep-alive connections |

## Files Changed
- `pkg/sandbox-manager/consts/consts.go` — Added `HTTPReadTimeout`, `HTTPWriteTimeout`, `HTTPIdleTimeout` constants
- `pkg/servers/e2b/core.go` — sandbox-manager HTTP server
- `pkg/sandbox-gateway/server/server.go` — sandbox-gateway peer server
- `pkg/proxy/server.go` — proxy HTTP server

## Checklist
- [x] Timeouts added to all three HTTP servers
- [x] Shared constants defined centrally to avoid duplication
- [x] Build passes (`go build ./...`)
- [x] All tests pass

**fixes #346**

PR #440, opened 2026-05-22
fix(pagination): keep list pages stable so sandboxes with the same claim time aren't skipped

### LINKED ISSUE #439: [BUG] Listing sandboxes can silently hide some of them when they share the same claim time
# [BUG] Listing sandboxes can silently hide some of them when they share the same claim time

**What's going on**

When you list sandboxes (`GET /v2/sandboxes`) or snapshots, the results come back one page at a time. Paging is handled by `utils.Paginator` in `pkg/utils/sandbox-manager/pagination.go`.

The way it remembers "where the last page ended" is by saving the **timestamp of the last item** as the `nextToken`. The next page then starts at the first item whose timestamp is **strictly after** that token:

```go
// pkg/utils/sandbox-manager/pagination.go
nextToken = p.GetKey(paged[len(paged)-1])          // the last item's timestamp
...
startIdx = sort.Search(len(items), func(i int) bool {
    return p.GetKey(items[i]) > p.NextToken         // strictly greater
})
```

The problem is that this timestamp is **not unique**. Sandboxes are sorted by their `claim-timestamp` annotation, which is written as `time.Now().Format(time.RFC3339)` — that only has **one-second precision** (see `pkg/sandbox-manager/infra/sandboxcr/claim.go`). Snapshots have the same issue (they sort by checkpoint creation time).

So if several sandboxes were claimed in the same second, they all share the exact same timestamp. And because the next page skips to "strictly after this timestamp," any of those same-second sandboxes that didn't fit on the page get **jumped over completely** — they never show up on any page.

**A simple example**

Say `limit = 2` and three sandboxes were all claimed in the same second (call it `t1`):

1. Page 1 returns the first two of them. `nextToken` is set to `t1`.
2. Page 2 asks for "the first item after `t1`" — but the third `t1` sandbox isn't *after* `t1`, it *is* `t1`. So it's skipped, and page 2 comes back empty.
3. That third sandbox is now invisible. The user thinks they have 2 sandboxes when they really have 3.

**Why this actually matters**

This is silent data loss in a public API. A sandbox that's really running (and costing resources) just disappears from the list, and nothing tells the user. Anything built on top of the list — dashboards, cleanup scripts, billing/quota checks — inherits the same blind spot.

And it's not a rare edge case:

- The max page size is 100 (`MaxListLimit`), and clients often ask for fewer, so paging kicks in for any decently sized account.
- Sandboxes are claimed in **batches** (many at once), so a bunch of them landing in the same one-second window is normal, not unusual. That's exactly when the bug bites.

It gets worse the more sandboxes you have and the faster you claim them — i.e. precisely under load.

**How to reproduce**

1. Claim 3 or more sandboxes for the same user within one second (a normal batch claim does this).
2. Call `GET /v2/sandboxes?limit=2`.
3. Keep following `x-next-token` until there are no more pages.
4. Count what you got back — it's fewer than what actually exists. The extra same-second sandboxes are missing.

**Suggested fix**

Make the "where did the last page end" marker unique by adding a tiebreaker, so two items with the same timestamp can still be told apart:

- Sort and page by the pair `(timestamp, unique-id)` instead of just the timestamp — sandbox ID for sandboxes, checkpoint ID for snapshots. Since the pair is unique, no item can fall through the crack at a page boundary anymore.
- The continuation token then carries both parts. It needs to be encoded (base64) because the combined value isn't safe to put in an HTTP header as-is. Callers don't need to care — the token is opaque to them.

This stays inside the existing paginator plus the two list call sites. It doesn't change *what* gets listed or the sort order (still by claim/creation time) — it just stops items from being skipped.

This is a standalone, pre-existing bug in the generic pagination helper — not specific to any driver, workload, or roadmap item. I checked the open and closed issues and PRs and found nothing about pagination, the list cursor, or `nextToken` (the `NextToken` field is even marked `// not implemented`), so this looks like a genuine gap rather than a duplicate. Happy to send a fix.

**Environment**

- Kubernetes version: N/A — this is a code-level bug
- Affects: `GET /v2/sandboxes` and the snapshots list, for any account large enough to need paging; worse with more sandboxes and higher claim rates

PRAteek-singHWY:
## fix(pagination): keep list pages stable so sandboxes with the same claim time aren't skipped

fixes #439

### Ⅰ. What this PR does

When you list sandboxes or snapshots, the results are paged by `utils.Paginator`.
It remembered where a page ended by saving the last item's timestamp as the
`nextToken`, and started the next page at the first item *strictly after* that
timestamp.

The catch: that timestamp isn't unique. Sandboxes sort by their `claim-timestamp`
annotation, which is `time.Now().Format(time.RFC3339)` — only one-second
precision. Snapshots sort by checkpoint creation time, same story. So when more
than `limit` items were claimed in the same second and a page boundary landed in
the middle of them, the leftovers got skipped and never appeared on any page.
Since sandboxes are claimed in batches, same-second groups are common, so this
shows up as real, running sandboxes silently missing from the list — especially
under load.

The fix is to make the page marker unique by adding a tiebreaker:

- `Paginator` gets an optional `GetUniqueKey`. When it's set, sorting and the
  page cursor use the pair `(GetKey, GetUniqueKey)` instead of just `GetKey`.
  Because the pair is unique, two items with the same timestamp can no longer
  straddle a page boundary and get lost.
- The continuation token now carries both parts, so it's base64url-encoded (the
  combined value uses a separator that isn't valid in an HTTP header). A token
  that doesn't decode is just used as-is, so a stale/old token degrades to a
  best-effort resume instead of erroring the whole call.
- The two list call sites pass a tiebreaker: sandbox ID for sandboxes, checkpoint
  ID for snapshots.

If `GetUniqueKey` isn't set, everything — order, behavior, and token format —
stays exactly as before, so existing callers are untouched.

fixes #<issue-number>

### Ⅲ. How to verify it

- `go test ./pkg/utils/sandbox-manager/...` — green, including two new tests:
  - `TestPaginator_Apply_DuplicateSortKeys`: pages over five items where three
    share one timestamp (`limit=2`) and checks every item comes back exactly
    once, none skipped. This fails on the old timestamp-only cursor.
  - `TestPaginator_Apply_UniqueKeyTokenRoundTrips`: the token is header-safe (no
    raw separator) and the next page decodes it back to the same position.
  - the existing `TestPaginator_Apply` cases still pass unchanged.
- `go test ./pkg/servers/e2b/...` — `TestListSandboxes_Pagination` and
  `TestListSnapshots` pass; the "start with nextToken" case now uses the encoded
  token.
- `go build ./...`, `go vet`, and `gofmt` are all clean.

### Ⅳ. Notes for reviewers

Contained change: the generic paginator plus the two e2b list call sites. It
doesn't change what's listed or the sort order (still claim/creation time) — only
the page boundary marker is made unique. The token is opaque to clients, so the
encoding change needs nothing on the client side.

### DIFF
--- a/pkg/servers/e2b/list.go
+++ b/pkg/servers/e2b/list.go
@@ -143,6 +143,9 @@ func (sc *Controller) ListSandboxes(r *http.Request) (web.ApiResponse[[]*models.
 		GetKey: func(sbx infra.Sandbox) string {
 			return sbx.GetAnnotations()[agentsv1alpha1.AnnotationClaimTime]
 		},
+		GetUniqueKey: func(sbx infra.Sandbox) string {
+			return sbx.GetSandboxID()
+		},
 	})
 	var headers map[string]string
 	if nextToken != "" {
@@ -284,6 +287,9 @@ func (sc *Controller) listCheckpointsForUser(ctx context.Context, user *models.C
 		NextToken: nextTokenParam,
 		GetKey: func(cp infra.CheckpointInfo) string {
 			return cp.CreationTimestamp // Sort by creation timestamp
+		},
+		GetUniqueKey: func(cp infra.CheckpointInfo) string {
+			return cp.CheckpointID
 		},
 		Filter: filter,
 	})
--- a/pkg/servers/e2b/list_test.go
+++ b/pkg/servers/e2b/list_test.go
@@ -18,6 +18,7 @@ package e2b
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"net/http"
 	"sort"
@@ -456,8 +457,15 @@ func TestListSandboxes_Pagination(t *testing.T) {
 			},
 		},
 		{
-			name:  "start with nextToken",
-			query: map[string]string{"limit": "2", "nextToken": "2024-01-01T00:00:02Z"},
+			// The page token is opaque: it encodes the sort key (claim time) plus a
+			// unique tiebreaker (sandbox ID) so equal-timestamp sandboxes are not
+			// skipped at a page boundary. Resuming from the token for sandbox idx 1
+			// must return the following sandboxes (idx 2,3).
+			name: "start with nextToken",
+			query: map[string]string{
+				"limit":     "2",
+				"nextToken": base64.RawURLEncoding.EncodeToString([]byte(claimTimes[1] + "\x00" + createdSandboxIDs[1])),
+			},
 			pages: []pageExpectation{
 				{count: 2, hasNextToken: true, startedAtIdxs: []int{2, 3}},
 			},
--- a/pkg/utils/sandbox-manager/pagination.go
+++ b/pkg/utils/sandbox-manager/pagination.go
@@ -17,8 +17,15 @@ limitations under the License.
 package utils
 
 import (
+	"encoding/base64"
 	"sort"
 )
+
+// cursorSep separates the sort key from the unique tiebreaker inside a composite
+// cursor. It is the NUL byte so that the tiebreaker always orders after the sort
+// key but before any printable continuation of a longer key (e.g. "1\x00z" sorts
+// before "10\x00a"), keeping the composite order consistent with (key, unique).
+const cursorSep = "\x00"
 
 // Paginator pages the selected objects.
 type Paginator[T any] struct {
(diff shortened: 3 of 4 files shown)


PRAteek-singHWY:
Hi @furykerry @AiRanthem @zmberg 
 This fixes a silent data-loss bug in list pagination: the page cursor was just the claim/creation timestamp
  (1-second RFC3339), so when more than limit sandboxes shared the same second at a page boundary, the extras got
  skipped and never showed up on any page. The fix adds a unique tiebreaker (sandbox/checkpoint ID) so the cursor
  pages on (timestamp, id) instead of timestamp alone - no behavior or sort-order change otherwise, just no more
  dropped items. Added regression tests covering the duplicate-timestamp case.

PRAteek-singHWY:
Hi @furykerry @AiRanthem @zmberg 

Thanks so much for all the guidance so far, really appreciate the time you put into reviewing  contributions.

Whenever you get a chance, I'd be grateful if you could take a look at the work in PRs I've opened 

#333
#337 
#342 
#347 
#354 
#355 
#409 
#414 
#440


No rush at all, I know everyone's busy with ongoing mentorship. I just wanted to gather them in one place so they're easy to find, and I'm happy to address any feedback or make changes wherever needed. Please let me know if anything should be split up, reworked, or if there's a different approach you'd prefer.

Happy to iterate based on your review and suggestions.

Thank you.

PR #445, opened 2026-05-25
followup updates for egress control

Kuromesi:
This is a followup fix for egress control. The changes can be concluded as:
1. rename egress control to traffic proxy
2. not override user-defined pod labels and annotations
3. fixed some health probe issues

### DIFF
--- a/pkg/utils/sidecarutils/sidecar_config_inject.go
+++ b/pkg/utils/sidecarutils/sidecar_config_inject.go
@@ -28,20 +28,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
-	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
 	"github.com/openkruise/agents/pkg/sandbox-manager/consts"
 	"github.com/openkruise/agents/pkg/utils"
 	"github.com/openkruise/agents/pkg/utils/webhookutils"
 )
-
-func IsRuntimeEnabled(sandbox *agentsv1alpha1.Sandbox, runtimeName string) bool {
-	for _, runtime := range sandbox.Spec.Runtimes {
-		if runtime.Name == runtimeName {
-			return true
-		}
-	}
-	return false
-}
 
 func fetchInjectionConfiguration(ctx context.Context, cli client.Client) (map[string]string, error) {
 	logger := logf.FromContext(ctx)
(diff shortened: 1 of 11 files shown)


pkg/utils/sidecarutils/utils.go
```diff
@@ -47,14 +48,20 @@ func InjectSandboxRuntimes(ctx context.Context, sandbox *agentsv1alpha1.Sandbox,
...
+	runtimes := sets.New[string]()
 	for _, runtime := range sandbox.Spec.Runtimes {
-		// should we return an error when inject config not exists?
-		runtimeInjectConfig, err := parseInjectConfig(ctx, runtime.Name, injectConfigMap)
+		if runtime.Name != "" {
+			runtimes.Insert(runtime.Name)
+		}
+	}
+
+	for runtime := range runtimes {
```
furykerry: sets.New[string]() is backed by a Go map, so iteration order is non-deterministic. If a sandbox has multiple generic (non-csi, non-agent-runtime) runtimes, the order in which they are injected into the pod (containers, init containers, volumes, labels, annotations appended) will vary between runs. While injectConflicts prevents duplicates, the ordering of injected resources in the pod spec will be non-deterministic. Consider keeping use sandbox.Spec.Runtimes for iteration, and skip the iteration if duplicated

pkg/utils/sidecarutils/sidecar_config_inject.go
```diff
@@ -167,45 +157,63 @@ func setAgentRuntimeContainer(ctx context.Context, podSpec *corev1.PodSpec, conf
 }
 
 func applyInjectionTemplate(ctx context.Context, pod *corev1.Pod, config SidecarInjectConfig) error {
+	if pod.Labels == nil {
```
furykerry, line 160: consider remove un-used context.Context and error variable

pkg/utils/sidecarutils/sidecar_config_inject.go
```diff
@@ -167,45 +157,63 @@ func setAgentRuntimeContainer(ctx context.Context, podSpec *corev1.PodSpec, conf
...
+
+	return nil
+}
+
+// injectConflicts checks if injection config conflicts with the pod. Only configuration below are checked:
+// 1. Containers
+// 2. Init containers
+// 3. Volumes
+// Annotations and Labels are not checked and user-defined values may override the injected values.
+func injectConflicts(pod *corev1.Pod, config SidecarInjectConfig) error {
```
furykerry: consider rename injectConflicts to checkInjectConflicts

PR #449, opened 2026-05-25
fix: return bad request for invalid api key creation

AiRanthem:
## Summary
- Validate create API key requests before they reach key storage.
- Return HTTP 400 when the request body is valid JSON but missing the required API key name.
- Cover both the normal middleware path and the handler fallback path with unit tests.

## Root Cause
`CheckCreateAPIKeyPermission` decoded `{}` successfully and defaulted the team name, but did not validate the required `name` field. The request then reached `CreateKey`, whose validation returned a plain error that `CreateAPIKey` wrapped as HTTP 500.

## Test Plan
- `GOPATH=/Users/sophon/sdk/gopath GOROOT=/Users/sophon/sdk/gopath/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64 /Users/sophon/Bin/go test ./pkg/servers/e2b -run 'TestCreateAPIKey$|TestCreateAPIKeyPermissionMiddleware$' -count=1`
- `GOPATH=/Users/sophon/sdk/gopath GOROOT=/Users/sophon/sdk/gopath/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64 /Users/sophon/Bin/go test ./pkg/servers/e2b -count=1`

### DIFF
--- a/pkg/servers/e2b/routes.go
+++ b/pkg/servers/e2b/routes.go
@@ -152,6 +152,9 @@ func (sc *Controller) CheckCreateAPIKeyPermission(ctx context.Context, r *http.R
 			Message: err.Error(),
 		}
 	}
+	if apiErr := validateCreateAPIKeyRequest(&request); apiErr != nil {
+		return ctx, apiErr
+	}
 
 	callerTeam := keys.TeamForKey(user)
 	targetTeamName := request.TeamName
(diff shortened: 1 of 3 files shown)


pkg/servers/e2b/routes.go
```diff
@@ -152,6 +152,9 @@ func (sc *Controller) CheckCreateAPIKeyPermission(ctx context.Context, r *http.R
 			Message: err.Error(),
 		}
 	}
+	if apiErr := validateCreateAPIKeyRequest(&request); apiErr != nil {
```
furykerry, line 155: if we had already validate the request in checkCreateAPIKeyPermission, is it still necessary to check in CreateAPIKey?

PR #450, opened 2026-05-25
update security provider handler

BH4AWS:

### DIFF
--- a/pkg/sandbox-manager/infra/sandboxcr/claim.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/claim.go
@@ -237,8 +237,12 @@ func TryClaimSandbox(ctx context.Context, opts infra.ClaimSandboxOptions, pickCa
 		metrics.SecurityToken, err = issueSecurityToken(ctx, sbx, opts.SecurityToken)
 		if err == nil {
 			metrics.Total += metrics.SecurityToken
-			// 4.1: to record security token refresh status in sandbox annotations
-			if err = recordSecurityTokenRefreshStatus(sbx, opts); err != nil {
+			// 4.1: to record security token refresh status in sandbox annotations.
+			// At this point modifyPickedSandbox has already persisted the locking
+			// patch, so additional annotation mutations on sbx.Sandbox would only
+			// live in memory. We patch via the apiserver here to actually persist
+			// the refresh status, and keep sbx.Sandbox in sync with the patched object.
+			if err = recordSecurityTokenRefreshStatus(ctx, cache.GetClient(), sbx, opts); err != nil {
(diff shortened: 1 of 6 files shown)


pkg/utils/runtime/runtime.go
```diff
@@ -162,6 +166,166 @@ func RunCommandWithRuntime(ctx context.Context, args RunCmdFuncArgs) (RunCommand
...
+const (
+	defaultRuntimeWriteTimeout  = 10 * time.Second
+	defaultRuntimeFilesUsername = "root"
+	runtimeFilesFieldName       = "file"
+)
+
+// runtimeFilesHTTPClient is the package-level HTTP client used by WriteFileWithRuntime.
+// It is a variable rather than a constant so tests can substitute their own transport
+// without monkey-patching net/http.DefaultClient.
+var runtimeFilesHTTPClient = &http.Client{Timeout: defaultRuntimeWriteTimeout}
```
furykerry: runtimeFilesHTTPClient is initialized with Timeout: defaultRuntimeWriteTimeout (10 s). WriteFileWithRuntime also creates a per-request context.WithTimeout(ctx, timeout). The effective timeout is the minimum of both, so any args.Timeout value > 10 s is silently ignored 

— the client-level timeout kills the request first.This is inconsistent with RunCommandWithRuntime (runtime.go:108), which uses http.DefaultClient (no built-in timeout) and relies solely on context timeout.

Consider: Remove Timeout from the package-level client and let the context deadline control it

pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -593,23 +597,32 @@ func modifyPickedSandbox(sbx *Sandbox, lockType infra.LockType, opts infra.Claim
...
-		AccessTokenExpiration: opts.SecurityToken.AccessTokenExpiration,
-	})
+	raw, err := identity.EncodeTokenRefreshStatus(identity.BuildTokenRefreshStatus(&opts.SecurityToken.TokenResponse))
 	if err != nil {
 		return fmt.Errorf("failed to marshal token refresh expiration status: %w", err)
 	}
-	annotations := sbx.GetAnnotations()
-	if annotations == nil {
-		annotations = make(map[string]string, 1)
+	patchBase := sbx.Sandbox.DeepCopy()
```
furykerry: patchBase deepcopy is not necessary

pkg/utils/runtime/runtime.go
BH4AWS: fixed

pkg/sandbox-manager/infra/sandboxcr/claim.go
BH4AWS: fixed

PR #472, opened 2026-05-28
Generate E2B SDK-compatible API keys

lxs137:

Generate E2B SDK-compatible API keys with e2b_ prefix.

[New E2B SDK (>= 2.25.0) versions](https://github.com/e2b-dev/E2B/pull/1356) validate API key format client-side, requiring
keys to match `^e2b_[0-9a-f]+$`. Previously generated keys used raw
UUID format which is now rejected by the SDK.

Changes:

1. Add GenerateAPIKey() using crypto/rand to produce "e2b_" + 40 hex
chars, matching the official E2B key format.
2. Update both secret and mysql key storage to use the new generator.
3. Add bypass_key_validation option to patch_e2b() for backward
compatibility with legacy keys.

```
[AUTH_FAILED(1002)] Failed to create sandbox: credentials rejected. Che
Ck E2B_API_KEY. Original: Invalid API key format: expected "e2b_" followed by hex characters (e.g. "e2b_0000000000000000000000000000000000000000"). Visit the API Ke
ys tab at https://e2b.dev/dashboard?tab=keys to get your API key.
```

### DIFF
--- a/cmd/sandbox-manager/main.go
+++ b/cmd/sandbox-manager/main.go
@@ -24,7 +24,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/google/uuid"
 	"github.com/spf13/pflag"
 	zapRaw "go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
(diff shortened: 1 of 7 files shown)


mishushakov:
we are relaxing this bit, please see
https://github.com/e2b-dev/E2B/pull/1360

furykerry:
@lxs137 thanks for your patch, however the patch does not address the compatibility problems of existing keys already saved in your cluster, we cannot force the users to change the key which may breaks existing sandboxes. Another contributor had submit  a more complete patch   https://github.com/openkruise/agents/pull/473, feel free to comment in this patch

PR #473, opened 2026-05-28
feat(e2b): add E2B >=v2.25.0 SDK-compatible API key encoding layer

AiRanthem:
## Summary

- E2B SDK `>=2.25.0` rejects API keys not matching `e2b_[0-9a-f]+`, breaking legacy UUID / `admin-*` keys before they ever reach the server. This PR adds a reversible compatibility encoding so server-issued and server-accepted keys satisfy the SDK validator without changing Secret/MySQL storage semantics.
- `POST /api-keys` now returns the SDK-compatible form via `PresentCreatedAPIKey`; storage still holds the raw key, MySQL still HMACs the raw key.
- `CheckApiKey` canonicalizes the presented key (raw or encoded) before lookup, so existing raw keys keep working unchanged.
- New `GET /api-keys/compatible` returns the SDK-compatible form of the currently authenticated credential (for clients that authenticated with a raw key and need to upgrade).
- Authentication failure messages and key-create logs no longer reflect raw or encoded keys.

Design details: `docs/specs/2026-05-28-e2b-api-key-compat-design.md`.

## Test plan

- [ ] `go test ./pkg/servers/e2b/... ./pkg/servers/e2b/keys/...`
- [ ] Verify legacy raw `X-API-KEY` still authenticates against `/sandboxes` and `/api-keys`.
- [ ] Verify an SDK-encoded `X-API-KEY` (returned from `POST /api-keys` or `GET /api-keys/compatible`) authenticates and routes to the same identity as its decoded raw form.
- [ ] Verify `GET /api-keys/compatible` returns the same encoded value when authenticated with either the raw or encoded form.
- [ ] Verify unauthorized responses do not echo back the presented key.

### DIFF
--- a/docs/specs/2026-05-28-e2b-api-key-compat-design.md
+++ b/docs/specs/2026-05-28-e2b-api-key-compat-design.md
@@ -1,0 +1,224 @@
+# E2B SDK-Compatible API Key Design
+
+## Context
+
+E2B SDK `>=2.25.0` validates API keys before sending requests. The SDK
+accepts only keys matching:
+
+```text
+e2b_[0-9a-f]+
+```
+
+`sandbox-manager` historically stores and authenticates raw OpenKruise Agents
+keys. Those keys are commonly UUID strings or custom admin keys, such as
+`admin-987654321`, and therefore fail the new SDK-side validation before the
+server can authenticate them.
+
+The compatibility layer must make server-issued and server-accepted keys pass
+the E2B SDK format without changing storage semantics. Existing Secret and
+MySQL backends must continue to store, index, and hash the original raw key.
+
+## Goals
+
+- Accept existing raw API keys exactly as before.
+- Accept SDK-compatible encoded API keys and map them to the same stored raw key.
+- Return SDK-compatible keys from `POST /api-keys`.
+- Provide a protected endpoint for clients to retrieve the SDK-compatible form
+  of the currently authenticated key.
+- Keep Secret storage payloads unchanged.
+- Keep MySQL schema and HMAC input unchanged.
+- Avoid reflecting raw or compatible API keys in authentication failure
+  responses or logs.
+
+## Non-Goals
+
+- Do not change the `keys.KeyStorage` interface.
+- Do not change key generation. Backends still generate raw UUID/custom keys.
+- Do not migrate existing Secret data or MySQL rows.
+- Do not store encoded keys as canonical credentials.
+- Do not make generic `e2b_...` keys valid unless they use the OpenKruise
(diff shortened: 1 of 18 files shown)


pkg/servers/e2b/routes_test.go
```diff
@@ -79,19 +123,26 @@ func TestCheckApiKey_WithRealSetup(t *testing.T) {
 			expectCtxUser: true,
 			expectedUser:  regularUser,
 		},
+		{
```
furykerry, line 126: consider add a case where the key starts with e2b_, but without openkruise encoded format

pkg/servers/e2b/keys/compat.go
```diff
@@ -0,0 +1,138 @@
...
+		e2bSDKCompatMagic,
+		e2bSDKCompatVersion,
+		len(rawBytes),
+		hex.EncodeToString(rawBytes),
+		e2bSDKCompatChecksum(rawBytes),
+	)
+}
+
+// DecodeFromE2BSDKCompatible decodes an OpenKruise Agents-compatible E2B SDK key.
+func DecodeFromE2BSDKCompatible(apiKey string) (string, bool) {
```
furykerry, line 67: consider add a fuzz tests that use encode+decode roundtrip, and ensure no golang panic exists

docs/specs/2026-05-28-e2b-api-key-compat-design.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Key Utility API
+
+Add `pkg/servers/e2b/keys/compat.go` with these helpers:
+
+```go
+func IsE2BSDKCompatible(apiKey string) bool
+func EncodeForE2BSDK(raw string) string
+func DecodeFromE2BSDKCompatible(apiKey string) (string, bool)
+func CanonicalizeForLookup(apiKey string) string
```
furykerry: plz update the func name

pkg/servers/e2b/routes_test.go
AiRanthem: Added in TestCheckApiKey_CanonicalizesE2BSDKCompatibleKeys

pkg/servers/e2b/keys/compat.go
AiRanthem: Added

docs/specs/2026-05-28-e2b-api-key-compat-design.md
AiRanthem: All stale names are updated

PR #475, opened 2026-05-28
feat(controller): add SecurityTokenRefreshReconciler for proactive token rotation

BH4AWS:
…ken rotation

Introduce a controller-runtime reconciler that proactively rotates the
security token of claimed sandboxes shortly before AccessTokenExpiration,
so the runtime never observes an expired credential.

The controller is gated by SecurityIdentityProvider feature gate and
auto-skips when the Sandbox CRD is not present, keeping clusters without
an identity provider behave exactly like before.

Core design
-----------
* Predicate-driven enqueue (predicate.go): only sandboxes carrying the
  claimed label AND a non-empty TokenRefreshStatus annotation flow into
  the workqueue, avoiding hot-loops on freshly created or already
  released sandboxes.
* Reconcile is policy-only: it parses AccessTokenExpiration, schedules
  refreshAt = expireAt - (refreshLeadTime ± jitter) and either requeues
  or hands over to the Refresher. Default lead time is 30m with ±10%
  jitter; failure inside the lead window requeues every 1m, giving
  ~30 retry chances before T-0.
* core.Refresher encapsulates issue / propagate / annotation-patch as a
  single side-effect surface, so the reconciler stays focused on timing.
  A defaultRefresher composes identity.IssueSandboxToken,
  identity.PropagateSandboxToken and a controller-runtime client patch.

Extension surface
-----------------
* SetupHook + RegisterSetupHook expose a downstream extension point at
  the tail of Add(). Hooks run after the reconciler is wired with the
  manager and may attach extra peer Runnables (e.g. CA-bundle sync,
  metrics emitters) without forking this package. Errors are wrapped
  with a hook-index prefix and short-circuit further hooks.

Identity helper symmetry
------------------------
* Add identity.PropagateSandboxToken as the symmetric twin of
  identity.IssueSandboxToken. Both helpers carry the uniform
  klog skeleton (action, propagatorCount, cost) so claim flow and
  refresh controller share the same observability surface.
* Migrate sandbox-manager/infra/sandboxcr claim flow to the new helper,
  collapsing the inline propagator block into a single call.

Configurability
---------------
Three flags allow operators to tune behaviour without rebuilding:
  --security-token-refresh-workers       (default 10)
  --security-token-refresh-lead-time     (default 30m)
  --security-token-refresh-jitter-ratio  (default 0.1)
  --security-token-refresh-retry-after   (default 1m)

Tests
-----
* Reconcile timing matrix: not-yet-due / due-now / inside-window /
  refresher-failure / malformed-annotation / empty-status / not-claimed.
* Predicate matrix: claim flips, annotation diff vs equal, deletion.
* Refresher table tests: issue error, propagate error verbatim, patch
  failure, happy path with annotation persistence assertion.
* SetupHook table tests: nil filter, registration order, first-error
  short-circuit with wrapped error.
* PropagateSandboxToken: success / provider error verbatim / nil resp.

### DIFF
--- a/pkg/controller/securitytokenrefresh/core/refresher_test.go
+++ b/pkg/controller/securitytokenrefresh/core/refresher_test.go
@@ -1,0 +1,189 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package core
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
+	"github.com/openkruise/agents/pkg/identity"
+	"github.com/openkruise/agents/pkg/utils"
+)
+
+// stubIdentityProvider is a hand-rolled IdentityProvider that returns the configured
+// TokenResponse / errors. It is installed via identity.RegisterProvider() for the
+// duration of a single sub-test.
+type stubIdentityProvider struct {
+	resp           *identity.TokenResponse
+	issueErr       error
+	propagateErr   error
+	propagateCalls int
+	issueCalls     int
+}
+
+func (s *stubIdentityProvider) IssueToken(_ context.Context, _ identity.TokenRequest) (*identity.TokenResponse, error) {
+	s.issueCalls++
+	if s.issueErr != nil {
+		return nil, s.issueErr
+	}
+	return s.resp, nil
+}
+
+func (s *stubIdentityProvider) PropagateSecurityToken(_ context.Context, _ *agentsv1alpha1.Sandbox, _ *identity.TokenResponse) error {
+	s.propagateCalls++
+	return s.propagateErr
+}
+
+// withProvider installs stub as the global identity provider for the test.
+// It is intentionally not parallelisable: the registry is a process-wide singleton.
+func withProvider(t *testing.T, stub *stubIdentityProvider) {
+	t.Helper()
+	identity.RegisterProvider(stub)
+}
+
+func newScheme(t *testing.T) *runtime.Scheme {
+	t.Helper()
+	s := runtime.NewScheme()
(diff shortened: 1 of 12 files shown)


pkg/controller/securitytokenrefresh/core/refresher_test.go
```diff
@@ -0,0 +1,241 @@
...
+				AccessTokenExpiration: "2099-01-01T00:00:00Z",
+			},
+			expect: identity.TokenRefreshStatus{
+				AccessTokenExpiration: "2099-01-01T00:00:00Z",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.expect, identity.BuildTokenRefreshStatus(tt.input))
```
furykerry: [core/refresher_test.go](https://github.com/openkruise/agents/pull/475/files#diff-...refresher_test.go) lines ~180–241 contain a TestBuildTokenRefreshStatus that exercises identity.BuildTokenRefreshStatus — a function already thoroughly tested in pkg/identity/security_token_helper_test.go. Testing another package's exported function from a different package's test file is unusual and adds maintenance burden.

pkg/controller/securitytokenrefresh/token_refresh_controller.go
```diff
@@ -0,0 +1,381 @@
...
+	if s == "" {
+		return time.Time{}, fmt.Errorf("accessTokenExpiration is empty")
+	}
+	return time.Parse(time.RFC3339, s)
+}
+
+// clampRequeue ensures the RequeueAfter passed to controller-runtime is at
+// least minRequeueAfter, so a near-zero or negative computed delay never
+// degenerates into an immediate-retry loop.
+func clampRequeue(d time.Duration) time.Duration {
```
furykerry, line 364: If the identity provider issues a token with a very short TTL (e.g., 10 minutes) and the refresh lead time is 30 minutes, then nextExpire - now - leadTime will be negative, and clampRequeue floors it to minRequeueAfter (1 second). This causes an immediate re-attempt to refresh a token that was just refreshed, potentially creating a tight loop until the provider issues a longer-lived token. While clampRequeue prevents an infinite zero-duration loop, the 1-second requeue is still aggressive. Consider adding a guard: if nextExpire.Sub(now) < refreshLeadTime, use a shorter effective lead time (e.g., nextExpire.Sub(now) / 2) instead of the full configured lead time, and document the design intent.

pkg/controller/securitytokenrefresh/token_refresh_controller.go
```diff
@@ -0,0 +1,381 @@
...
+// gated by the SecurityIdentityProvider feature gate so that clusters without
+// an identity provider behave exactly like before.
+package securitytokenrefresh
+
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"math/rand"
```
furykerry: The existing claim.go imports math/rand/v2, but the new controller uses math/rand. While functionally correct, this is inconsistent with the rest of the codebase's direction.

pkg/controller/securitytokenrefresh/core/refresher_test.go
BH4AWS: fixed

pkg/controller/securitytokenrefresh/token_refresh_controller.go
BH4AWS: fixed

pkg/controller/securitytokenrefresh/token_refresh_controller.go
BH4AWS: fixed

PR #491, opened 2026-06-03
feat(sandbox): add Pausing transitional phase to sandbox lifecycle

zmberg:
## Summary
- Introduce `SandboxPausing` as an intermediate phase between `Running` and `Paused`, making the lifecycle `Running -> Pausing -> Paused` so that "pause in progress" is distinguishable from "fully paused".
- Simplify resumability checks: `Paused` phase now always means fully paused, removing the need to inspect the `Paused` condition.
- Refine `sandbox_status_abnormal` metrics to detect stuck `Pausing`/`Resuming` states using a 5-minute time-based threshold instead of simple phase/condition mismatch.

## Test plan
- [x] Unit tests updated for `common_control`, `rateLimiter`, `sandbox_controller`, `metrics`, and `utils`
- [ ] Verify `make generate manifests` passes after API type change
- [ ] E2E: pause a running sandbox and confirm it transitions through `Pausing` -> `Paused`
- [ ] E2E: resume a paused sandbox and confirm `Resuming` -> `Running`
- [ ] Verify `sandbox_status_abnormal` metric fires only after 5min stuck threshold

### DIFF
--- a/pkg/controller/sandbox/sandbox_controller.go
+++ b/pkg/controller/sandbox/sandbox_controller.go
@@ -243,8 +243,11 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (cr
 		requeueAfter, err = r.getControl(args.Pod).EnsureSandboxRunning(ctx, args)
 	case agentsv1alpha1.SandboxRunning:
 		err = r.getControl(args.Pod).EnsureSandboxUpdated(ctx, args)
-	case agentsv1alpha1.SandboxPaused:
+	case agentsv1alpha1.SandboxPausing:
 		err = r.EnsureSandboxPaused(ctx, args)
+	case agentsv1alpha1.SandboxPaused:
+		// Fully paused, nothing to do (resume is handled by calculateStatus)
+		klog.V(5).InfoS("sandbox is fully paused, skipping reconcile", "sandbox", klog.KObj(box))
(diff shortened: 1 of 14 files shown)


pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -349,12 +352,12 @@ func calculateStatus(args core.EnsureFuncArgs) (*agentsv1alpha1.SandboxStatus, b
 			return newStatus, true
 		}
 
-		// If it is paused, first set the sandbox to the Paused state.
-		// To prevent loss of state information, the state immediately before Paused must currently be Running.
+		// If it is paused, first set the sandbox to the Pausing state.
+		// To prevent loss of state information, the state immediately before Pausing must currently be Running.
 		if box.Spec.Paused {
```
AiRanthem, line 357: The old controller may have already produced pause-in-progress objects with `Phase=Paused`, `Spec.Paused=true`, `Paused condition=false/nil`, and Pods still existing or being deleted. The new version treats `Phase=Paused` as fully paused and skips `EnsureSandboxPaused`, so old objects can no longer continue completing the pause.

Fundamentally, this issue comes from not handling compatibility between the previous pausing state (`phase = Paused`, `condition = false`) and the new pausing state properly. It will also affect the `IsSandboxResumable` judgment in `sandbox-manager`.

PR #493, opened 2026-06-03
fix(proxy): make route deletes resourceVersion-aware and stop the gateway/manager route tables from diverging

### LINKED ISSUE #492: [BUG] Gateway and manager disagree on a paused sandbox's route, and a deleted route can silently come back to life
# [BUG] Gateway and manager disagree on a paused sandbox's route, and a deleted route can silently come back to life

**What's going on**

Both the `sandbox-gateway` and the `sandbox-manager` keep an in-memory route table that maps a sandbox ID (`namespace--name`) to its pod IP and current state. Lifecycle events (claim / pause / resume / delete) are pushed between replicas over the memberlist `POST /refresh` endpoint, and on the gateway side a second writer — a CR-watch controller — also updates the same table from the Sandbox informer. The Envoy Go filter then reads that table and only routes traffic when the entry's `State == running`.

Reading through this path, three things stand out, and together they let the table drift away from reality.

**1. The two `/refresh` handlers do opposite things with the same message.**

The push payload is identical (a `proxy.Route` with a `state` field), but the two receivers disagree on what a non-running, non-dead state means:

```go
// pkg/proxy/server.go (manager) — deletes ONLY on "dead", keeps everything else
if route.State == v1alpha1.SandboxStateDead {
    s.DeleteRoute(route.ID)
} else {
    s.SetRoute(ctx, route)            // a {paused} push KEEPS the entry
}
```

```go
// pkg/sandbox-gateway/server/server.go — keeps ONLY "running", deletes everything else
if route.State == v1alpha1.SandboxStateRunning {
    registry.GetRegistry().Update(route.ID, route)
} else {
    registry.GetRegistry().Delete(route.ID)   // a {paused} push DROPS the entry
}
```

Pause pushes `{state: "paused"}` to every peer (`pkg/sandbox-manager/api.go`), so for the exact same event the manager's table ends up holding a `paused` entry while the gateway's table has nothing at all. The two data planes are supposed to be mirror images of each other and they aren't. Today the filter happens to reject both "paused" and "missing" with a 502, so it's latent — but the moment either side does anything more nuanced with a non-running route (wake-on-traffic resume, draining, surfacing state to a client), the two planes will behave differently for the same sandbox.

**2. Deleting a route throws away the resourceVersion ordering that updates carefully maintain.**

`Update` is written to be safe against out-of-order events — it compares `resourceVersion` and refuses to let an older event overwrite a newer one:

```go
// pkg/sandbox-gateway/registry/registry.go
func (r *Registry) Update(id string, route proxy.Route) bool {
    // LoadOrStore + CompareAndSwap, guarded by IsResourceVersionNewer(...)
}

func (r *Registry) Delete(id string) {
    r.entries.Delete(id)              // unconditional — no version check at all
}
```

`Delete` has no such guard. Once it empties an entry, the *next* write — no matter how stale — lands as a brand-new first write and wins unconditionally, because there's no longer an older value to compare against. So the careful ordering invariant only holds until the first delete.

That's a problem because the gateway has **two** writers racing on the same key with no ordering between them: the `/refresh` peer push and the CR-watch controller (both launched in `cmd/sandbox-gateway/main.go`). Picture a sandbox that gets deleted:

1. A `{dead}` push arrives → the entry is deleted.
2. The CR-watch controller reconciles from an informer cache that hasn't seen the delete yet → it reads the still-`running` Sandbox and calls `Update`. Because the entry was just deleted, this is a *first write*, so it's accepted unconditionally and the route is **resurrected as `running`**, pointing at the old pod IP.

A stale or reordered peer push can do the same thing. The entry self-heals once the informer catches up, but in the meantime the filter will happily route live client traffic to a pod IP that's been freed and possibly handed to a different sandbox. The whole point of the `resourceVersion` check is to prevent exactly this, and the delete path quietly opts out of it.

**3. (Smaller, same neighbourhood) the manager looks routes up by the wrong key when refreshing them.**

The manager stores routes under the full sandbox ID (`namespace--name`), but its periodic refresh loads them back by name only:

```go
// pkg/sandbox-manager/infra/sandboxcr/infra.go
oldRoute, exists := i.Proxy.LoadRoute(sbx.GetName())   // name only
newRoute := proxyutils.DefaultGetRouteFunc(sbx)        // stored under namespace--name
if !exists || newRoute.State != oldRoute.State || newRoute.IP != oldRoute.IP {
    i.Proxy.SetRoute(logs.NewContext(), newRoute)
}
```

For any non-empty namespace `exists` is always `false`, so the "skip if unchanged" shortcut never fires and the compare against `oldRoute` is meaningless. It's harmless today (it just always writes), but it's the same class of mistake — inconsistent keying of the same table — so it's worth fixing alongside.

**Why it actually matters**

The serious one is #2: routing real traffic to a dead/recycled pod IP, even briefly, is a correctness and isolation problem (a request meant for sandbox A can reach a pod now belonging to sandbox B). #1 is a latent consistency bug between two data planes that will turn into a real one as soon as non-running routes start being used for anything. Both get worse exactly under churn — high claim/pause/resume/delete rates and multiple replicas — which is when informer caches lag and peer pushes reorder.

**How to reproduce**

The divergence (#1) is deterministic:

1. Run a sandbox-manager and a sandbox-gateway in the same memberlist cluster.
2. Pause a running sandbox.
3. Inspect both route tables: the manager keeps a `paused` entry, the gateway has no entry. Same event, different state.

The resurrection (#2) is a race, but a realistic one:

1. Have a sandbox `running` in the gateway registry.
2. Delete it (a `{dead}` push removes the entry) while the gateway's Sandbox informer cache still shows it `running`.
3. Trigger a reconcile of that Sandbox before the cache catches up → the controller re-adds it as `running`.
4. The entry is back, pointing at the now-freed pod IP, and the filter routes to it until the cache converges.

**Suggested fix**

Make the route table the single arbiter and stop the two writers from being able to contradict each other:

- Give "delete" the same `resourceVersion` discipline as "update" — keep a short-lived tombstone carrying the resourceVersion at deletion time so a stale write can't resurrect a removed route as a fresh first write. The tombstone expires after a short TTL (by then the cluster has converged) and is swept inline, so there's no new background goroutine to leak.
- Pull the keep-vs-delete decision into one shared predicate (`delete iff state == dead`) and use it in *both* `/refresh` handlers, so an identical push produces an identical table on both planes. Storing a `paused` entry on the gateway is harmless — the filter already gates on `running` — and it lines the two planes back up.
- Apply the same versioned-delete model to the manager's route store too, since it's the same table and the same bug.
- Fix the manager's refresh lookup to use the full sandbox ID.

This stays inside the routing/registry code (`pkg/proxy`, `pkg/sandbox-gateway`) and the manager's route refresh; it doesn't touch the CRDs or the claim/scale paths.

This is a standalone, pre-existing correctness bug in the route-table consumer/registry code. It's not tied to any roadmap or feature-project deliverable and isn't driver- or workload-specific.

It's distinct from the already-filed route work: #413/#414 is about *how* `/refresh` pushes are sent (connection pooling, retry budget, fan-out on `SyncRouteWithPeers`) — this is about how the receivers *consume* those pushes (versioned deletes, the second CR-watch writer, and the divergent keep/delete semantics). #427/#428 is about authenticating the `/refresh` endpoint, a separate concern in the same handler. I checked the open and closed issues/PRs and found nothing covering the registry delete ordering, the dual-writer resurrection, or the keep/delete divergence. Happy to send a fix.

**Environment**

- Kubernetes version: N/A — code-level defect
- Affects: any multi-replica sandbox-manager + sandbox-gateway deployment; the resurrection window widens with claim/lifecycle churn and replica count

PRAteek-singHWY:
# fix(proxy): make route deletes resourceVersion-aware and stop the gateway/manager route tables from diverging

Fixes #492

## What this fixes

The in-memory route table that the gateway and manager keep in sync over `POST /refresh` has three related consistency problems (full write-up in #492):

1. **The two `/refresh` handlers interpret the same message differently.** For an identical `{state: "paused"}` push, the manager keeps the entry (`delete` only on `dead`) while the gateway drops it (`update` only on `running`). The two data planes end up holding different state for the same sandbox.
2. **`Delete` ignores `resourceVersion`.** `Update` carefully refuses to let an older event overwrite a newer one, but `Delete` empties the entry unconditionally — so the next write, however stale, lands as a fresh first write and wins. With two writers racing on the gateway (the `/refresh` push and the CR-watch controller), a delete followed by a lagging-cache `Update` **resurrects** a removed route, and the Envoy filter then routes live traffic to a freed/recycled pod IP until the informer catches up.
3. **The manager refreshes routes under the wrong key** — it stores under `namespace--name` but looks up by name only, so its "skip if unchanged" shortcut never fires.

## What changed

**A small shared route store that keeps ordering across deletes (`pkg/proxy/routestore`).**
The core fix is a concurrent, `resourceVersion`-aware route table that applies the *same* monotonic-version discipline to deletes that `Update` already applied to writes. A delete leaves a short-lived **tombstone** carrying the resourceVersion at deletion time; a later write is only accepted if it is genuinely newer, so a stale or reordered event can no longer resurrect a route that a newer event already removed. Tombstones expire after a short TTL (by which point the cluster has converged) and are swept inline on access and via an explicit `GC()` — there is **no new background goroutine** to leak. Both the gateway `registry` and the manager `proxy.Server` now use this store, so the bug is fixed in both data planes rather than just one.

**One shared keep-vs-delete predicate (`proxyutils.ShouldDeleteRoute`).**
Both `/refresh` handlers now decide identically: delete iff the state is `dead`, otherwise store the route with its state. Storing a `paused` entry on the gateway is harmless — the filter already gates on `running` — and it realigns the two planes for the same event.

**Versioned deletes everywhere they're called.**
`registry.Delete` and `proxy.Server.DeleteRoute` now take the deleting event's `resourceVersion` (empty falls back to the resourceVersion of the entry being removed, which is what the CR-watch controller's not-found path has). All call sites — the two `/refresh` handlers, the gateway controller's delete branches, the manager's reconcile/delete paths — pass it through.

**The keying fix.**
The manager's `refreshRoute` now looks routes up by the full sandbox ID, matching how they're stored.

**A short design note** in `docs/proposals/` documents the route-table ownership model (registry as arbiter; both writers idempotent and commutative under resourceVersion ordering), so the invariant is written down rather than implicit.

## Testing

- New table-driven tests for `routestore` covering: first write, newer/older/equal-version writes, delete-then-stale-write (the resurrection case), tombstone expiry, and concurrent writers.
- Updated/added tests for both `/refresh` handlers (shared predicate, paused no longer dropped on the gateway), the gateway controller's delete branches, and the registry.
- `go build ./...` and `go test` on the changed packages pass; `gofmt`/`goimports` clean.

## Scope

Self-contained to the routing/registry code (`pkg/proxy`, `pkg/sandbox-gateway`) plus the manager's route refresh. No CRD, claim, or scale changes. Distinct from #413/#414 (which is about *sending* `/refresh` pushes) and #427/#428 (authenticating the endpoint) — this is about how the receivers *consume* those pushes.

### DIFF
--- a/docs/proposals/20260603-route-table-consistency.md
+++ b/docs/proposals/20260603-route-table-consistency.md
@@ -1,0 +1,115 @@
+---
+title: Route Table Consistency for sandbox-manager and sandbox-gateway
+authors:
+- "@PRAteek-singHWY"
+creation-date: 2026-06-03
+last-updated: 2026-06-03
+status: implementable
+---
+
+# Route Table Consistency for sandbox-manager and sandbox-gateway
+
+| Metadata    | Details            |
+|-------------|--------------------|
+| **Author**  | @PRAteek-singHWY   |
+| **Status**  | Implementable      |
+| **Created** | 2026-06-03         |
+
+## Summary
+
+Both the sandbox-manager (`pkg/proxy`) and the sandbox-gateway
+(`pkg/sandbox-gateway/registry`) keep an in-memory route table mapping a sandbox
+ID (`namespace--name`) to its pod IP and state. Entries reach a replica through
+two unordered channels: peer `POST /refresh` pushes (memberlist) and, on the
+gateway, a Sandbox CR-watch controller reconciling from an informer cache. This
+note documents the consistency model those tables must uphold and the design
+that enforces it.
+
+## Motivation
+
+Three defects existed in the consumer side of route synchronization:
+
+1. **Divergent delete semantics.** For an identical `{state: "paused"}` push the
+   manager handler kept the entry (delete only on `dead`) while the gateway
+   handler dropped it (keep only `running`), so the two data planes held
+   different state for the same event.
+2. **Version-blind deletes.** `Update` refused to let an older `resourceVersion`
+   overwrite a newer one, but `Delete` erased the entry — and the recorded
+   version — unconditionally. The next write, however stale, then landed as a
+   fresh first write. With two unordered writers, a delete followed by a
+   lagging-cache update could *resurrect* a removed route, and the Envoy filter
+   would route live traffic to a freed pod IP until the informer converged.
+3. **Inconsistent keying.** The manager's periodic refresh looked routes up by
+   name while storing them under `namespace--name`, making its "skip if
+   unchanged" comparison meaningless.
+
+## Design
+
+### Invariant
+
+A route table entry is only mutated by an event whose `resourceVersion` is at
+least as new as the version currently recorded for that slot — **including
+deletes**. Writers never need to coordinate with each other; the store is the
+single arbiter, and writes from any source are idempotent and commutative under
+resourceVersion ordering.
+
+### Shared store (`pkg/proxy/routestore`)
+
+Both tables delegate to one implementation:
+
+- `Set(id, route)` keeps the existing CAS + `IsResourceVersionNewer` (`>=`)
+  semantics for live entries.
+- `Delete(id, resourceVersion)` replaces the entry with a short-lived
+  **tombstone** stamped with the larger of the deleting event's resourceVersion
+  and the one recorded on the entry (an empty value — e.g. an informer
+  not-found — falls back to the recorded one).
+- A write may replace an active tombstone only if it is **strictly newer**
+  (`IsResourceVersionReallyNewer`, `>`). An equal version is the very write the
+  deletion superseded; accepting it would resurrect the route. A genuine
+  recreate always carries a strictly higher version (etcd resourceVersions are
+  monotonic), so it passes.
+- Tombstones expire after `DefaultTombstoneTTL` (10 minutes — by then the
+  level-triggered reconcilers have converged) and are reclaimed lazily on
+  access, by an inline sweep once a threshold accumulates, and by an explicit
+  `GC()` hooked into the manager's periodic route reconciliation. No new
+  background goroutine is introduced.
+
+### One keep-vs-delete rule
+
+`proxyutils.ShouldDeleteRoute(state)` (`delete iff state == dead`) is the single
+predicate used by both `/refresh` handlers. Every non-dead state — creating,
+available, paused, running — is stored with its state; consumers decide
+routability from `Route.State` (the gateway filter already serves only
+`running`). An identical push therefore produces an identical table on both
+data planes.
+
+### Writers
+
+- Manager `/refresh`, gateway `/refresh`: shared predicate + versioned delete.
+- Gateway CR-watch controller: updates with the object's resourceVersion;
+  deletes pass the observed resourceVersion (deletion-timestamped object) or
+  empty (not-found), falling back to the recorded one.
+- Manager periodic reconciler: orphan deletes pass the route's own
+  resourceVersion and trigger tombstone GC; the refresh path now keys lookups
+  by the full sandbox ID.
+
+## Alternatives
+
+- **Single writer (controller-only or push-only).** Rejected: the push path is
+  the low-latency channel for lifecycle changes, the watch path is the
+  level-triggered backstop; both are wanted. Making the store order-safe keeps
+  both without coordination.
+- **Reference counting / per-id locks.** Heavier than needed; the
+  resourceVersion already provides a total order per object.
+- **Unbounded tombstones.** Correct but leaks memory under churn; TTL + lazy
+  sweep bounds it without a goroutine.
+
+## Test Plan
+
+- Table-driven unit tests on the store: version arbitration, delete-then-stale
+  -write (resurrection guard), equal-version-vs-tombstone, empty-version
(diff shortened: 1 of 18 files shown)


PRAteek-singHWY:
@furykerry @zmberg @AiRanthem PTAL when you get a chance, this fixes #492: both `/refresh` handlers now share a single keep-vs-delete rule (delete only on `dead`), and route deletes are resourceVersion-aware via a short-lived tombstone in a small shared store (`pkg/proxy/routestore`), so a stale peer push or lagging-cache reconcile can no longer resurrect a deleted route and send traffic to a freed pod IP.

It also fixes the manager's `refreshRoute` looking routes up by name instead of the full `namespace--name` key. The consistency model is written down in `docs/proposals/20260603-route-table-consistency.md`, and the resurrection guard is covered by table-driven tests at the store, handler, and controller level.

PR #510, opened 2026-06-05
feat(e2b): raise create timeout floor to 300s with skip override

AiRanthem:
## Summary
- Raise the minimum create-sandbox timeout from 30s to 300s (`DefaultMinTimeoutSeconds` → `DefaultMinCreateTimeoutSeconds`).
- Add the `e2b.agents.kruise.io/skip-create-timeout-min` extension so callers can opt out of the min floor; the max cap still applies.
- Update Go unit tests and the E2B Python tests to the new floor.

## Test Plan
- [x] `go test ./pkg/servers/e2b/...` passes
- [ ] E2B Python integration tests run against a live cluster


AiRanthem:
This issue will be fixed in another PR.

PR #511, opened 2026-06-06
fix(updateops): skip sandboxes whose template already matches patch target

zmberg:
## Summary
- Sandboxes created during an ops run that already have the latest template were incorrectly counted as `updatingReplicas`, causing the ops to never complete
- Add `sandboxNoNeedUpdate` classification in `classifySandbox` to skip these sandboxes from both patching and counting
- Covers two scenarios: candidates whose template already matches the patch, and sandboxes already labeled by ops but never needing upgrade (stuck updating)

## Test plan
- [x] Unit tests for `isSandboxTemplateMatchPatch` (6 cases: match, differ, nil template, empty patch, partial match, partial differ)
- [x] Unit tests for `classifySandbox` with `sandboxNoNeedUpdate` (9 cases covering both entry paths)
- [x] Integration tests: `TestReconcile_SkipsSandboxWhoseTemplateAlreadyMatchesPatch`, `TestReconcile_UpdatingNotSkippedWhenGenerationMismatch`, `TestReconcile_UpdatingNotSkippedWhenConditionExists`
- [x] E2E test: `should skip sandboxes whose template already matches the patch target`

### DIFF
--- a/pkg/controller/sandboxupdateops/patch.go
+++ b/pkg/controller/sandboxupdateops/patch.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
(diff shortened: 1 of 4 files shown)


pkg/controller/sandboxupdateops/patch.go
```diff
@@ -29,6 +30,32 @@ import (
...
+	}
+	// Re-marshal the original through the same path to normalize field ordering
+	var original, merged interface{}
+	if err := json.Unmarshal(originalBytes, &original); err != nil {
+		return false
+	}
+	if err := json.Unmarshal(mergedBytes, &merged); err != nil {
+		return false
+	}
+	return reflect.DeepEqual(original, merged)
```
furykerry: is original necessary, just use sbx.Spec.Template

pkg/controller/sandboxupdateops/patch.go
```diff
@@ -29,6 +30,32 @@ import (
...
+	if sbx.Spec.Template == nil || len(ops.Spec.Patch.Raw) == 0 {
+		return false
+	}
+	originalBytes, err := json.Marshal(sbx.Spec.Template)
+	if err != nil {
+		return false
+	}
+	mergedBytes, err := strategicpatch.StrategicMergePatch(originalBytes, ops.Spec.Patch.Raw, &v1.PodTemplateSpec{})
+	if err != nil {
+		return false
```
furykerry, line 47: consider add a log here for diagnose potential patch problems

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -284,13 +287,19 @@ func (s sandboxUpdateState) String() string {
...
 	default:
 		return "Unknown"
 	}
 }
 
-func classifySandbox(sbx *agentsv1alpha1.Sandbox, opsName string) sandboxUpdateState {
-	if sbx.Labels[agentsv1alpha1.LabelSandboxUpdateOps] != opsName {
+func classifySandbox(sbx *agentsv1alpha1.Sandbox, ops *agentsv1alpha1.SandboxUpdateOps) sandboxUpdateState {
+	if sbx.Labels[agentsv1alpha1.LabelSandboxUpdateOps] != ops.Name {
+		if isSandboxTemplateMatchPatch(sbx, ops) && sbx.Status.Phase == agentsv1alpha1.SandboxRunning &&
```
furykerry: if the sandbox status is not running yet, we shall further check the phase of sanboxupdateops, if it is not completed or failed, we shall not add the sandbox as candidate yet.

test/e2e/sandboxupdateops_test.go
```diff
@@ -791,5 +791,65 @@ var _ = Describe("SandboxUpdateOps E2E", func() {
...
+			}
+			for i := 0; i < 2; i++ {
+				waitSandboxRunning(sandboxes[i])
+				klog.InfoS("Sandbox is Running", "name", sandboxes[i].Name, "index", i)
+			}
+
+			By("Creating SandboxUpdateOps with patch target matching current template")
+			ops := &agentsv1alpha1.SandboxUpdateOps{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("ops-skip-%s", labelValue[:10]),
```
furykerry: consider change const 10 to more meaningful expression, e.g. len("batch-skip"), so that future changes will not accidentally break the test cases.

PR #513, opened 2026-06-07
fix(sidecarutils): stabilize init container injection order for backward compatibility

BH4AWS:
…ard compatibility

The previous loop-based injection iterated sandbox.Spec.Runtimes in declaration order, which could produce different InitContainers ordering when the spec listed csi-mount before agent-runtime. This breaks compatibility with pre-existing pods that were injected by the older two-phase logic (agent-runtime first, then csi-mount).

Refactor doSidecarInjection into two phases:
- Phase 1: inject agent-runtime and csi-mount in a deterministic fixed order (matching historical behavior), regardless of spec declaration.
- Phase 2: inject remaining runtimes (e.g. traffic-proxy) in their declaration order via the generic conflict-check + apply path.

Also add TestDoSidecarInjection_FixedOrder with table-driven cases covering normal order, reversed order, and interleaved traffic-proxy to verify InitContainers stability.

### DIFF
--- a/pkg/utils/sidecarutils/utils.go
+++ b/pkg/utils/sidecarutils/utils.go
@@ -49,23 +49,40 @@ func InjectSandboxRuntimes(ctx context.Context, sandbox *agentsv1alpha1.Sandbox,
 	return doSidecarInjection(ctx, sandbox, pod, config)
 }
 
+// fixedOrderRuntimes defines the injection order for built-in runtimes.
+// Preserve historical injection order: agent-runtime first, then csi-mount,
+// to keep init container ordering stable for pre-existing pods regardless of
+// the declaration order in sandbox.Spec.Runtimes.
+var fixedOrderRuntimes = []string{
+	agentsv1alpha1.RuntimeConfigForInjectAgentRuntime,
+	agentsv1alpha1.RuntimeConfigForInjectCsiMount,
+}
+
 func doSidecarInjection(ctx context.Context, sandbox *agentsv1alpha1.Sandbox, pod *corev1.Pod, injectConfigMap map[string]string) error {
 	logger := logf.FromContext(ctx)
 
-	injectedRuntimes := sets.NewString()
+	// Build a quick-lookup set of declared runtimes for the fixed-order phase.
+	declared := sets.NewString()
 	for _, r := range sandbox.Spec.Runtimes {
-		runtime := r.Name
-		if injectedRuntimes.Has(runtime) {
-			logger.V(5).Info("skipping duplicate runtime, already processed", "runtime", runtime)
+		declared.Insert(r.Name)
+	}
+
+	injectedRuntimes := sets.NewString()
+
+	// Phase 1: inject built-in runtimes in a deterministic fixed order so that
+	// the resulting InitContainers sequence is stable across spec permutations.
+	for _, runtime := range fixedOrderRuntimes {
+		if !declared.Has(runtime) {
 			continue
 		}
-		injectedRuntimes.Insert(runtime)
 
 		runtimeInjectConfig, err := parseInjectConfig(ctx, runtime, injectConfigMap)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/utils/sidecarutils/utils.go
```diff
@@ -75,16 +91,30 @@ func doSidecarInjection(ctx context.Context, sandbox *agentsv1alpha1.Sandbox, po
...
 	}
 
+	// Phase 2: inject remaining runtimes in their declaration order. New runtime
+	// types (e.g. traffic-proxy) fall through to the generic handler.
+	for _, r := range sandbox.Spec.Runtimes {
+		runtime := r.Name
+		if injectedRuntimes.Has(runtime) {
+			continue
+		}
+		injectedRuntimes.Insert(runtime)
```
furykerry: consider move Insert to the position just before applyInjectionTemplate

pkg/utils/sidecarutils/utils.go
```diff
@@ -49,14 +49,30 @@ func InjectSandboxRuntimes(ctx context.Context, sandbox *agentsv1alpha1.Sandbox,
...
-			logger.V(5).Info("skipping duplicate runtime, already processed", "runtime", runtime)
+		declared.Insert(r.Name)
+	}
+
+	injectedRuntimes := sets.NewString()
+
+	// Phase 1: inject built-in runtimes in a deterministic fixed order so that
+	// the resulting InitContainers sequence is stable across spec permutations.
+	for _, runtime := range fixedOrderRuntimes {
+		if !declared.Has(runtime) {
```
furykerry, line 75: consider move L78 Insert to the position just before switch runtime

pkg/utils/sidecarutils/utils.go
```diff
@@ -75,16 +91,30 @@ func doSidecarInjection(ctx context.Context, sandbox *agentsv1alpha1.Sandbox, po
 			if !isContainersExists(pod.Spec.InitContainers, runtimeInjectConfig.Sidecars) && !isContainersExists(pod.Spec.Containers, runtimeInjectConfig.Sidecars) {
 				setCSIMountContainer(ctx, &pod.Spec, runtimeInjectConfig)
 			}
```
furykerry, line 94: consider add a log similar to traffic-proxy, e.g. `logger.V(5).Info("injecting runtime", "name", runtime)
`

pkg/utils/sidecarutils/utils_test.go
```diff
@@ -1659,6 +1659,89 @@ func TestDoSidecarInjectionDuplicateRuntimes(t *testing.T) {
...
+		"csiSidecar": [{"name": "csi-sidecar", "image": "csi:v1"}],
+		"volume": [{"name": "csi-vol", "emptyDir": {}}]
+	}`
+
+	tests := []struct {
+		name     string
+		runtimes []agentsv1alpha1.RuntimeConfig
+	}{
+		{
+			name: "normal order: agent-runtime then csi-mount",
```
furykerry, line 1689: The current test cases all declare both agent-runtime and csi-mount. Adding a case where only one of them appears (e.g., only agent-runtime without csi-mount) would verify that Phase 1 correctly skips the absent runtime and the InitContainers list is not malformed.

pkg/utils/sidecarutils/utils.go
BH4AWS: done

pkg/utils/sidecarutils/utils.go
BH4AWS: done

pkg/utils/sidecarutils/utils.go
BH4AWS: done

pkg/utils/sidecarutils/utils_test.go
BH4AWS: done

PR #514, opened 2026-06-08
fix(sandboxset): add legacy revision hash compat to prevent sandbox recreation on upgrade

zmberg:

After the revision hash algorithm was changed from release-v0.3 (ControllerRevision-based FNV-32 over `spec.template` with `$patch:replace` wrapper) to the current approach (`json.Marshal` of `SandboxTemplateSpec` including VolumeClaimTemplates, PersistentContents, Runtimes), existing sandboxes would be incorrectly flagged as outdated and mass-recreated during a controller upgrade.

This PR adds backward-compatible legacy hash computation so that existing sandboxes are preserved:

- `computeLegacyRevisionHash` replicates the v0.3 algorithm for inline templates
- `isRevisionUpdated` helper checks both new and legacy hashes during update-group classification
- `buildUpdateGroups` and `scaleDown` accept `legacyRevision` for dual comparison
- `templateRef` path skips legacy compat (templateRef was introduced after the algorithm change, no legacy sandboxes exist)

**Key invariant:** When the SandboxSet template is actually modified, the legacy hash also changes — so rolling updates are still correctly triggered for real config changes.

NONE

1. Run unit tests:
   ```bash
   go test ./pkg/controller/sandboxset/ -count=1 -timeout 60s
   ```
2. Key test cases to review:
   - `TestLegacyHashChangesWhenTemplateModified` — verifies real config changes still trigger rolling update
   - `TestTemplateModificationTriggersRollingUpdateEndToEnd` — end-to-end scenario with `buildUpdateGroups`
   - `TestBuildUpdateGroupsWithLegacyRevision` — verifies old-hash sandboxes are compat'd as Updated

- The legacy hash function only handles inline templates. SandboxSets using `templateRef` skip the legacy path entirely (return empty string), since `templateRef` was introduced after the hash algorithm change.
- `computeLegacyRevisionHash` errors are intentionally non-fatal — if it fails, we simply skip the fallback comparison and only use the new hash.
- The `initResult` struct bundles `status` + `legacyRevision` to avoid adding extra return values to `initNewStatus`.

### DIFF
--- a/pkg/controller/sandboxset/revision.go
+++ b/pkg/controller/sandboxset/revision.go
@@ -21,7 +21,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"hash/fnv"
+	"strconv"
 
+	apps "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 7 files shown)


pkg/controller/sandboxset/revision.go
```diff
@@ -75,3 +75,46 @@ func computeRevisionHash(spec *agentsv1alpha1.SandboxTemplateSpec) (string, erro
...
+// VolumeClaimTemplates, PersistentContents, or Runtimes.
+func (r *Reconciler) computeLegacyRevisionHash(_ context.Context, sbs *agentsv1alpha1.SandboxSet) (string, error) {
+	// Legacy SandboxSets only use inline templates; templateRef was introduced
+	// after the hash algorithm change, so no backward compat is needed for it.
+	if sbs.Spec.TemplateRef != nil || sbs.Spec.Template == nil {
+		return "", nil
+	}
+	template := sbs.Spec.Template
+
+	templateRaw, err := json.Marshal(template)
```
furykerry: the old logic use`runtime.Encode` to marshal objects into byte arrays, consider keep the old logic

pkg/controller/sandboxset/revision_test.go
```diff
@@ -327,3 +327,397 @@ func TestComputeRevisionHash_InlineAndTemplateRefConsistency(t *testing.T) {
 	assert.Equal(t, hashInline, hashRef,
 		"inline and templateRef describing the same content must produce identical hash")
 }
+
+// TestComputeLegacyRevisionHash verifies the backward-compatible legacy hash
+// computation that replicates the release-v0.3 algorithm.
+func TestComputeLegacyRevisionHash(t *testing.T) {
```
furykerry, line 337: plz add a test case that uses a SandboxSet fixture with a hash value computed by running the actual v0.3 controller code:

pkg/controller/sandboxset/utils.go
```diff
@@ -58,7 +64,16 @@ func (r *Reconciler) initNewStatus(ctx context.Context, ss *agentsv1alpha1.Sandb
 	newStatus.UpdateRevision = hash
 	newStatus.ObservedGeneration = ss.Generation
 	newStatus.CurrentRevision = name
-	return newStatus, nil
+
+	// Compute legacy hash for backward compatibility with sandboxes created
+	// before the hash algorithm was changed. Errors are non-fatal: if legacy
+	// hash cannot be computed, we simply skip the fallback comparison.
+	legacyHash, _ := r.computeLegacyRevisionHash(ctx, ss)
```
furykerry: consider log the error at debug level

pkg/controller/sandboxset/revision.go
```diff
@@ -75,3 +75,46 @@ func computeRevisionHash(spec *agentsv1alpha1.SandboxTemplateSpec) (string, erro
...
+//
+// Legacy algorithm:
+//  1. JSON-marshal the PodTemplateSpec (inline or from referenced SandboxTemplate)
+//  2. Unmarshal to map[string]interface{} and inject "$patch": "replace"
+//  3. Wrap as {"spec": {"template": <map>}}
+//  4. JSON-marshal the wrapper and FNV-32 hash the bytes
+//
+// The legacy hash only considers spec.template; it does NOT include
+// VolumeClaimTemplates, PersistentContents, or Runtimes.
+func (r *Reconciler) computeLegacyRevisionHash(_ context.Context, sbs *agentsv1alpha1.SandboxSet) (string, error) {
```
furykerry, line 96: context.Context is not used, consider remove the param variable

pkg/controller/sandboxset/revision.go
```diff
@@ -75,3 +82,112 @@ func computeRevisionHash(spec *agentsv1alpha1.SandboxTemplateSpec) (string, erro
...
+
+// HashControllerRevision hashes the contents of revision's Data using FNV hashing. If probe is not nil, the byte value
+// of probe is added written to the hash as well. The returned hash will be a safe encoded string to avoid bad words.
+func HashControllerRevision(revision *apps.ControllerRevision, probe *int32) string {
+	hf := fnv.New32()
+	if len(revision.Data.Raw) > 0 {
+		hf.Write(revision.Data.Raw)
+	}
+	if revision.Data.Object != nil {
+		DeepHashObject(hf, revision.Data.Object)
```
furykerry: DeepHashObject is only called from HashControllerRevision, which itself is only called from NewControllerRevision. But NewControllerRevision only writes Data.Raw (never Data.Object), so the revision.Data.Object != nil branch in HashControllerRevision is unreachable. More importantly, DeepHashObject uses dump.ForHash which internally uses go-spew, but go-spew is only an indirect dependency in go.mod. If it's not actually being exercised in production, this is dead code that adds maintenance burden.

Consider removing DeepHashObject entirely and simplifying HashControllerRevision to only handle Data.Raw,

pkg/controller/sandboxset/utils.go
```diff
@@ -49,7 +50,13 @@ type GroupedSandboxes struct {
 	Dead      []*agentsv1alpha1.Sandbox // Sandboxes should be deleted
 }
 
-func (r *Reconciler) initNewStatus(ctx context.Context, ss *agentsv1alpha1.SandboxSet) (*agentsv1alpha1.SandboxSetStatus, error) {
+// initResult bundles the output of initNewStatus to avoid excessive return values.
+type initResult struct {
```
furykerry: consider rename initResult to something more specific like initStatusResult

pkg/controller/sandboxset/utils.go
```diff
@@ -58,7 +65,19 @@ func (r *Reconciler) initNewStatus(ctx context.Context, ss *agentsv1alpha1.Sandb
...
 	newStatus.ObservedGeneration = ss.Generation
 	newStatus.CurrentRevision = name
-	return newStatus, nil
+
+	// Compute legacy hash for backward compatibility with sandboxes created
+	// before the hash algorithm was changed. Errors are non-fatal: if legacy
+	// hash cannot be computed, we simply skip the fallback comparison.
+	legacyHash, err := r.computeLegacyRevisionHash(ctx, ss)
+	if err != nil {
+		klog.Warningf("failed to compute legacy revision hash, error: %s", err.Error())
```
furykerry: using klog.ErrorS?

pkg/controller/sandboxset/revision.go
```diff
@@ -75,3 +82,112 @@ func computeRevisionHash(spec *agentsv1alpha1.SandboxTemplateSpec) (string, erro
...
+	if err = json.Unmarshal(str, &raw); err != nil {
+		return "", err
+	}
+	if spec, ok := raw["spec"].(map[string]interface{}); ok {
+		if template, ok := spec["template"].(map[string]interface{}); ok {
+			template["$patch"] = "replace"
+			specCopy["template"] = template
+		}
+	}
+	patch, _ := json.Marshal(map[string]interface{}{"spec": specCopy})
```
furykerry: plz do not ignore error, and return error

PR #516, opened 2026-06-08
update run_envd.sh script

BH4AWS:

### DIFF
--- a/pkg/agent-runtime/envd-run.sh
+++ b/pkg/agent-runtime/envd-run.sh
@@ -1,6 +1,30 @@
 #!/bin/bash
 set -e  # Enable exit on error for the entire script
 
+# Parse arguments. Supported flags:
+#   --nolog   Suppress envd stdout/stderr (do not redirect to /proc/1/fd/1).
+#             Default behaviour pipes envd output to PID 1's stdout so that
+#             container log collectors pick it up.
+#   --help    Show usage information and exit.
+NOLOG=false
+for arg in "$@"; do
+    case "$arg" in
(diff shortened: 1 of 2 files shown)


pkg/agent-runtime/envd-run.sh
```diff
@@ -1,6 +1,22 @@
...
 set -e  # Enable exit on error for the entire script
 
+# Parse arguments. Supported flags:
+#   nolog  Suppress envd stdout/stderr (do not redirect to /proc/1/fd/1).
+#          Default behaviour pipes envd output to PID 1's stdout so that
+#          container log collectors pick it up.
+NOLOG=false
+for arg in "$@"; do
+    case "$arg" in
+        nolog)
```
furykerry: Issue: The nolog argument is parsed as a bare positional token without any dash prefix. This violates the POSIX/GNU command-line argument convention where options use --long or -s prefixes, and positional arguments are data operands.

consider：
```
NOLOG=false
for arg in "$@"; do
    case "$arg" in
        --nolog)
            NOLOG=true
            ;;
        --help|-h)
            echo "Usage: envd-run.sh [OPTIONS]"
            echo "  --nolog    Suppress envd stdout/stderr (redirect to /dev/null)"
            echo "  --help     Show this help message"
            exit 0
            ;;
        *)
            echo "Error: unknown argument '$arg'" >&2
            exit 1
            ;;
    esac
done
```

BH4AWS: fixed

PR #518, opened 2026-06-09
test(ci): expand sandbox-manager e2e coverage

AiRanthem:
## Summary
- add a latest E2B gateway CI matrix entry with E2B auth disabled
- run MySQL key storage E2E only through the gateway and cover 3 sandbox-manager replicas
- add an auth-disabled test runner mode that skips API-key-only checks

## Test Plan
- [x] bash -n hack/run-e2b-e2e-test.sh
- [x] ruby YAML parse for updated workflow files
- [x] kustomize build config/sandbox-manager
- [x] simulated kustomize builds for auth-disabled and MySQL patches

### DIFF
--- a/.github/workflows/e2e-e2b-latest.yaml
+++ b/.github/workflows/e2e-e2b-latest.yaml
@@ -131,6 +131,7 @@ jobs:
             {
               echo ""
               echo "# Disable E2B auth (appended by e2e-e2b-latest auth-disabled matrix)"
+              echo "# Positional path targets --e2b-enable-auth in config/sandbox-manager/deployment.yaml."
               echo "- op: replace"
               echo "  path: /spec/template/spec/containers/0/args/8"
               echo "  value: --e2b-enable-auth=false"
--- a/.github/workflows/e2e-e2b-latest.yaml
+++ b/.github/workflows/e2e-e2b-latest.yaml
@@ -34,8 +34,13 @@ jobs:
         include:
           - name: without sandbox-gateway
             with_gateway: false
+            e2b_enable_auth: true
           - name: with sandbox-gateway
             with_gateway: true
+            e2b_enable_auth: true
+          - name: with sandbox-gateway, auth disabled
+            with_gateway: true
+            e2b_enable_auth: false
     name: sandbox (${{ matrix.name }})
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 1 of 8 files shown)


pkg/servers/e2b/keys/mysql.go
```diff
@@ -191,17 +191,39 @@ func (k *mysqlKeyStorage) ensureAdminKey(ctx context.Context, adminTeamID uint)
...
+	var existing TeamAPIKeyEntity
+	if err := k.db.WithContext(ctx).Unscoped().Where(&TeamAPIKeyEntity{UID: key.UID}).
+		First(&existing).Error; err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return fmt.Errorf("create admin key: duplicate conflict but admin uid %s not present", key.UID)
+		}
+		return fmt.Errorf("lookup admin key after duplicate insert: %w", err)
+	}
+
+	return k.updateAdminKey(ctx, key)
```
furykerry, line 219: check whether existing key equals key ?

.github/workflows/e2e-e2b-mysql-latest.yaml
```diff
@@ -128,10 +140,40 @@ jobs:
...
+            echo "  path: /spec/replicas"
+            echo "  value: 3"
+            echo "- op: replace"
+            echo "  path: /spec/template/spec/containers/0/resources/limits/cpu"
+            echo '  value: "500m"'
+            echo "- op: replace"
+            echo "  path: /spec/template/spec/containers/0/resources/limits/memory"
+            echo '  value: "1Gi"'
+            echo "- op: replace"
+            echo "  path: /spec/template/spec/containers/0/resources/requests/cpu"
```
furykerry: consider change the deploymet.yaml of sandbox-manager directly, we can reduce request and keep limits, so that the resource requirement can remain the same even in three replicas case

.github/workflows/e2e-e2b-latest.yaml
```diff
@@ -122,6 +127,17 @@ jobs:
 
       - name: Install Kruise Agents
         run: |
+          if [ "${{ matrix.e2b_enable_auth }}" = "false" ]; then
+            {
+              echo ""
+              echo "# Disable E2B auth (appended by e2e-e2b-latest auth-disabled matrix)"
+              echo "- op: replace"
+              echo "  path: /spec/template/spec/containers/0/args/8"
```
furykerry, line 136: he appended patch uses path: /spec/template/spec/containers/0/args/8 — a hard-coded positional index. If the base deployment manifest changes arg ordering (e.g. a new flag is inserted before index 8), this silently patches the wrong argument.

Consider add a comment noting the positional dependency and pinning it to a specific flag name:

.github/workflows/e2e-e2b-mysql-latest.yaml
AiRanthem: DONE

.github/workflows/e2e-e2b-latest.yaml
AiRanthem: DONE

pkg/servers/e2b/keys/mysql.go
AiRanthem: ADDED

Issue #515, opened 2026-06-08
[feature request] Support memory resize in sandboxClaim process


PersistentJZH:
**What would you like to be added**:
Currently, sandbox claims only support cpu inplace update resize, but memory resize also requires.

**Why is this needed**:

Issue #2, opened 2025-12-09
[feature request] support scale sub-resource for sandboxset


furykerry:
add scale sub-resource for sandboxset , one can refers to : https://book.kubebuilder.io/reference/generating-crd.html#subresources

**What would you like to be added**:
1. add label selector in sandboxset status
2. add kubebuilder mark in the struct definition

**Why is this needed**:
1.ease sandboxset scaling operation
2. support future auto scaler integration.

ZhaoQing7892:
I can work on this！

furykerry:
fixed by https://github.com/openkruise/agents/pull/6

PR #519, opened 2026-06-09
feat: support memory resize for sandbox claims

PersistentJZH:

SandboxClaim resource resize previously only supported CPU. Kubernetes in-place pod resize supports both CPU and memory, with behavior controlled by the pod container `resizePolicy`.

Memory resize should be supported through both the CRD API and the E2B-compatible API path, **while preserving the resize policy already defined on the underlying SandboxSet/Sandbox pod template.**

- Allow SandboxClaim inplaceUpdate.resources and E2B extensions to resize memory in addition to CPU.

- Add unit and e2e coverage for CPU and memory resize behavior, including RestartContainer and NotRequired resizePolicy cases.

https://github.com/openkruise/agents/issues/515

### DIFF
--- a/pkg/controller/sandboxclaim/core/common_control.go
+++ b/pkg/controller/sandboxclaim/core/common_control.go
@@ -44,6 +44,7 @@ import (
 	annotationutils "github.com/openkruise/agents/pkg/utils/annotations"
 	"github.com/openkruise/agents/pkg/utils/csiutils"
 	utilfeature "github.com/openkruise/agents/pkg/utils/feature"
+	"github.com/openkruise/agents/pkg/utils/inplaceupdate"
 	runtimeclient "github.com/openkruise/agents/pkg/utils/runtime"
 	"github.com/openkruise/agents/pkg/utils/timeout"
 )
@@ -140,6 +141,32 @@ func (c *commonControl) EnsureClaimClaiming(ctx context.Context, args ClaimArgs)
 				c.recorder.Event(claim, "Warning", "FeatureGateDisabled", msg)
 				TransitionToCompleted(args.NewStatus, "FeatureGateDisabled", msg)
 				return NoRequeue(), nil
+			}
+		}
+	}
+
+	// Step 7.1: Validate the inplace update request up front so invalid input
+	// fails fast with an accurate reason instead of spinning until
+	// ClaimTimeout with a misleading NoAvailableSandboxes event.
+	if claim.Spec.InplaceUpdate != nil && claim.Spec.InplaceUpdate.Resources != nil {
+		res := claim.Spec.InplaceUpdate.Resources
+		if err := sandboxcr.ValidateResizeResources(res.Requests, res.Limits); err != nil {
+			msg := err.Error()
+			log.Info(msg)
+			c.recorder.Event(claim, "Warning", "InvalidInplaceUpdateResources", msg)
+			TransitionToCompleted(args.NewStatus, "InvalidInplaceUpdateResources", msg)
+			return NoRequeue(), nil
+		}
+		if sandboxSet.Spec.Template != nil {
+			for i := range sandboxSet.Spec.Template.Spec.Containers {
+				templateContainer := &sandboxSet.Spec.Template.Spec.Containers[i]
+				if err := inplaceupdate.CheckContainerMemoryDownscale(templateContainer, res.Requests, res.Limits); err != nil {
+					msg := fmt.Sprintf("container %q: %v", templateContainer.Name, err)
+					log.Info(msg)
+					c.recorder.Event(claim, "Warning", "MemoryDownscaleNotSupported", msg)
+					TransitionToCompleted(args.NewStatus, "MemoryDownscaleNotSupported", msg)
+					return NoRequeue(), nil
+				}
 			}
 		}
 	}
--- a/pkg/controller/sandboxclaim/core/common_control.go
+++ b/pkg/controller/sandboxclaim/core/common_control.go
@@ -44,7 +44,6 @@ import (
 	annotationutils "github.com/openkruise/agents/pkg/utils/annotations"
 	"github.com/openkruise/agents/pkg/utils/csiutils"
 	utilfeature "github.com/openkruise/agents/pkg/utils/feature"
-	"github.com/openkruise/agents/pkg/utils/inplaceupdate"
 	runtimeclient "github.com/openkruise/agents/pkg/utils/runtime"
 	"github.com/openkruise/agents/pkg/utils/timeout"
 )
@@ -147,7 +146,10 @@ func (c *commonControl) EnsureClaimClaiming(ctx context.Context, args ClaimArgs)
 
 	// Step 7.1: Validate the inplace update request up front so invalid input
 	// fails fast with an accurate reason instead of spinning until
-	// ClaimTimeout with a misleading NoAvailableSandboxes event.
+	// ClaimTimeout with a misleading NoAvailableSandboxes event. Only
+	// sandbox-independent rules are checked here; revision-dependent checks
+	// (memory downscale, QoS class change) run per candidate in
+	// sandboxcr.preCheckCandidate.
 	if claim.Spec.InplaceUpdate != nil && claim.Spec.InplaceUpdate.Resources != nil {
 		res := claim.Spec.InplaceUpdate.Resources
 		if err := sandboxcr.ValidateResizeResources(res.Requests, res.Limits); err != nil {
@@ -156,18 +158,6 @@ func (c *commonControl) EnsureClaimClaiming(ctx context.Context, args ClaimArgs)
 			c.recorder.Event(claim, "Warning", "InvalidInplaceUpdateResources", msg)
 			TransitionToCompleted(args.NewStatus, "InvalidInplaceUpdateResources", msg)
 			return NoRequeue(), nil
-		}
-		if sandboxSet.Spec.Template != nil {
-			for i := range sandboxSet.Spec.Template.Spec.Containers {
-				templateContainer := &sandboxSet.Spec.Template.Spec.Containers[i]
-				if err := inplaceupdate.CheckContainerMemoryDownscale(templateContainer, res.Requests, res.Limits); err != nil {
-					msg := fmt.Sprintf("container %q: %v", templateContainer.Name, err)
-					log.Info(msg)
-					c.recorder.Event(claim, "Warning", "MemoryDownscaleNotSupported", msg)
-					TransitionToCompleted(args.NewStatus, "MemoryDownscaleNotSupported", msg)
-					return NoRequeue(), nil
-				}
-			}
 		}
 	}
 
--- a/pkg/sandbox-manager/infra/sandboxcr/claim.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/claim.go
@@ -77,9 +77,12 @@ func ValidateAndInitClaimOptions(opts infra.ClaimSandboxOptions) (infra.ClaimSan
 			return infra.ClaimSandboxOptions{}, fmt.Errorf("resources must specify at least one of requests or limits")
 		}
 		for _, rl := range []corev1.ResourceList{res.Requests, res.Limits} {
-			if cpu, ok := rl[corev1.ResourceCPU]; ok {
-				if cpu.IsZero() || cpu.Cmp(resource.Quantity{}) < 0 {
-					return infra.ClaimSandboxOptions{}, fmt.Errorf("target cpu must be a positive value")
+			for resourceName, quantity := range rl {
+				if !supportedResizeResources[resourceName] {
+					continue
+				}
+				if quantity.IsZero() || quantity.Cmp(resource.Quantity{}) < 0 {
+					return infra.ClaimSandboxOptions{}, fmt.Errorf("target %s must be a positive value", resourceName)
 				}
 			}
 		}
@@ -679,7 +682,8 @@ func buildResourceResizedPod(pod *corev1.Pod, requests, limits corev1.ResourceLi
 
 // supportedResizeResources defines which resources are allowed for in-place resize.
 var supportedResizeResources = map[corev1.ResourceName]bool{
-	corev1.ResourceCPU: true,
+	corev1.ResourceCPU:    true,
+	corev1.ResourceMemory: true,
 }
 
 // setContainerResources updates the container's requests and limits for resources
--- a/pkg/sandbox-manager/infra/sandboxcr/claim.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/claim.go
@@ -88,9 +88,12 @@ func ValidateAndInitClaimOptions(opts infra.ClaimSandboxOptions) (infra.ClaimSan
 			return infra.ClaimSandboxOptions{}, fmt.Errorf("resources must specify at least one of requests or limits")
 		}
 		for _, rl := range []corev1.ResourceList{res.Requests, res.Limits} {
-			if cpu, ok := rl[corev1.ResourceCPU]; ok {
-				if cpu.IsZero() || cpu.Cmp(resource.Quantity{}) < 0 {
-					return infra.ClaimSandboxOptions{}, fmt.Errorf("target cpu must be a positive value")
+			for resourceName, quantity := range rl {
+				if !supportedResizeResources[resourceName] {
+					continue
+				}
+				if quantity.IsZero() || quantity.Cmp(resource.Quantity{}) < 0 {
+					return infra.ClaimSandboxOptions{}, fmt.Errorf("target %s must be a positive value", resourceName)
 				}
 			}
 		}
@@ -833,7 +836,8 @@ func buildResourceResizedPod(pod *corev1.Pod, requests, limits corev1.ResourceLi
 
 // supportedResizeResources defines which resources are allowed for in-place resize.
 var supportedResizeResources = map[corev1.ResourceName]bool{
-	corev1.ResourceCPU: true,
+	corev1.ResourceCPU:    true,
+	corev1.ResourceMemory: true,
 }
 
 // setContainerResources updates the container's requests and limits for resources
--- a/pkg/sandbox-manager/infra/sandboxcr/claim.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/claim.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand/v2"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -44,12 +45,14 @@ import (
 	"github.com/openkruise/agents/pkg/identity"
 	"github.com/openkruise/agents/pkg/sandbox-manager/config"
 	"github.com/openkruise/agents/pkg/sandbox-manager/consts"
+	managererrors "github.com/openkruise/agents/pkg/sandbox-manager/errors"
 	"github.com/openkruise/agents/pkg/sandbox-manager/infra"
 	"github.com/openkruise/agents/pkg/sandbox-manager/logs"
 	"github.com/openkruise/agents/pkg/servers/e2b/models"
 	"github.com/openkruise/agents/pkg/tracing"
 	"github.com/openkruise/agents/pkg/utils"
 	"github.com/openkruise/agents/pkg/utils/expectations"
+	"github.com/openkruise/agents/pkg/utils/inplaceupdate"
 	"github.com/openkruise/agents/pkg/utils/runtime"
 	timeoututils "github.com/openkruise/agents/pkg/utils/timeout"
 
@@ -104,46 +107,70 @@ func ValidateAndInitClaimOptions(opts infra.ClaimSandboxOptions) (infra.ClaimSan
 	return opts, nil
 }
 
-// validateInplaceUpdateResources validates in-place update resource targets.
-// Unsupported resource names are rejected, values must be positive, and
-// requests must not exceed limits. Validation runs in a fixed order so error
-// messages are deterministic regardless of map iteration order.
+// validateInplaceUpdateResources validates in-place update resource targets
+// through the shared ValidateResizeResources.
 func validateInplaceUpdateResources(inplace *config.InplaceUpdateOptions) error {
 	if inplace == nil || inplace.Resources == nil {
 		return nil
 	}
-	res := inplace.Resources
-	if len(res.Requests) == 0 && len(res.Limits) == 0 {
-		return fmt.Errorf("resources must specify at least one of requests or limits")
+	return ValidateResizeResources(inplace.Resources.Requests, inplace.Resources.Limits)
+}
+
+// ValidateResizeResources validates in-place resize targets shared by the E2B
+// create path and the SandboxClaim controller. Unsupported resource names are
+// rejected, values must be positive, and requests must not exceed limits.
+// Validation runs in a fixed order so error messages are deterministic
+// regardless of map iteration order. All returned errors are classified as
+// ErrorBadRequest so callers map them to HTTP 400.
+func ValidateResizeResources(requests, limits corev1.ResourceList) error {
+	if len(requests) == 0 && len(limits) == 0 {
+		return managererrors.NewError(managererrors.ErrorBadRequest, "resources must specify at least one of requests or limits")
 	}
-	for name := range res.Requests {
-		if !supportedResizeResources[name] {
-			return fmt.Errorf("resource %s is not supported for in-place resize", name)
-		}
-	}
-	for name := range res.Limits {
-		if !supportedResizeResources[name] {
-			return fmt.Errorf("resource %s is not supported for in-place resize", name)
-		}
+	unknown := unsupportedResourceNames(requests, limits)
+	if len(unknown) > 0 {
+		return managererrors.NewError(managererrors.ErrorBadRequest, "resource %s is not supported for in-place resize", unknown[0])
 	}
 	for _, name := range []corev1.ResourceName{corev1.ResourceCPU, corev1.ResourceMemory} {
-		if quantity, ok := res.Requests[name]; ok {
+		if quantity, ok := requests[name]; ok {
 			if quantity.IsZero() || quantity.Cmp(resource.Quantity{}) < 0 {
-				return fmt.Errorf("target %s must be a positive value", name)
+				return managererrors.NewError(managererrors.ErrorBadRequest, "target %s must be a positive value", name)
 			}
 		}
-		if quantity, ok := res.Limits[name]; ok {
+		if quantity, ok := limits[name]; ok {
 			if quantity.IsZero() || quantity.Cmp(resource.Quantity{}) < 0 {
-				return fmt.Errorf("target %s must be a positive value", name)
+				return managererrors.NewError(managererrors.ErrorBadRequest, "target %s must be a positive value", name)
 			}
 		}
-		req, hasReq := res.Requests[name]
-		lim, hasLim := res.Limits[name]
+		req, hasReq := requests[name]
+		lim, hasLim := limits[name]
 		if hasReq && hasLim && req.Cmp(lim) > 0 {
-			return fmt.Errorf("target %s request %s must not exceed limit %s", name, req.String(), lim.String())
+			return managererrors.NewError(managererrors.ErrorBadRequest, "target %s request %s must not exceed limit %s", name, req.String(), lim.String())
 		}
 	}
 	return nil
+}
+
+// unsupportedResourceNames returns the sorted resource names present in the
+// given lists that are not allowed for in-place resize, so the reported name
+// is deterministic even with several invalid keys.
+func unsupportedResourceNames(requests, limits corev1.ResourceList) []corev1.ResourceName {
+	seen := make(map[corev1.ResourceName]struct{}, len(requests)+len(limits))
+	for name := range requests {
+		if !supportedResizeResources[name] {
+			seen[name] = struct{}{}
+		}
+	}
+	for name := range limits {
+		if !supportedResizeResources[name] {
+			seen[name] = struct{}{}
+		}
+	}
+	names := make([]corev1.ResourceName, 0, len(seen))
+	for name := range seen {
+		names = append(names, name)
+	}
+	sort.Slice(names, func(i, j int) bool { return names[i] < names[j] })
+	return names
 }
 
 // chooseLockString returns the lock string to use for the current attempt.
@@ -745,7 +772,7 @@ func modifyPickedSandbox(sbx *Sandbox, lockType infra.LockType, opts infra.Claim
 		}
 		if opts.InplaceUpdate.Resources != nil {
 			if err := sbx.SetResources(opts.InplaceUpdate.Resources.Requests, opts.InplaceUpdate.Resources.Limits); err != nil {
-				return terminalMutationError{stage: "inplace update", err: err}
+				return terminalValidationError{err: err}
 			}
 		}
 	}
@@ -882,12 +909,16 @@ var supportedResizeResources = map[corev1.ResourceName]bool{
 }
 
 // setContainerResources updates the container's requests and limits for resources
-// listed in supportedResizeResources. Unsupported resource types are silently ignored.
+// listed in supportedResizeResources. Unsupported resource types are skipped at
+// this layer; callers must validate targets with ValidateResizeResources first.
 // A resource is also skipped if it was not originally set on the container.
 // Memory may only be increased: a target memory request or limit lower than the
-// container's current value returns an error because in-place memory downscale
-// is not supported.
+// container's current value returns a BadRequest error because in-place memory
+// downscale is not supported.
 func setContainerResources(container *corev1.Container, requests, limits corev1.ResourceList) (bool, error) {
+	if err := inplaceupdate.CheckContainerMemoryDownscale(container, requests, limits); err != nil {
+		return false, managererrors.WrapError(managererrors.ErrorBadRequest, err, "%s", err.Error())
+	}
 	changed := false
 	for resName, target := range requests {
 		if !supportedResizeResources[resName] {
@@ -896,9 +927,6 @@ func setContainerResources(container *corev1.Container, requests, limits corev1.
 		cur, ok := container.Resources.Requests[resName]
 		if !ok || cur.IsZero() {
 			continue
-		}
-		if resName == corev1.ResourceMemory && target.Cmp(cur) < 0 {
-			return false, fmt.Errorf("target memory request %s must not be lower than the current value %s: in-place memory downscale is not supported", target.String(), cur.String())
 		}
 		if container.Resources.Requests == nil {
 			container.Resources.Requests = corev1.ResourceList{}
@@ -913,9 +941,6 @@ func setContainerResources(container *corev1.Container, requests, limits corev1.
 		cur, ok := container.Resources.Limits[resName]
 		if !ok || cur.IsZero() {
 			continue
-		}
-		if resName == corev1.ResourceMemory && target.Cmp(cur) < 0 {
-			return false, fmt.Errorf("target memory limit %s must not be lower than the current value %s: in-place memory downscale is not supported", target.String(), cur.String())
 		}
 		if container.Resources.Limits == nil {
 			container.Resources.Limits = corev1.ResourceList{}
--- a/pkg/sandbox-manager/infra/sandboxcr/claim.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/claim.go
@@ -42,6 +42,7 @@ import (
 	cacheutils "github.com/openkruise/agents/pkg/cache/utils"
 	"github.com/openkruise/agents/pkg/controller/sandboxset"
 	"github.com/openkruise/agents/pkg/identity"
+	"github.com/openkruise/agents/pkg/sandbox-manager/config"
 	"github.com/openkruise/agents/pkg/sandbox-manager/consts"
 	"github.com/openkruise/agents/pkg/sandbox-manager/infra"
 	"github.com/openkruise/agents/pkg/sandbox-manager/logs"
@@ -82,21 +83,8 @@ func ValidateAndInitClaimOptions(opts infra.ClaimSandboxOptions) (infra.ClaimSan
 	if opts.InplaceUpdate != nil && opts.InplaceUpdate.Image == "" && opts.InplaceUpdate.Resources == nil {
 		return infra.ClaimSandboxOptions{}, fmt.Errorf("inplace update requires at least one of image or resources to be set")
 	}
-	if opts.InplaceUpdate != nil && opts.InplaceUpdate.Resources != nil {
-		res := opts.InplaceUpdate.Resources
-		if len(res.Requests) == 0 && len(res.Limits) == 0 {
-			return infra.ClaimSandboxOptions{}, fmt.Errorf("resources must specify at least one of requests or limits")
-		}
-		for _, rl := range []corev1.ResourceList{res.Requests, res.Limits} {
-			for resourceName, quantity := range rl {
-				if !supportedResizeResources[resourceName] {
-					continue
-				}
-				if quantity.IsZero() || quantity.Cmp(resource.Quantity{}) < 0 {
-					return infra.ClaimSandboxOptions{}, fmt.Errorf("target %s must be a positive value", resourceName)
-				}
-			}
-		}
+	if err := validateInplaceUpdateResources(opts.InplaceUpdate); err != nil {
+		return infra.ClaimSandboxOptions{}, err
 	}
 	if opts.CandidateCounts <= 0 {
 		opts.CandidateCounts = consts.DefaultPoolingCandidateCounts
@@ -114,6 +102,48 @@ func ValidateAndInitClaimOptions(opts infra.ClaimSandboxOptions) (infra.ClaimSan
 		opts.ReserveFailedSandboxFor = ptr.To(DefaultReserveFailedSandboxFor)
 	}
 	return opts, nil
+}
+
+// validateInplaceUpdateResources validates in-place update resource targets.
+// Unsupported resource names are rejected, values must be positive, and
+// requests must not exceed limits. Validation runs in a fixed order so error
+// messages are deterministic regardless of map iteration order.
+func validateInplaceUpdateResources(inplace *config.InplaceUpdateOptions) error {
+	if inplace == nil || inplace.Resources == nil {
+		return nil
+	}
+	res := inplace.Resources
+	if len(res.Requests) == 0 && len(res.Limits) == 0 {
+		return fmt.Errorf("resources must specify at least one of requests or limits")
+	}
+	for name := range res.Requests {
+		if !supportedResizeResources[name] {
+			return fmt.Errorf("resource %s is not supported for in-place resize", name)
+		}
+	}
+	for name := range res.Limits {
+		if !supportedResizeResources[name] {
+			return fmt.Errorf("resource %s is not supported for in-place resize", name)
+		}
+	}
+	for _, name := range []corev1.ResourceName{corev1.ResourceCPU, corev1.ResourceMemory} {
+		if quantity, ok := res.Requests[name]; ok {
+			if quantity.IsZero() || quantity.Cmp(resource.Quantity{}) < 0 {
+				return fmt.Errorf("target %s must be a positive value", name)
+			}
+		}
+		if quantity, ok := res.Limits[name]; ok {
+			if quantity.IsZero() || quantity.Cmp(resource.Quantity{}) < 0 {
+				return fmt.Errorf("target %s must be a positive value", name)
+			}
+		}
+		req, hasReq := res.Requests[name]
+		lim, hasLim := res.Limits[name]
+		if hasReq && hasLim && req.Cmp(lim) > 0 {
+			return fmt.Errorf("target %s request %s must not exceed limit %s", name, req.String(), lim.String())
+		}
+	}
+	return nil
 }
 
 // chooseLockString returns the lock string to use for the current attempt.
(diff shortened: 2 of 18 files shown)


furykerry:
## Code Review — memory resize for sandbox claims

Reviewed at head `21a6187fd55f373a8aaf3271a4358fcf3d999ff5` (fetched from `refs/pull/519/head`). Line numbers below refer to that commit.

### Change Summary

**What changed** — Enables in-place **memory** resize for sandbox claims, which previously only honored CPU. The infra-layer whitelist `supportedResizeResources` gains `corev1.ResourceMemory`, its positivity validation is generalized from a CPU-specific check to a loop over all supported resources, the E2B adapter gains two new metadata extension keys (`kruise.io/memory-request`, `kruise.io/memory-limit`) via a table-driven parse loop, the E2B clone path now rejects `InplaceUpdate.Resources` (not just `Image`), and the CRD field docs are updated ("CPU and memory are supported"). No new API fields; ~1000 of the 1147 added lines are tests.

**Why** — Feature request #515. Kubernetes in-place pod resize supports both CPU and memory, with restart behavior governed by the container's `resizePolicy`; the CPU-only restriction was an artificial limitation (the `TODO` in `sandboxclaim_types.go`). The stated goal includes preserving the warm pod's existing `resizePolicy`.

**Affected modules & external behavior**
- **sandbox-manager / infra (`infra/sandboxcr`)**: `ValidateAndInitClaimOptions` and `setContainerResources` now accept memory.
- **API layer (`pkg/servers/e2b`)**: two new metadata extension keys; clone requests carrying `Resources` now return **400** where they were previously silently ignored — a backward-incompatible behavior change for existing clone callers that pass `kruise.io/cpu-request`.
- **CRD `SandboxClaim`**: description-only change; `config/crd/bases/...` is consistent with `make manifests`, no hand edits, and no deepcopy change is needed since no field was added.
- **controller / `pkg/utils/inplaceupdate`**: production changes are comments and log strings only — the QoS pre-check, infeasible/deferred handling, and resize-patch builder were already resource-agnostic.
- Backward compatible for the CRD path; the clone path is a breaking API behavior change.

**Recommended reading order**
1. `api/v1alpha1/sandboxclaim_types.go:126-140` — field semantics that define the whole feature contract ("CPU and memory are supported", QoS invariant, "ignored if originally unset").
2. `pkg/sandbox-manager/infra/sandboxcr/claim.go:85-99, 837-869` — the actual enablement: validation loop, `supportedResizeResources` whitelist, `setContainerResources`.
3. `pkg/servers/e2b/models/extensions.go:54-55, 240-276` — E2B extension keys and the refactored parse loop that produces `InplaceUpdate.Resources`.
4. `pkg/servers/e2b/create.go:72-84, 111-113, 242-247` — **read together**: the pre-existing memory rejection guard vs. the new clone rejection. This is where the feature breaks (Critical #1).
5. `pkg/utils/inplaceupdate/inplace_update.go:248-276, 558-579, 654-666` — not modified, but decides whether a resize is emitted and when it is considered complete; the `>=` "covered" semantics is what makes a downscale a no-op (Warnings #2/#3).
6. `pkg/utils/inplaceupdate/inplace_update_test.go:64-110, 413-470` — new `applyResizeSubresourcePatch` fake and the `resizePolicy`-must-not-leak guard; explains how "preserve resizePolicy" is verified.
7. `test/e2e/sandboxclaim_test.go:1074-1660` — the four new specs (RestartContainer, CPU NotRequired, memory NotRequired, init-container non-mutation) that document intended runtime behavior.

---

### Critical Issues (Must Fix)

#### 1. Memory resize is still hard-rejected on the E2B create path — the E2B half of the feature is unreachable

- **Location**: `pkg/servers/e2b/create.go:72-84` (guard), `:111-113` (call site), `pkg/servers/e2b/create_test.go:995-1018` (test locking the old behavior)
- **Issue**: `CreateSandbox` calls `validateCreateResourceOverride` immediately after request parsing, and that function still returns `400 "memory inplace update is not supported"` whenever `Resources.Requests[memory]` or `Resources.Limits[memory]` is present. The new extension keys are parsed successfully and then unconditionally rejected before dispatch to `createSandboxWithClaim`:

```go
 72: func validateCreateResourceOverride(request models.NewSandboxRequest) *web.ApiError {
 77:     if _, ok := res.Requests[corev1.ResourceMemory]; ok {
 78:             return &web.ApiError{Code: http.StatusBadRequest, Message: "memory inplace update is not supported"}
111:     if validateErr := validateCreateResourceOverride(request); validateErr != nil {
```

- **Why**: The PR description states memory resize should be supported "through both the CRD API and the E2B-compatible API path", but `ExtensionKeyClaimWithMemoryRequest` / `ExtensionKeyClaimWithMemoryLimit` cannot work end-to-end. Worse, the branch still carries `TestCreateSandbox_MemoryOverrideRejectedBeforeQuotaAcquire`, which asserts the rejection — so CI stays green while the feature is broken, and the PR contradicts itself (`models/extensions_test.go` asserts memory parses into `InplaceUpdate.Resources`; `create_test.go` asserts the server rejects exactly that). This most likely came from merging a newer master (the guard landed while this PR was open) without reconciling.
- **Fix**: remove `validateCreateResourceOverride` and its invocation, and flip the test into a positive assertion, e.g.

```go
func TestBuildClaimOptions_MemoryOverridePropagated(t *testing.T) {
	req := models.NewSandboxRequest{
		TemplateID: "claim-template",
		Extensions: models.NewSandboxRequestExtension{
			InplaceUpdate: models.InplaceUpdateExtension{
				Resources: &models.InplaceUpdateResourcesExtension{
					Requests: corev1.ResourceList{corev1.ResourceMemory: resource.MustParse("512Mi")},
					Limits:   corev1.ResourceList{corev1.ResourceMemory: resource.MustParse("1Gi")},
				},
			},
		},
	}
	opts := /* buildClaimOptions(...) */
	assert.Equal(t, resource.MustParse("512Mi"), opts.InplaceUpdate.Resources.Requests[corev1.ResourceMemory])
	assert.Equal(t, resource.MustParse("1Gi"), opts.InplaceUpdate.Resources.Limits[corev1.ResourceMemory])
}
```

Please also add one E2B-level test (or E2B E2E case) that creates a sandbox with `kruise.io/memory-request` and asserts the resulting pod memory, otherwise this path stays untested.

---

### Warnings (Should Fix)

#### 2. A downscale-only resize is silently a no-op, and the docs now imply otherwise

- **Location**: `api/v1alpha1/sandboxclaim_types.go:132-140` (doc); behavior in `pkg/utils/inplaceupdate/inplace_update.go:263` and `:654-666`
- **Issue**: `DefaultBuildResizeContainers` skips a container when `IsResourceSatisfied(desired, actual)`, and `isResourceListCovered` treats `actual >= desired` as satisfied (documented at `:654-657`). A claim that only *lowers* memory (pool 256Mi → claim 128Mi) therefore produces **no resize patch at all**: `isMetadataOnlyChange` (`pkg/controller/sandbox/core/common_inplace_update_handler.go:272`) also returns true, the claim reaches `Completed`, and the pod keeps its original memory. The Sandbox CR template then permanently disagrees with the live pod, which later resurfaces as a container delta in `pkg/controller/checkpoint/utils.go:60`.
- **Why it matters more now**: for CPU this was a latent oddity; for memory, "claim a smaller sandbox out of a large-memory pool" is a primary use case, and the API comment updated by this PR ("CPU and memory are supported") gives no hint that downscaling is ignored.
- **Recommendation**: reject a memory downscale explicitly instead of silently dropping it (same fix as #3), and state the rule in the field doc:

```go
// Requests specifies the target resource requests for each container.
// CPU and memory are supported. The container's original request must already be set;
// otherwise the value is ignored. Memory may only be increased: a target lower than the
// container's current memory is rejected.
```

#### 3. Grow + downscale in one claim can report success before the downscale is applied — reject memory downscale rather than tightening the comparison

- **Location**: `pkg/utils/inplaceupdate/inplace_update.go:558-579` (`isPodResourceResizeCompleted` → `IsResourceSatisfied`)
- **Issue**: When a claim raises CPU and lowers memory, a resize patch *is* emitted (CPU triggers it) carrying the whole `origin.Resources`, including the smaller memory. Completion is then judged with `actual >= desired`, so the memory part is considered satisfied the moment it is observed, even though the kubelet has not applied it yet — and with `resizePolicy: {memory, RestartContainer}` applying it means **killing and restarting the container**. The claim can therefore hand the caller a sandbox that is about to restart, losing warm runtime state.
- **Do not** switch `isPodResourceResizeCompleted` to `ResourcesExactlyEqual`: resources are legitimately normalized/rounded on the way to the live pod (cgroup / page alignment on the kubelet side, LimitRange or mutating-webhook adjustments, CPU unit normalization), which is exactly why `isResourceListCovered` uses `Cmp` with `>=` today. Exact equality would make the completion check flaky and could block the wait indefinitely on clusters that round values.
- **Recommendation**: keep the tolerant `>=` comparison and remove the problematic direction instead — **reject a memory downscale up front**. Once no downscale can reach the resize path, `actual >= desired` is correct by construction and no completion-check change is needed.
  - Primary guard, sandbox-manager path — `setContainerResources` / `Sandbox.SetResources` is the only place that sees both the current value and the target, and `modifyPickedSandbox` (`claim.go:702`) already returns an error, so plumbing one through is cheap:

```go
// SetResources applies in-place resource resize to the first container.
func (s *Sandbox) SetResources(requests, limits corev1.ResourceList) error { ... }

// inside setContainerResources, for memory only:
if resName == corev1.ResourceMemory && target.Cmp(cur) < 0 {
	return fmt.Errorf("target memory %s must not be lower than the current value %s: "+
		"in-place memory downscale is not supported", target.String(), cur.String())
}
```

  - Defense in depth, controller path — add the same direction check next to the existing QoS pre-check (`common_inplace_update_handler.go:170-183`) so a directly edited Sandbox CR fails with an `InplaceUpdate` condition carrying a clear message, instead of hanging or diverging.
  - If a memory downscale is wanted later, it needs its own design (wait for the restart to finish and the pod to become ready again before completing the claim, or a memory-specific direction-aware comparison with a normalization tolerance). Rejecting it now keeps the semantics honest.

#### 4. No `requests <= limits` validation — fails late, after a warm sandbox is already locked

- **Location**: `pkg/sandbox-manager/infra/sandboxcr/claim.go:85-99`
- **Issue**: Validation only checks positivity per resource. `memory-request: 1Gi` + `memory-limit: 512Mi` (an easy Mi/Gi mistake) passes admission, a warm sandbox is picked and locked, and the failure surfaces as an opaque apiserver validation error during the pod update — burning a pooled sandbox and returning a confusing message.
- **Recommendation**:

```go
for _, name := range []corev1.ResourceName{corev1.ResourceCPU, corev1.ResourceMemory} {
	req, hasReq := res.Requests[name]
	lim, hasLim := res.Limits[name]
	if hasReq && hasLim && req.Cmp(lim) > 0 {
		return infra.ClaimSandboxOptions{}, fmt.Errorf("target %s request %s must not exceed limit %s", name, req.String(), lim.String())
	}
}
```

#### 5. Restart semantics of memory resize are undocumented in the API

- **Location**: `api/v1alpha1/sandboxclaim_types.go:126-142`
- **Issue**: This is the one place where memory differs qualitatively from CPU: with the common `resizePolicy: {memory, RestartContainer}` — which the new E2E spec deliberately exercises — a memory resize **restarts the sandbox container** and discards in-container state. Nothing in the field docs warns a claim author about that.
- **Recommendation**: add one sentence to the `Resources` doc comment and regenerate the CRD: "Memory resize may restart the container when the pod template declares `resizePolicy: {resourceName: memory, restartPolicy: RestartContainer}`."

#### 6. Test-suite shape: three near-duplicate E2E specs, and missing negative coverage for memory

- **Location**: `test/e2e/sandboxclaim_test.go:1074-1660`
- **Issue**: The RestartContainer / CPU-NotRequired / memory-NotRequired specs are ~200 lines each and differ only in `resizePolicy`, target resources, and restart expectation; each creates its own `SandboxSet` with up-to-2-minute pool waits plus up-to-4-minute claim waits, materially lengthening the suite. The warm-pod discovery and restart-count recording blocks are copy-pasted verbatim, and repo convention prefers table-driven tests.
- **Recommendation**: collapse into a `DescribeTable` over `(policy, resourceName, target, expectRestart)` with shared helpers such as `createResizePolicyPool(...)`, `waitWarmSandbox(...)`, `recordMainContainerIdentity(...)`. Also missing: a memory-specific QoS-rejection case (only "CPU resize that would change QoS class" exists at `:390`) and a case asserting the new memory-downscale rejection from #2/#3.

---

### Suggestions (Consider Improving)

- `pkg/servers/e2b/models/extensions.go:248-256`: the limit rows rely on the implicit zero value of `isRequest`, which reads like an omission. Prefer an explicit field value or a small enum such as `target: targetRequests|targetLimits`.
- `pkg/sandbox-manager/infra/sandboxcr/claim.go:90-99`: iterating a map makes the reported resource name nondeterministic when both CPU and memory are invalid. Validating in a fixed order (`[]corev1.ResourceName{cpu, memory}`) keeps error messages stable and testable.
- `claim.go:844` ("Unsupported resource types are silently ignored"): now that the whitelist is user-facing for two resources, consider rejecting unknown resource names (e.g. `nvidia.com/gpu`, `ephemeral-storage`) in `ValidateAndInitClaimOptions` so callers learn their request had no effect.
- `supportedResizeResources` as `map[corev1.ResourceName]bool` could be a `sets.Set[corev1.ResourceName]` for intent clarity — purely cosmetic.
- The widened clone-path rejection (`create.go:242`) turns previously-succeeding requests into 400s; worth calling out explicitly in the PR description / release notes.

---

### Positive Findings

- **Layering respected**: protocol parsing stays in `pkg/servers/e2b/models`, the resource whitelist stays in `infra/sandboxcr`, and the feature gate stays controller-side (`common_control.go:136-139`) and was already resource-agnostic — no API→Infra shortcuts and no `pkg/features` leakage into the manager path.
- **Good reuse instead of duplication**: `CheckResizeQoSChange`, `computeQoSClass`, `checkPodResizeInfeasible` and `buildResourcePatch` already handled memory correctly, so the diff correctly touches none of them.
- **`applyResizeSubresourcePatch` (`inplace_update_test.go:67-110`) is a real improvement** over the previous `c.Patch`: it emulates the key-wise strategic-merge semantics of `pods/resize`, so tests no longer silently drop system-injected resources such as `ephemeral-storage`; the added `resizePolicy` and `initContainers` assertions (`:1283`) are useful regression guards for the "preserve resizePolicy" requirement.
- **Test refactors follow project convention**: `TestValidateAndInitClaimOptions_ResourceResize` and `TestCreateSandboxWithClone_InplaceUpdateRejected` were converted from single-case to table-driven with descriptive names.
- **Generated-file discipline**: the CRD yaml contains only the regenerated description text matching the Go doc comment; no manual CRD or `zz_generated.deepcopy.go` edits.
- **E2E covers the interesting runtime matrix** (RestartContainer vs NotRequired, restart-count and container-ID identity checks, init-container non-mutation) rather than only asserting spec fields — and it passes on the CI kind cluster (K8s 1.32 with `InPlacePodVerticalScaling=true`, i.e. the direct-patch fallback path, not just the 1.33 subresource path).

PersistentJZH:
> ## Code Review — memory resize for sandbox claims
> Reviewed at head `21a6187fd55f373a8aaf3271a4358fcf3d999ff5` (fetched from `refs/pull/519/head`). Line numbers below refer to that commit.
> 
> ### Change Summary
> **What changed** — Enables in-place **memory** resize for sandbox claims, which previously only honored CPU. The infra-layer whitelist `supportedResizeResources` gains `corev1.ResourceMemory`, its positivity validation is generalized from a CPU-specific check to a loop over all supported resources, the E2B adapter gains two new metadata extension keys (`kruise.io/memory-request`, `kruise.io/memory-limit`) via a table-driven parse loop, the E2B clone path now rejects `InplaceUpdate.Resources` (not just `Image`), and the CRD field docs are updated ("CPU and memory are supported"). No new API fields; ~1000 of the 1147 added lines are tests.
> 
> **Why** — Feature request #515. Kubernetes in-place pod resize supports both CPU and memory, with restart behavior governed by the container's `resizePolicy`; the CPU-only restriction was an artificial limitation (the `TODO` in `sandboxclaim_types.go`). The stated goal includes preserving the warm pod's existing `resizePolicy`.
> 
> **Affected modules & external behavior**
> 
> * **sandbox-manager / infra (`infra/sandboxcr`)**: `ValidateAndInitClaimOptions` and `setContainerResources` now accept memory.
> * **API layer (`pkg/servers/e2b`)**: two new metadata extension keys; clone requests carrying `Resources` now return **400** where they were previously silently ignored — a backward-incompatible behavior change for existing clone callers that pass `kruise.io/cpu-request`.
> * **CRD `SandboxClaim`**: description-only change; `config/crd/bases/...` is consistent with `make manifests`, no hand edits, and no deepcopy change is needed since no field was added.
> * **controller / `pkg/utils/inplaceupdate`**: production changes are comments and log strings only — the QoS pre-check, infeasible/deferred handling, and resize-patch builder were already resource-agnostic.
> * Backward compatible for the CRD path; the clone path is a breaking API behavior change.
> 
> **Recommended reading order**
> 
> 1. `api/v1alpha1/sandboxclaim_types.go:126-140` — field semantics that define the whole feature contract ("CPU and memory are supported", QoS invariant, "ignored if originally unset").
> 2. `pkg/sandbox-manager/infra/sandboxcr/claim.go:85-99, 837-869` — the actual enablement: validation loop, `supportedResizeResources` whitelist, `setContainerResources`.
> 3. `pkg/servers/e2b/models/extensions.go:54-55, 240-276` — E2B extension keys and the refactored parse loop that produces `InplaceUpdate.Resources`.
> 4. `pkg/servers/e2b/create.go:72-84, 111-113, 242-247` — **read together**: the pre-existing memory rejection guard vs. the new clone rejection. This is where the feature breaks (Critical [init sandbox and snadboxset controller #1](https://github.com/openkruise/agents/pull/1)).
> 5. `pkg/utils/inplaceupdate/inplace_update.go:248-276, 558-579, 654-666` — not modified, but decides whether a resize is emitted and when it is considered complete; the `>=` "covered" semantics is what makes a downscale a no-op (Warnings [[feature request] support scale sub-resource for sandboxset #2](https://github.com/openkruise/agents/issues/2)/[[feature request] support autoscaler for sandboxset #3](https://github.com/openkruise/agents/issues/3)).
> 6. `pkg/utils/inplaceupdate/inplace_update_test.go:64-110, 413-470` — new `applyResizeSubresourcePatch` fake and the `resizePolicy`-must-not-leak guard; explains how "preserve resizePolicy" is verified.
> 7. `test/e2e/sandboxclaim_test.go:1074-1660` — the four new specs (RestartContainer, CPU NotRequired, memory NotRequired, init-container non-mutation) that document intended runtime behavior.
> 
> ### Critical Issues (Must Fix)
> #### 1. Memory resize is still hard-rejected on the E2B create path — the E2B half of the feature is unreachable
> * **Location**: `pkg/servers/e2b/create.go:72-84` (guard), `:111-113` (call site), `pkg/servers/e2b/create_test.go:995-1018` (test locking the old behavior)
> * **Issue**: `CreateSandbox` calls `validateCreateResourceOverride` immediately after request parsing, and that function still returns `400 "memory inplace update is not supported"` whenever `Resources.Requests[memory]` or `Resources.Limits[memory]` is present. The new extension keys are parsed successfully and then unconditionally rejected before dispatch to `createSandboxWithClaim`:
> 
> ```go
>  72: func validateCreateResourceOverride(request models.NewSandboxRequest) *web.ApiError {
>  77:     if _, ok := res.Requests[corev1.ResourceMemory]; ok {
>  78:             return &web.ApiError{Code: http.StatusBadRequest, Message: "memory inplace update is not supported"}
> 111:     if validateErr := validateCreateResourceOverride(request); validateErr != nil {
> ```
> 
> * **Why**: The PR description states memory resize should be supported "through both the CRD API and the E2B-compatible API path", but `ExtensionKeyClaimWithMemoryRequest` / `ExtensionKeyClaimWithMemoryLimit` cannot work end-to-end. Worse, the branch still carries `TestCreateSandbox_MemoryOverrideRejectedBeforeQuotaAcquire`, which asserts the rejection — so CI stays green while the feature is broken, and the PR contradicts itself (`models/extensions_test.go` asserts memory parses into `InplaceUpdate.Resources`; `create_test.go` asserts the server rejects exactly that). This most likely came from merging a newer master (the guard landed while this PR was open) without reconciling.
> * **Fix**: remove `validateCreateResourceOverride` and its invocation, and flip the test into a positive assertion, e.g.
> 
> ```go
> func TestBuildClaimOptions_MemoryOverridePropagated(t *testing.T) {
> 	req := models.NewSandboxRequest{
> 		TemplateID: "claim-template",
> 		Extensions: models.NewSandboxRequestExtension{
> 			InplaceUpdate: models.InplaceUpdateExtension{
> 				Resources: &models.InplaceUpdateResourcesExtension{
> 					Requests: corev1.ResourceList{corev1.ResourceMemory: resource.MustParse("512Mi")},
> 					Limits:   corev1.ResourceList{corev1.ResourceMemory: resource.MustParse("1Gi")},
> 				},
> 			},
> 		},
> 	}
> 	opts := /* buildClaimOptions(...) */
> 	assert.Equal(t, resource.MustParse("512Mi"), opts.InplaceUpdate.Resources.Requests[corev1.ResourceMemory])
> 	assert.Equal(t, resource.MustParse("1Gi"), opts.InplaceUpdate.Resources.Limits[corev1.ResourceMemory])
> }
> ```
> 
> Please also add one E2B-level test (or E2B E2E case) that creates a sandbox with `kruise.io/memory-request` and asserts the resulting pod memory, otherwise this path stays untested.
> 
> ### Warnings (Should Fix)
> #### 2. A downscale-only resize is silently a no-op, and the docs now imply otherwise
> * **Location**: `api/v1alpha1/sandboxclaim_types.go:132-140` (doc); behavior in `pkg/utils/inplaceupdate/inplace_update.go:263` and `:654-666`
> * **Issue**: `DefaultBuildResizeContainers` skips a container when `IsResourceSatisfied(desired, actual)`, and `isResourceListCovered` treats `actual >= desired` as satisfied (documented at `:654-657`). A claim that only _lowers_ memory (pool 256Mi → claim 128Mi) therefore produces **no resize patch at all**: `isMetadataOnlyChange` (`pkg/controller/sandbox/core/common_inplace_update_handler.go:272`) also returns true, the claim reaches `Completed`, and the pod keeps its original memory. The Sandbox CR template then permanently disagrees with the live pod, which later resurfaces as a container delta in `pkg/controller/checkpoint/utils.go:60`.
> * **Why it matters more now**: for CPU this was a latent oddity; for memory, "claim a smaller sandbox out of a large-memory pool" is a primary use case, and the API comment updated by this PR ("CPU and memory are supported") gives no hint that downscaling is ignored.
> * **Recommendation**: reject a memory downscale explicitly instead of silently dropping it (same fix as [[feature request] support autoscaler for sandboxset #3](https://github.com/openkruise/agents/issues/3)), and state the rule in the field doc:
> 
> ```go
> // Requests specifies the target resource requests for each container.
> // CPU and memory are supported. The container's original request must already be set;
> // otherwise the value is ignored. Memory may only be increased: a target lower than the
> // container's current memory is rejected.
> ```
> 
> #### 3. Grow + downscale in one claim can report success before the downscale is applied — reject memory downscale rather than tightening the comparison
> * **Location**: `pkg/utils/inplaceupdate/inplace_update.go:558-579` (`isPodResourceResizeCompleted` → `IsResourceSatisfied`)
> * **Issue**: When a claim raises CPU and lowers memory, a resize patch _is_ emitted (CPU triggers it) carrying the whole `origin.Resources`, including the smaller memory. Completion is then judged with `actual >= desired`, so the memory part is considered satisfied the moment it is observed, even though the kubelet has not applied it yet — and with `resizePolicy: {memory, RestartContainer}` applying it means **killing and restarting the container**. The claim can therefore hand the caller a sandbox that is about to restart, losing warm runtime state.
> * **Do not** switch `isPodResourceResizeCompleted` to `ResourcesExactlyEqual`: resources are legitimately normalized/rounded on the way to the live pod (cgroup / page alignment on the kubelet side, LimitRange or mutating-webhook adjustments, CPU unit normalization), which is exactly why `isResourceListCovered` uses `Cmp` with `>=` today. Exact equality would make the completion check flaky and could block the wait indefinitely on clusters that round values.
> * **Recommendation**: keep the tolerant `>=` comparison and remove the problematic direction instead — **reject a memory downscale up front**. Once no downscale can reach the resize path, `actual >= desired` is correct by construction and no completion-check change is needed.
>   
>   * Primary guard, sandbox-manager path — `setContainerResources` / `Sandbox.SetResources` is the only place that sees both the current value and the target, and `modifyPickedSandbox` (`claim.go:702`) already returns an error, so plumbing one through is cheap:
> 
> ```go
> // SetResources applies in-place resource resize to the first container.
> func (s *Sandbox) SetResources(requests, limits corev1.ResourceList) error { ... }
> 
> // inside setContainerResources, for memory only:
> if resName == corev1.ResourceMemory && target.Cmp(cur) < 0 {
> 	return fmt.Errorf("target memory %s must not be lower than the current value %s: "+
> 		"in-place memory downscale is not supported", target.String(), cur.String())
> }
> ```
> 
> * Defense in depth, controller path — add the same direction check next to the existing QoS pre-check (`common_inplace_update_handler.go:170-183`) so a directly edited Sandbox CR fails with an `InplaceUpdate` condition carrying a clear message, instead of hanging or diverging.
> * If a memory downscale is wanted later, it needs its own design (wait for the restart to finish and the pod to become ready again before completing the claim, or a memory-specific direction-aware comparison with a normalization tolerance). Rejecting it now keeps the semantics honest.
> 
> #### 4. No `requests <= limits` validation — fails late, after a warm sandbox is already locked
> * **Location**: `pkg/sandbox-manager/infra/sandboxcr/claim.go:85-99`
> * **Issue**: Validation only checks positivity per resource. `memory-request: 1Gi` + `memory-limit: 512Mi` (an easy Mi/Gi mistake) passes admission, a warm sandbox is picked and locked, and the failure surfaces as an opaque apiserver validation error during the pod update — burning a pooled sandbox and returning a confusing message.
> * **Recommendation**:
> 
> ```go
> for _, name := range []corev1.ResourceName{corev1.ResourceCPU, corev1.ResourceMemory} {
> 	req, hasReq := res.Requests[name]
> 	lim, hasLim := res.Limits[name]
> 	if hasReq && hasLim && req.Cmp(lim) > 0 {
> 		return infra.ClaimSandboxOptions{}, fmt.Errorf("target %s request %s must not exceed limit %s", name, req.String(), lim.String())
> 	}
> }
> ```
> 
> #### 5. Restart semantics of memory resize are undocumented in the API
> * **Location**: `api/v1alpha1/sandboxclaim_types.go:126-142`
> * **Issue**: This is the one place where memory differs qualitatively from CPU: with the common `resizePolicy: {memory, RestartContainer}` — which the new E2E spec deliberately exercises — a memory resize **restarts the sandbox container** and discards in-container state. Nothing in the field docs warns a claim author about that.
> * **Recommendation**: add one sentence to the `Resources` doc comment and regenerate the CRD: "Memory resize may restart the container when the pod template declares `resizePolicy: {resourceName: memory, restartPolicy: RestartContainer}`."
> 
> #### 6. Test-suite shape: three near-duplicate E2E specs, and missing negative coverage for memory
> * **Location**: `test/e2e/sandboxclaim_test.go:1074-1660`
> * **Issue**: The RestartContainer / CPU-NotRequired / memory-NotRequired specs are ~200 lines each and differ only in `resizePolicy`, target resources, and restart expectation; each creates its own `SandboxSet` with up-to-2-minute pool waits plus up-to-4-minute claim waits, materially lengthening the suite. The warm-pod discovery and restart-count recording blocks are copy-pasted verbatim, and repo convention prefers table-driven tests.
> * **Recommendation**: collapse into a `DescribeTable` over `(policy, resourceName, target, expectRestart)` with shared helpers such as `createResizePolicyPool(...)`, `waitWarmSandbox(...)`, `recordMainContainerIdentity(...)`. Also missing: a memory-specific QoS-rejection case (only "CPU resize that would change QoS class" exists at `:390`) and a case asserting the new memory-downscale rejection from [[feature request] support scale sub-resource for sandboxset #2](https://github.com/openkruise/agents/issues/2)/[[feature request] support autoscaler for sandboxset #3](https://github.com/openkruise/agents/issues/3).
> 
> ### Suggestions (Consider Improving)
> * `pkg/servers/e2b/models/extensions.go:248-256`: the limit rows rely on the implicit zero value of `isRequest`, which reads like an omission. Prefer an explicit field value or a small enum such as `target: targetRequests|targetLimits`.
> * `pkg/sandbox-manager/infra/sandboxcr/claim.go:90-99`: iterating a map makes the reported resource name nondeterministic when both CPU and memory are invalid. Validating in a fixed order (`[]corev1.ResourceName{cpu, memory}`) keeps error messages stable and testable.
> * `claim.go:844` ("Unsupported resource types are silently ignored"): now that the whitelist is user-facing for two resources, consider rejecting unknown resource names (e.g. `nvidia.com/gpu`, `ephemeral-storage`) in `ValidateAndInitClaimOptions` so callers learn their request had no effect.
> * `supportedResizeResources` as `map[corev1.ResourceName]bool` could be a `sets.Set[corev1.ResourceName]` for intent clarity — purely cosmetic.
> * The widened clone-path rejection (`create.go:242`) turns previously-succeeding requests into 400s; worth calling out explicitly in the PR description / release notes.
> 
> ### Positive Findings
> * **Layering respected**: protocol parsing stays in `pkg/servers/e2b/models`, the resource whitelist stays in `infra/sandboxcr`, and the feature gate stays controller-side (`common_control.go:136-139`) and was already resource-agnostic — no API→Infra shortcuts and no `pkg/features` leakage into the manager path.
> * **Good reuse instead of duplication**: `CheckResizeQoSChange`, `computeQoSClass`, `checkPodResizeInfeasible` and `buildResourcePatch` already handled memory correctly, so the diff correctly touches none of them.
> * **`applyResizeSubresourcePatch` (`inplace_update_test.go:67-110`) is a real improvement** over the previous `c.Patch`: it emulates the key-wise strategic-merge semantics of `pods/resize`, so tests no longer silently drop system-injected resources such as `ephemeral-storage`; the added `resizePolicy` and `initContainers` assertions (`:1283`) are useful regression guards for the "preserve resizePolicy" requirement.
> * **Test refactors follow project convention**: `TestValidateAndInitClaimOptions_ResourceResize` and `TestCreateSandboxWithClone_InplaceUpdateRejected` were converted from single-case to table-driven with descriptive names.
> * **Generated-file discipline**: the CRD yaml contains only the regenerated description text matching the Go doc comment; no manual CRD or `zz_generated.deepcopy.go` edits.
> * **E2E covers the interesting runtime matrix** (RestartContainer vs NotRequired, restart-count and container-ID identity checks, init-container non-mutation) rather than only asserting spec fields — and it passes on the CI kind cluster (K8s 1.32 with `InPlacePodVerticalScaling=true`, i.e. the direct-patch fallback path, not just the 1.33 subresource path).

**For "Memory resize is still hard-rejected on the E2B create path":**

fixed. Removed validateCreateResourceOverride and its call, and replaced the old rejection test with a positive full-flow test asserting memory request/limit propagate to the claimed sandbox.



**For "A downscale-only resize is silently a no-op":**

fixed as recommended. 



**For "Grow + downscale can report success before the downscale is applied":**

this is resolved by the above rejection. 



**For "No requests <= limits validation":**

added. request > limit for CPU/memory is now rejected before any sandbox is picked or locked.



**For "Restart semantics of memory resize are undocumented":**

added the sentence to the Resources doc comment and regenerated the CRD manifests.



**For "Test-suite shape":**

added the two missing negative E2E cases (memory QoS rejection and memory downscale rejection).


**For "e2e cases collapse into a DescribeTable"**
seems no need to do

furykerry:
## Review round 2 — status of previous comments + new findings

Re-reviewed at head `7cfc1576a01b8993e510de86905256ed00d616d3` (incremental `+730/-72` over `21a6187`). This comment only covers what changed since [my previous review](https://github.com/openkruise/agents/pull/519#issuecomment-5291381644) — please read that one for the full change summary.

### Previous comments — status

| Item | Status | Note |
|---|---|---|
| C1 — E2B create path hard-rejected memory | ✅ Fixed | Guard and call site removed; `TestCreateSandbox_MemoryOverridePropagatedToClaimedSandbox` now drives the full create flow and asserts CR resources, quota acquire count, and that unrelated CPU is preserved — stronger than what I proposed |
| W2 — downscale silently a no-op | ✅ Fixed | Explicit rejection in `setContainerResources` + `CheckMemoryDownscale`, plus field docs |
| W3 — do not use `ResourcesExactlyEqual` | ✅ Fixed as revised | Completion check keeps the tolerant `>=`; the direction is removed at the source instead |
| W4 — `requests <= limits` | ⚠️ Partial | Implemented, but bypassed on the CRD path (see W2 below) |
| W5 — restart semantics in API docs | ✅ Fixed | Added to both `Requests` and `Limits`, CRD regenerated |
| W6 — E2E table-driven + memory QoS coverage | ⚠️ Partial | Memory QoS rejection spec added; the table-driven refactor was not done (now 6 near-identical specs, E2E at `+982/-19`) |
| Suggestion — explicit `isRequest` | ✅ Fixed | Replaced by a typed `resourceTarget` enum, all four rows explicit |
| Suggestion — deterministic validation errors | ✅ Partial | cpu/memory validated in fixed order; the unsupported-name loops still iterate maps (S3 below) |

---

### New Critical Issues (Must Fix)

#### C1. The new rejections return HTTP 500 instead of 400 on the E2B path, with a misleading "callback failed" message

- **Location**: `pkg/sandbox-manager/infra/sandboxcr/claim.go:744-750`, `errors.go:53-64`, `infra.go:184` → `infra.go:204-223`, `pkg/sandbox-manager/api.go:117-123`, `pkg/servers/e2b/create.go:56-69`
- **Issue**: the downscale error is wrapped in `terminalMutationError`, which carries no `managererrors` code. Call chain: `modifyPickedSandbox` → `errors.As(claimErr, &terminalMutationError{})` stops the retry → `buildClaimError`'s `errors.As(err, &mErr)` does not match → `WrapError(ErrorInternal, ...)` → `preserveTypedError` keeps Internal → `mapInfraErrorToApiError` default branch → **500**. The same applies to every `fmt.Errorf` in the new `validateInplaceUpdateResources` (unsupported resource name, request exceeding limit, non-positive value).
- **Why**: these are pure client input errors and must map to 400. In addition, `terminalMutationError.Error()` renders as `inplace update callback failed: ...`, while the type's own doc scopes it to "user-supplied mutation callback" — a validation failure is disguised as a callback failure, which misleads triage. 5xx on client input also pollutes SLO/alerting.
- **Fix** (small change; `errors.As` traverses `Unwrap`, so the terminal-retry behavior is preserved):

```go
// claim.go — setContainerResources / SetResources
if resName == corev1.ResourceMemory && target.Cmp(cur) < 0 {
    return false, managererrors.NewError(managererrors.ErrorBadRequest,
        "target memory %s must not be lower than the current value %s: in-place memory downscale is not supported",
        target.String(), cur.String())
}

// validateInplaceUpdateResources — same for every branch
return managererrors.NewError(managererrors.ErrorBadRequest, "target %s must be a positive value", name)
```

Please also add an e2b-level test asserting `apiErr.Code == http.StatusBadRequest`: the current tests only cover the success path, so CI cannot catch this.

#### C2. `CheckMemoryDownscale` runs before `isMetadataOnlyChange` and permanently blocks all in-place updates for pods whose live memory was raised by the environment

- **Location**: `pkg/controller/sandbox/core/common_inplace_update_handler.go:145-161`
- **Issue**: the check compares `box.Spec.Template` against the **live pod** and sits after the revision check but before `isMetadataOnlyChange` and the QoS pre-check, so it executes for *any* pending in-place update (image, metadata, resources) whenever `pod hash != UpdateRevision`. When the pod's actual memory was raised by the environment — LimitRange minimum, a mutating webhook, or VPA in-place actuation — `template(128Mi) < pod(256Mi)` holds, so an **image-only or metadata-only rollout is misclassified as a memory downscale**, gets `InplaceUpdate=False/Failed` and `return true, nil`, and can never progress.
- **Why**: this repo deliberately tolerates that divergence today — `common_inplace_update_handler.go:266-270` ("Extra resources injected into the pod (e.g., by LimitRanger or other admission webhooks) are ignored"), the `isResourceListCovered` doc ("some environments round up or adjust container resources"), and `pkg/controller/checkpoint/utils.go:60` which explicitly mentions "VPA-modified resources". The new check inverts that tolerance into a hard failure. Existing coverage only exercises *extra resource keys* (`common_inplace_update_handler_test.go:1539`, subset match), never *a higher value for the same key*, so green CI does not disprove this.
- **Fix**: the controller cannot distinguish "template intentionally lowers memory" from "environment raised the pod", so it should not fail terminally. Prefer making it advisory and keeping enforcement at the write paths:

```go
if err := inplaceupdate.CheckMemoryDownscale(box, pod); err != nil {
    klog.FromContext(ctx).Info("skipping memory downscale", "sandbox", klog.KObj(box), "reason", err.Error())
    handler.GetRecorder().Eventf(box, corev1.EventTypeWarning, "MemoryDownscaleSkipped", err.Error())
    // fall through: metadata-only patch, QoS pre-check and normal resize must still run
}
```

If the hard rejection is kept, please at least move it after `isMetadataOnlyChange` and add a regression test: *"pod memory raised by admission, image update still succeeds"*.

---

### New Warnings (Should Fix)

#### W1. On the CRD claim path the rejection surfaces as a ClaimTimeout spin with a misleading `NoAvailableSandboxes` event

- **Location**: `pkg/controller/sandboxclaim/core/common_control.go:152-183`; the new E2E spec ("Verifying the claim times out with zero claimed replicas")
- **Issue**: a terminal claim error is only `log.Error`'d; the controller then emits `Warning NoAvailableSandboxes "No available sandboxes in pool X"`, writes `Claiming sandboxes: 0/1 claimed`, and retries every `ClaimRetryInterval` until `ClaimTimeout`, ending as `Completed` + `ClaimedReplicas=0` + `TimedOut`. The real reason never reaches the SandboxClaim, so users will diagnose an empty pool. The new E2E freezes this behavior instead of flagging it.
- **Recommendation**: reuse the existing Step 7 precondition pattern (`common_control.go:134-145`, `FeatureGateDisabled` → `TransitionToCompleted`) and compare the claim's memory target against `sandboxSet.Spec.Template`, which is already in scope in `EnsureClaimClaiming`. That fails fast with an accurate reason and also removes the "result depends on which candidate was picked" non-determinism in heterogeneous pools.

#### W2. The three new validations are entirely bypassed on the CRD path

- **Location**: `claim.go:64-105` (`ValidateAndInitClaimOptions`) vs `claim.go:199` (`TryClaimSandbox`); caller `common_control.go:246`
- **Issue**: `ValidateAndInitClaimOptions` is only invoked from `Infra.ClaimSandbox` (`infra.go:139`). The SandboxClaim controller calls `TryClaimSandbox` directly, so the resource whitelist, `requests <= limits`, and the positivity checks protect the E2B path only. A SandboxClaim with `memory request 1Gi / limit 512Mi` still writes an invalid template, gets rejected by the apiserver, and then times out per W1.
- **Recommendation**: add declarative CEL validation (the repo already uses this pattern at `sandboxclaim_types.go:36,49,110`) restricting keys to cpu/memory and requiring `requests[k] <= limits[k]`, and/or call the shared `validateInplaceUpdateResources` from `buildClaimOptions` through the W1 fail-fast path.

#### W3. Rejecting unsupported resource names is a breaking change; docs and comments are out of sync

- **Location**: `claim.go:118-127`, `claim.go:884-887` (comment), `api/v1alpha1/sandboxclaim_types.go:131-148`
- **Issue**: `ephemeral-storage` / `nvidia.com/gpu` used to be accepted-and-ignored and now fail (E2B path). The CRD field docs still only mention "ignored if the original value is unset", with nothing about the cpu/memory whitelist, and `setContainerResources` still says "Unsupported resource types are silently ignored", which now contradicts the validator (though it remains true on the CRD path — easy to misread).
- **Recommendation**: document the whitelist on the fields and regenerate the CRD, fix the stale comment, and call the incompatibility out in the PR description / release notes.

#### W4. Two classes of invalid input produce two different user-visible outcomes, and the new E2E does not pin the claim outcome

- **Location**: the new memory QoS E2E spec; `common_inplace_update_handler.go:186-200`
- **Issue**: for a QoS-breaking resize the CR template **is** written, the controller rejects the resize, and the CR stays permanently diverged from the pod (exactly the divergence the downscale guard was introduced to prevent) while the claim most likely still succeeds and hands back a sandbox whose resources were never applied. For a downscale the claim times out instead. The new spec asserts the sandbox condition and pod resources but never the claim's terminal phase, so the contract is not pinned.
- **Recommendation**: assert the claim's terminal phase/conditions in both specs, and consider moving the QoS check to the same pre-claim validation point as W1/W2 so that "invalid input is always rejected before claiming" holds uniformly.

#### W5. E2E duplication keeps growing

- **Location**: `test/e2e/sandboxclaim_test.go` (`+982/-19`, six structurally near-identical specs)
- **Recommendation**: still worth a `DescribeTable(policy, resourceName, target, expectRestart, expectReject)` plus `createResizePolicyPool` / `waitWarmSandbox` / `recordMainContainerIdentity` helpers. Every spec currently builds its own SandboxSet with 2–4 minute waits, and suite runtime is growing noticeably.

---

### Suggestions

- **S1**: the downscale rule now exists in two places (`claim.go:892,915` and `inplaceupdate.CheckMemoryDownscale`) and they already differ — infra only touches `Containers[0]` while `CheckMemoryDownscale` covers every container, and the messages differ too. Extract one shared helper to prevent drift.
- **S2**: `terminalMutationError` is reused for a validation error, which conflicts with its own doc ("user-supplied mutation callback"). Consider a dedicated `terminalValidationError`, or a stage string that reads correctly.
- **S3**: the unsupported-name loops in `validateInplaceUpdateResources` still iterate maps, so with several invalid keys the reported one is non-deterministic (the fixed order only covers cpu/memory). Sorting the keys makes the message testable.
- **S4**: the downscale E2E already asserts `AnnotationOwner` is empty (good); also asserting that `LabelSandboxIsClaimed` was not set would pin the "candidate pool not polluted" invariant harder.

---

### Positive findings for this round

- The previous Critical is fixed more thoroughly than suggested: the new test exercises the real create flow and asserts CR resources, quota acquire count, and non-interference with CPU — behavioral, not structural.
- The rejection point is well chosen: it happens in `modifyPickedSandbox`, before `lockPickedSandbox` / admission acquire, so **no quota or lock is leaked**, the candidate is returned via `pickCache.Delete`, and `errors.As` still matches through `fmt.Errorf("%w")` so the retry is correctly terminated.
- Solid unit tests: `TestCheckMemoryDownscale` (nil template, container absent from template, equal values accepted), `TestBuildResourceResizedPod_MemoryDownscaleRejected`, and the new request>limit / unknown-resource cases — all table-driven per repo convention.
- `TestHandleInPlaceUpdateCommon_MemoryDownscaleRejected` matches the style of the adjacent `..._QoSChangeRejected` test, including the correct `SandboxHashImmutablePart` setup.
- `resourceTarget` enum is a clear improvement over `isRequest`, and `clone_test.go:695` now propagates the `SetResources` error instead of discarding it.
- Layering and generated-file discipline hold: `config` is already imported by `clone.go` / `infra.go` in the same package, so no new cross-layer dependency; the CRD diff is description-only and matches the Go comments.

**Verification note**: derived from reading the head above file-by-file and following the call chains; I did not run tests locally. CI on this head is green (unit-tests / sandbox / sandbox-upgrade E2E) — C1 (only the success path is asserted) and C2 (no "higher value for the same key" case) are precisely the blind spots that CI cannot cover.

pkg/controller/sandboxclaim/core/common_control.go
```diff
@@ -144,6 +145,32 @@ func (c *commonControl) EnsureClaimClaiming(ctx context.Context, args ClaimArgs)
 		}
 	}
 
+	// Step 7.1: Validate the inplace update request up front so invalid input
+	// fails fast with an accurate reason instead of spinning until
+	// ClaimTimeout with a misleading NoAvailableSandboxes event.
+	if claim.Spec.InplaceUpdate != nil && claim.Spec.InplaceUpdate.Resources != nil {
+		res := claim.Spec.InplaceUpdate.Resources
+		if err := sandboxcr.ValidateResizeResources(res.Requests, res.Limits); err != nil {
```
furykerry, line 155: this newly introduced code blocks is not covered in UT

pkg/controller/sandboxclaim/core/common_control.go
```diff
@@ -144,6 +145,32 @@ func (c *commonControl) EnsureClaimClaiming(ctx context.Context, args ClaimArgs)
...
+			msg := err.Error()
+			log.Info(msg)
+			c.recorder.Event(claim, "Warning", "InvalidInplaceUpdateResources", msg)
+			TransitionToCompleted(args.NewStatus, "InvalidInplaceUpdateResources", msg)
+			return NoRequeue(), nil
+		}
+		if sandboxSet.Spec.Template != nil {
+			for i := range sandboxSet.Spec.Template.Spec.Containers {
```
furykerry: sandboxset may in the progress of rollout, under such condition, sandbox of old revision may still have valid sandbox for resizing. Consider move the check to preCheckCandidate, and do the downscale check and qos change check

pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -113,6 +107,72 @@ func ValidateAndInitClaimOptions(opts infra.ClaimSandboxOptions) (infra.ClaimSan
...
+// Validation runs in a fixed order so error messages are deterministic
+// regardless of map iteration order. All returned errors are classified as
+// ErrorBadRequest so callers map them to HTTP 400.
+func ValidateResizeResources(requests, limits corev1.ResourceList) error {
+	if len(requests) == 0 && len(limits) == 0 {
+		return managererrors.NewError(managererrors.ErrorBadRequest, "resources must specify at least one of requests or limits")
+	}
+	unknown := unsupportedResourceNames(requests, limits)
+	if len(unknown) > 0 {
+		return managererrors.NewError(managererrors.ErrorBadRequest, "resource %s is not supported for in-place resize", unknown[0])
```
furykerry: consider list all unknow resource in the error message

PR #529, opened 2026-06-11
refactor(sandbox): optimize resume flow by decoupling phase transition from pod readiness

zmberg:
## Summary
- Sandbox Resuming→Running transition no longer waits for Pod Ready; Pod Phase=Running is sufficient
- Initialize (runtime re-init + CSI re-mount) now runs in EnsureSandboxUpdated after Pod Ready=True, with retry on failure
- Condition/event logic for RuntimeInitialized moved into defaultSandboxInitializer, eliminating duplication across call sites
- RuntimeInitialized condition set to Pending at resume entry (calculateStatus), enabling status observability before initialization completes

## Test plan
- [x] `go test ./pkg/controller/sandbox/core/...` all pass
- [x] New `TestCommonControl_EnsureSandboxUpdated_InitializePath` covers: pod not ready (wait), init succeeds (Ready=True), init fails (error), init retry after failure
- [x] Updated `TestCommonControl_EnsureSandboxResumed` tests reflect new behavior (Phase=Running without Ready check)
- [x] Coverage: EnsureSandboxUpdated 92%, syncSandboxStatusFromPod 100%, EnsureSandboxResumed 84%

### DIFF
--- a/pkg/controller/sandbox/core/common_control.go
+++ b/pkg/controller/sandbox/core/common_control.go
@@ -30,6 +30,8 @@ import (
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/distribution/reference"
+
 	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
 	"github.com/openkruise/agents/pkg/agent-runtime/storages"
 	"github.com/openkruise/agents/pkg/utils"
@@ -301,7 +303,7 @@ func isContainersConsistent(pod *corev1.Pod, box *agentsv1alpha1.Sandbox) bool {
 				"container", initContainer.Name)
 			return false
 		}
-		if statusImage != initContainer.Image {
+		if !imageRefsEqual(initContainer.Image, statusImage) {
 			klog.InfoS("init container image mismatch between spec and status, waiting",
 				"sandbox", klog.KObj(box),
 				"container", initContainer.Name,
@@ -311,6 +313,23 @@ func isContainersConsistent(pod *corev1.Pod, box *agentsv1alpha1.Sandbox) bool {
 		}
 	}
 	return true
+}
+
+// imageRefsEqual compares two image references accounting for registry normalization.
+// Container runtimes may expand short names (e.g. "img:latest" → "docker.io/library/img:latest").
+func imageRefsEqual(a, b string) bool {
+	if a == b {
+		return true
+	}
+	return normalizeImageRef(a) == normalizeImageRef(b)
+}
+
+func normalizeImageRef(img string) string {
+	named, err := reference.ParseNormalizedNamed(img)
+	if err != nil {
+		return img
+	}
+	return reference.TagNameOnly(named).String()
 }
 
 // hasUpgradeAction checks if the sandbox has a non-empty upgrade action configured.
--- a/pkg/controller/sandbox/core/common_control.go
+++ b/pkg/controller/sandbox/core/common_control.go
@@ -76,6 +76,7 @@ func NewCommonControl(args SandboxControlArgs) SandboxControl {
 			client:          args.Client,
 			apiReader:       args.APIReader,
 			storageRegistry: storages.NewStorageProvider(),
+			recorder:        args.Recorder,
 		},
 	}
 	return control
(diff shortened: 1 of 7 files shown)


pkg/controller/sandbox/core/common_control.go
```diff
@@ -110,6 +111,24 @@ func (r *commonControl) EnsureSandboxUpdated(ctx context.Context, args EnsureFun
...
+	if initCond != nil && initCond.Reason != agentsv1alpha1.SandboxConditionRuntimeInitReasonSucceeded {
+		pCond := utils.GetPodCondition(&pod.Status, corev1.PodReady)
+		if pCond == nil || pCond.Status != corev1.ConditionTrue {
+			klog.InfoS("Waiting for pod ready before initialization", "sandbox", klog.KObj(box))
+			return nil
+		}
+		if err := r.initializer.Initialize(ctx, box, newStatus); err != nil {
+			return err
+		}
+		rCond := utils.GetSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionReady))
```
furykerry: On the resume path, the Ready condition is not guaranteed to exist in newStatus at this point， consider add a nil guard

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -413,6 +413,14 @@ func (r *SandboxReconciler) calculateStatus(ctx context.Context, args core.Ensur
 				LastTransitionTime: metav1.Now(),
 			}
 			utils.SetSandboxCondition(newStatus, rCond)
+			// Initialize will run in EnsureSandboxUpdated after pod becomes ready
```
furykerry: it seems not necessary to set this condition

pkg/controller/sandbox/core/common_control.go
```diff
@@ -260,40 +261,26 @@ func (r *commonControl) EnsureSandboxResumed(ctx context.Context, args EnsureFun
 			PodUID:   pod.UID,
 		}
 
-		// Delete all Checkpoint CRs after successful resume (pod is running and ready)
 		r.checkpointControl.Cleanup(ctx, box)
 
-		// re-initialize sandbox after resuming or upgrading (includes runtime re-init and CSI storage re-mount)
```
furykerry, line 270: is it possible to leave Initialize logic in ensureSandboxResume, so that the complete resume workflow can be finished in single reconciliation without requeue. Even with Initialization, we can still set sandbox resume condition to true

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -355,7 +355,7 @@ func (r *SandboxReconciler) updateSandboxStatus(ctx context.Context, newStatus a
 	return nil
 }
 
-func (r *SandboxReconciler) calculateStatus(ctx context.Context, args core.EnsureFuncArgs) (*agentsv1alpha1.SandboxStatus, bool) {
+func (r *SandboxReconciler) calculateStatus(args core.EnsureFuncArgs) (*agentsv1alpha1.SandboxStatus, bool) {
```
furykerry: consider keep ctx, and we need change plain klog logging to context sensitive logging

pkg/controller/sandbox/core/common_control.go
zmberg: This isn't achievable at the moment, unless in the future the agent-runtime is separated out as a sidecar container and we only check the sidecar container's readiness for dynamic mount operations. For now, we still have to rely on Pod Ready. There's no way around it.

pkg/controller/sandbox/core/common_control.go
```diff
@@ -110,6 +111,20 @@ func (r *commonControl) EnsureSandboxUpdated(ctx context.Context, args EnsureFun
 		newStatus.Message = "Sandbox Pod Not Found"
 		return nil
 	}
+
+	// If RuntimeInitialized is pending (set during resume), wait for Pod Ready then run Initialize
+	initCond := utils.GetSandboxCondition(newStatus, string(agentsv1alpha1.RuntimeInitialized))
+	if initCond != nil && initCond.Reason != agentsv1alpha1.SandboxConditionRuntimeInitReasonSucceeded {
```
furykerry: why not check `initCond.Status != ConditionTrue` directly, which can be more simple and less rely on fragile condition message.

pkg/controller/sandbox/core/common_control_test.go
```diff
@@ -481,6 +481,175 @@ func TestCommonControl_EnsureSandboxUpdated(t *testing.T) {
...
+				Box:       baseSandbox.DeepCopy(),
+				NewStatus: newStatus,
+			})
+
+			if tt.expectError != "" {
+				if err == nil {
+					t.Fatalf("expected error containing %q, got nil", tt.expectError)
+				}
+				if !contains(err.Error(), tt.expectError) {
+					t.Fatalf("expected error containing %q, got %q", tt.expectError, err.Error())
```
furykerry: The project convention (and sandbox_initializer_test.go in this same PR) uses assert.Contains(t, err.Error(), tt.expectError). Consider using testify consistently.

pkg/controller/sandbox/core/common_control.go
```diff
@@ -260,44 +261,30 @@ func (r *commonControl) EnsureSandboxResumed(ctx context.Context, args EnsureFun
...
 		}
 
-		// Delete all Checkpoint CRs after successful resume (pod is running and ready)
 		r.checkpointControl.Cleanup(ctx, box)
 
 		if !isContainersConsistent(pod, box) {
 			return nil
 		}
 
-		// re-initialize sandbox after resuming or upgrading (includes runtime re-init and CSI storage re-mount)
```
furykerry, line 270: When pod.Status.Phase == corev1.PodRunning but isContainersConsistent(pod, box) returns false, the method returns nil after setting newStatus.Phase = SandboxRunning but before setting the RuntimeInitialized=Pending condition. On subsequent reconciles, the phase is Running, so EnsureSandboxUpdated is called. but the RuntimeInitialized=Pending condition is the sole trigger for EnsureSandboxUpdated to run Initialize. Setting it after the isContainersConsistent gate creates a window where it may never be set.

Consider move the RuntimeInitialized=Pending condition setting before the isContainersConsistent check, so that even if consistency fails on one reconcile, the next reconcile in EnsureSandboxUpdated will still pick it up:

PR #533, opened 2026-06-12
feat(controller): add Commit controller with registry auth and job orchestration

Luckydog691:

Introduces the Commit controller (step 2 of the sandbox-commit feature), which drives the full lifecycle of a `Commit` CR:

- **CommitGate** feature gate (`Commit`, default off) guards controller registration.
- **Reconciler** manages `Pending → Running → Succeeded/Failed` phase transitions and TTL-based auto-deletion.
- **CommonControl** resolves registry credentials from `spec.registryAuth.secrets` (user-specified only) and spawns a Kubernetes `Job` to perform the actual commit.
- **pkg/controller/commit/job** package provides shared Job template construction (`job_template.go`), naming utilities (`utils.go`), and runtime config (`config.go`).

NONE

1. Run unit tests:
   ```bash
   go test ./pkg/controller/commit/...
   ```
2. Enable the feature gate (`--feature-gates=Commit=true`) and create a `Commit` CR targeting a running sandbox pod — verify phase transitions in `.status.phase`.
3. Confirm that a `Job` is created in the same namespace with the correct image and env vars.
4. After the Job completes, verify the `Commit` reaches `Succeeded` and is auto-deleted when TTL expires.

- `pkg/controller/commit/job` is intentionally kept minimal here — only the template/config pieces needed by the controller. The job-runner binary (nerdctl, registry push logic) lands in step 3.
- Registry auth: only `spec.registryAuth.secrets` is used. No fallback to namespace keyring or SA imagePullSecrets. If nil, the commit attempts an anonymous push.
- Job creation uses `GenerateName`; the controller looks up Jobs by `LabelCommitUID` field index (not by name).

---

- `config/rbac/role.yaml` — generated RBAC rules for commits, jobs, secrets, pods, events (controller-gen output)
- `pkg/utils/constant.go` — adds `CommitAnnotationModeKey` constant used by the controller


pkg/controller/commit/commit_controller.go
```diff
@@ -0,0 +1,266 @@
...
+	start := time.Now()
+	defer func() {
+		klog.V(4).InfoS("Reconcile finished", "name", req.Name, "namespace", req.Namespace, "elapsed", time.Since(start), "err", err)
+	}()
+
+	// Fetch target pod directly from API server to bypass informer cache filtering.
+	// The Pod informer may be scoped by label selector (CachePodLabelSelector),
+	// but Commit can target any Pod, not just Sandbox-managed ones.
+	pod := &corev1.Pod{}
+	if err = r.APIReader.Get(ctx, client.ObjectKey{Namespace: commit.Namespace, Name: commit.Spec.PodName}, pod); err != nil {
```
furykerry: is it possible to use informer cache instead ?

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,362 @@
...
+
+func (r *commonControl) EnsureCommitDeleted(ctx context.Context, args *EnsureFuncArgs) (time.Duration, error) {
+	commit := args.Commit
+
+	_, err := utils.PatchFinalizer(ctx, r.Client, commit, utils.RemoveFinalizerOpType, agentsv1alpha1.CommitFinalizer)
+	if err != nil {
+		klog.ErrorS(err, "remove commit finalizer failed", "commit", klog.KObj(commit))
+		return 0, err
+	}
+	klog.InfoS("remove commit finalizer success", "commit", klog.KObj(commit))
```
furykerry: what finalize operation is required for commit? clean commit job?

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,362 @@
...
+		}
+	}
+
+	klog.InfoS("No registry secret found, commit will attempt anonymous push", "commit", klog.KObj(commit))
+	return ""
+}
+
+// resolveRegistrySecretByKeyring lists all dockerconfigjson secrets in the Commit's namespace,
+// parses their auth configs, and returns the name of the first secret that contains
+// credentials matching the target registry host.
```
furykerry: credentials matching the target registry host does not necessarily means a push credentials, can we just rely on user specified secrets?

pkg/job/job_template.go
```diff
@@ -0,0 +1,235 @@
...
+	// Map CommitSpec.TimeoutSeconds to Job's ActiveDeadlineSeconds
+	var activeDeadlineSeconds *int64
+	if g.Commit.Spec.TimeoutSeconds > 0 {
+		deadline := int64(g.Commit.Spec.TimeoutSeconds)
+		activeDeadlineSeconds = &deadline
+	}
+
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      MakeJobName(string(g.Commit.UID)),
```
furykerry: just use generateName?

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,362 @@
...
+			args.NewStatus.CompletionTime = &now
+			return 0, nil
+		}
+		return 0, fmt.Errorf("failed to get job: %w", err)
+	}
+
+	done, success := jobutil.IsJobCompleted(job)
+	if !done {
+		klog.InfoS("Job still running, will requeue", "job", klog.KObj(job), "commit", klog.KObj(commit))
+		return 30 * time.Second, nil
```
furykerry: plz replace constant requeue duration with constant variable

pkg/controller/commit/commit_event_handler.go
```diff
@@ -0,0 +1,82 @@
...
+			Name:      commitName,
+		},
+	})
+}
+
+func (e *enqueueRequestForJob) Create(_ context.Context, evt event.TypedCreateEvent[client.Object], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	e.addEvent(q, evt.Object)
+}
+
+func (p *enqueueRequestForJob) Delete(_ context.Context, evt event.TypedDeleteEvent[client.Object], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
```
furykerry: shall we enqueue for job delete event?

pkg/controller/commit/commit_event_handler.go
```diff
@@ -0,0 +1,82 @@
...
+	commitName, ok := job.Labels[jobutil.LabelCommitName]
+	if !ok {
+		return
+	}
+
+	complete, success := jobutil.IsJobCompleted(job)
+	if !complete {
+		return
+	}
+	klog.InfoS("job complete add commit job event", "commitName", commitName, "success", success, "job", klog.KObj(job))
```
furykerry: avoid logging in event handler, blocking io will block event handler thread

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,362 @@
...
+	g := &jobutil.JobGenerator{Commit: commit, Pod: pod}
+
+	// Resolve registry auth secret (same namespace, mounted directly by name)
+	g.DockerConfigSecretName = r.resolveRegistrySecretName(ctx, commit, pod)
+
+	job, err := g.GenerateCommitJob()
+	if err != nil {
+		return fmt.Errorf("failed to generate commit job: %v", err)
+	}
+	if err := r.Client.Create(ctx, job); err != nil {
```
furykerry: plz use expectation to avoid duplicated reconcile due to stale cache

pkg/controller/commit/commit_controller.go
```diff
@@ -0,0 +1,266 @@
...
+// +kubebuilder:rbac:groups=agents.kruise.io,resources=commits/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=agents.kruise.io,resources=commits/finalizers,verbs=update
+// +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list
+// +kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=events,verbs=create
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=create;delete;get;watch;list
+
+func (r *CommitReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	klog.InfoS("Reconcile", "name", req.Name, "namespace", req.Namespace)
```
furykerry: plz use `logf.FromContext(ctx)` (controller-runtime/pkg/log) instead of klog in controller

pkg/controller/commit/job/config.go
```diff
@@ -0,0 +1,94 @@
+/*
```
furykerry, line 1: plz rename job to more specific name such as commitjob

pkg/controller/commit/commit_controller.go
Luckydog691: Will switch to `r.Get` (informer cache). Commit only targets Sandbox pods, which always carry the `created-by` label and are guaranteed in the cache even with `CachePodLabelSelector` enabled — same as the sandbox controller does.

pkg/controller/commit/commit_controller.go
Luckydog691: Will do. Will replace all `klog` with `logf.FromContext(ctx)` in the controller layer.

pkg/controller/commit/core/common_control.go
Luckydog691: No explicit cleanup needed. The Job has an `OwnerReference` (Controller=true) pointing to the Commit, so Kubernetes GC handles cascade deletion. Removing the finalizer is sufficient.

pkg/controller/commit/core/common_control.go
Luckydog691: Will do. Will extract `30 * time.Second` to a named constant.

pkg/controller/commit/core/common_control.go
Luckydog691: Will do. Will use `ScaleExpectations` aligned with the sandbox controller pattern: `ExpectScale` after Job create, `ObserveScale` in the event handler, check `SatisfiedExpectations` before re-creating.

pkg/controller/commit/commit_event_handler.go
Luckydog691: Yes. Will add enqueue logic in the `Delete` handler, same as the checkpoint event handler does — extract Commit name from Job labels and add to the workqueue.

pkg/controller/commit/commit_event_handler.go
Luckydog691: Will do. Will remove `klog` calls from the event handler.

pkg/job/job_template.go
Luckydog691: Will do. Will switch to `GenerateName` and write the generated Job name back to the Commit annotation so the controller can look it up later.

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,392 @@
...
+}
+
+func (r *commonControl) listCRJobPods(ctx context.Context, commit *agentsv1alpha1.Commit) (*corev1.PodList, error) {
+	jobPods := &corev1.PodList{}
+	// Filter by the commit-uid label set on the Job pod template. This works
+	// with GenerateName (where the Job name is not deterministic) and avoids
+	// relying on the batch.kubernetes.io/job-name label which requires knowing
+	// the generated name.
+	matchingLabels := client.MatchingLabels{
+		jobutil.LabelCommitUID: string(commit.UID),
```
furykerry: consider add an index to commit id label

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,392 @@
...
+
+func (r *commonControl) EnsureCommitUpdated(ctx context.Context, args *EnsureFuncArgs) (time.Duration, error) {
+	log := log.FromContext(ctx)
+	commit := args.Commit
+	log.Info("EnsureCommitUpdated", "commit", klog.KObj(commit), "commitID", commit.Status.CommitID)
+
+	// Read the generated Job name from annotation; fall back to MakeJobName for backward compatibility.
+	jobName := commit.Annotations[utils.CommitAnnotationJobNameKey]
+	if jobName == "" {
+		jobName = jobutil.MakeJobName(string(commit.UID))
```
furykerry: Issue: The fallback jobName = jobutil.MakeJobName(string(commit.UID)) produces a deterministic name like agent-job-abc123, but GenerateName on the Job spec appends a random suffix (e.g., agent-job-abc123-xyz9z). The Get-by-name will never match the actual Job.
Consider: Replace the single-name Get with a List filtered by LabelCommitUID, matching the approach already used in listCRJobPods. Alternatively, use the existing listCRJobPods to find the Job pod and then look up the owning Job via ownerReferences:

pkg/controller/commit/commit_event_handler.go
```diff
@@ -0,0 +1,102 @@
...
+	if !ok {
+		return
+	}
+	commitName, ok := job.Labels[jobutil.LabelCommitName]
+	if !ok {
+		return
+	}
+	// Clean up any stale Create expectation to prevent expectation leaks.
+	commitKey := types.NamespacedName{Namespace: job.Namespace, Name: commitName}.String()
+	core.ScaleExpectations.ObserveScale(commitKey, expectations.Create, job.Name)
```
furykerry: Issue: The Delete handler calls ObserveScale(commitKey, expectations.Create, job.Name), which observes a Create event. This is semantically wrong — the Job was deleted, not created. This call will not correctly satisfy or clean up the existing Create expectation.

Fix: Use DeleteExpectations to clean up, or simply remove the ObserveScale call from the Delete handler since the reconcile loop already handles timeouts:

pkg/controller/commit/commit_controller_test.go
```diff
@@ -0,0 +1,577 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
+	"github.com/openkruise/agents/pkg/controller/commit/core"
+	"github.com/openkruise/agents/pkg/utils"
+)
+
+func newTestScheme() *runtime.Scheme {
+	scheme := runtime.NewScheme()
```
furykerry, line 43: Issue: _ = batchv1.AddToScheme(scheme) is called explicitly, but clientgoscheme.AddToScheme already registers batchv1. This is inconsistent with common_control_test.go which correctly omits it.
Fix: Remove the explicit batchv1.AddToScheme call (and the batchv1 import if unused) to align with the other test file.

pkg/job/utils_test.go
```diff
@@ -0,0 +1,206 @@
...
+		t.Errorf("expected Status=False for non-zero exit code, got %s", cond.Status)
+	}
+	if cond.Type != "CommitContainer" || cond.Reason != "CommitContainerFailed" {
+		t.Errorf("unexpected condition: %+v", cond)
+	}
+}
+
+// TestGetCommitCondition_UnknownExitCode covers the early-return branch when the
+// container's terminated exit code is not registered in CommitJobExitCodeMap.
+func TestGetCommitCondition_UnknownExitCode(t *testing.T) {
```
furykerry: TestGetCommitCondition_*, TestEnvConfig_* should be consolidated into table-driven form.

pkg/job/utils.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	return false, false
+}
+
+type CommitConditionValue struct {
+	conditionType   string
+	conditionReason string
+}
+
+var CommitJobExitCodeMap = map[int32]CommitConditionValue{
```
furykerry: Issue: CommitConditionValue is exported but its fields (conditionType, conditionReason) are unexported. External packages can reference the type but never construct or inspect it. This makes CommitJobExitCodeMap difficult to use outside the job package.
Fix: Either export the fields (ConditionType, ConditionReason) or make the struct unexported (commitConditionValue). If the map is intended for external use, the fields should be exported.

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+
+	jobPods, err := r.listCRJobPods(ctx, commit)
+	if err != nil {
+		return 0, fmt.Errorf("failed to list commit job pods: %w", err)
+	}
+	if len(jobPods.Items) > 0 {
+		log.Info("commit job pod already exists, transitioning to Running", "commit", klog.KObj(commit))
+		now := metav1.Now()
+		args.NewStatus.StartTime = &now
```
furykerry: In common_control.go EnsureCommitRunning, when a pre-existing Job pod is found, args.NewStatus.StartTime = &now overwrites any previously set start time. If the controller re-enters the Pending handler (e.g., after a transient error), the start time shifts forward on each attempt.
Consider preserving the existing StartTime from commit.Status.StartTime if already set.

pkg/controller/commit/core/common_control.go
Luckydog691: Will add a field index for LabelCommitUID on Pods and Jobs to speed up List queries.

pkg/controller/commit/core/common_control.go
Luckydog691: Good catch. Will replace Get-by-name with List-by-LabelCommitUID, same as listCRJobPods.

pkg/controller/commit/commit_event_handler.go
Luckydog691: Will remove the ObserveScale call from Delete handler. The reconcile timeout already handles stale expectations.

pkg/controller/commit/commit_controller_test.go
Luckydog691: Will remove the redundant batchv1.AddToScheme call.

pkg/job/utils_test.go
Luckydog691: Will do. Will consolidate into table-driven form.

pkg/job/utils.go
Luckydog691: Will make the struct unexported since it is only used internally.

pkg/controller/commit/core/common_control.go
Luckydog691: Will preserve existing StartTime from commit.Status if already set.

pkg/controller/commit/core/common_control.go
Luckydog691: Will remove Tier 2/3 fallback and only rely on user-specified spec.registryAuth.secrets.

pkg/controller/commit/job/config.go
Luckydog691: Will move pkg/job to pkg/controller/commit/job since only the commit controller uses it.

pkg/job/job_template.go
Luckydog691: Update: instead of writing the Job name to an annotation, we now look up Jobs by LabelCommitUID field index (List + MatchingFields). This is simpler and avoids maintaining a separate annotation. The annotation approach was removed.

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	now := metav1.Now()
+	if err = r.applyCommitJob(ctx, commit, pod); err != nil {
+		log.Error(err, "EnsureCommitRunning failed", "commit", klog.KObj(commit))
+		r.Recorder.Eventf(commit, corev1.EventTypeWarning, "JobCreationFailed", "Failed to create commit job: %v", err)
+		args.NewStatus.StartTime = &now
+		args.NewStatus.Phase = agentsv1alpha1.CommitPhaseFailed
+		args.NewStatus.CompletionTime = &now
+		return 0, nil
```
furykerry, line 98: is it possible to return error so that the controller can backoff and retry later

pkg/controller/commit/commit_event_handler.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	commitName, ok := job.Labels[jobutil.LabelCommitName]
+	if !ok {
+		return
+	}
+
+	// Observe the Create expectation so the controller knows the Job is in the cache.
+	commitKey := types.NamespacedName{Namespace: job.Namespace, Name: commitName}.String()
+	core.ScaleExpectations.ObserveScale(commitKey, expectations.Create, job.Name)
```
furykerry: avoid observe in event handler which has lock operation , observe in the reconcile loop

pkg/controller/commit/commit_event_handler.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+var _ handler.TypedEventHandler[client.Object, reconcile.Request] = &enqueueRequestForJob{}
+
+// enqueueRequestForJob enqueues a Commit reconcile request only when the underlying
+// commit Job reaches a terminal state (Complete or Failed). Intermediate Job/Pod
+// status changes are filtered out to avoid unnecessary reconciles.
+type enqueueRequestForJob struct{}
+
+func (p *enqueueRequestForJob) addEvent(q workqueue.TypedRateLimitingInterface[reconcile.Request], obj runtime.Object) {
```
furykerry: plz add ut for this func

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,261 @@
...
+	if err := r.Client.List(ctx, jobList, client.InNamespace(commit.Namespace), client.MatchingFields{jobutil.IndexFieldCommitUID: string(commit.UID)}); err != nil {
+		return 0, fmt.Errorf("failed to list commit jobs: %w", err)
+	}
+	if len(jobList.Items) > 0 {
+		log.Info("commit job already exists, transitioning to Running", "commit", klog.KObj(commit))
+		setCommitRunning(args.NewStatus, commit)
+		return 0, nil
+	}
+
+	// Fallback: check for orphan job pods (e.g., Job was deleted but pod remains).
```
furykerry: is it necessary ? job deletion will trigger cascade deletion of job pods, so in normal cases, no job pods will exists, is it worth the extra list operation for possible abnormal manual orphan deletion?

pkg/controller/commit/commit_controller.go
```diff
@@ -0,0 +1,288 @@
...
+	log := log.FromContext(ctx)
+	commit := args.Commit
+	requeueAfter, err := ensureFunc(ctx, args)
+	if err != nil {
+		if retErr := r.updateCommitStatus(ctx, *args.NewStatus, commit); retErr != nil {
+			log.Error(retErr, "Failed to update commit status on error", "commit", klog.KObj(commit))
+		}
+		return ctrl.Result{}, err
+	}
+	return ctrl.Result{RequeueAfter: requeueAfter}, r.updateCommitStatus(ctx, *args.NewStatus, commit)
```
furykerry: plz avoid return RequeueAfter and error together, controller-runtime will drop requeueAfter if error is not nil, so it is better to explicitly check the error return by updateCommitStatus, and decide to use default backoff or requeueAfter

pkg/controller/commit/job/utils.go
```diff
@@ -0,0 +1,115 @@
...
+				"type", conditionValue.conditionType,
+				"reason", conditionValue.conditionReason)
+			status := metav1.ConditionTrue
+			if cs.State.Terminated.ExitCode != 0 {
+				status = metav1.ConditionFalse
+			}
+			cond := &metav1.Condition{
+				Type:               conditionValue.conditionType,
+				Status:             status,
+				Reason:             conditionValue.conditionReason,
```
furykerry, line 114: plz also copy the condition message

pkg/controller/commit/commit_controller.go
```diff
@@ -0,0 +1,288 @@
...
+// +kubebuilder:rbac:groups=core,resources=events,verbs=create
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=create;delete;get;watch;list
+
+func (r *CommitReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	log := log.FromContext(ctx)
+	log.Info("Reconcile", "name", req.Name, "namespace", req.Namespace)
+	commit := &agentsv1alpha1.Commit{}
+	if err = r.Get(ctx, req.NamespacedName, commit); err != nil {
+		if errors.IsNotFound(err) {
+			log.V(4).Info("Commit not found", "name", req.Name, "namespace", req.Namespace)
```
furykerry, line 117: plz delete commit ScaleExpectations

pkg/controller/commit/commit_controller.go
```diff
@@ -0,0 +1,288 @@
...
+	// Get control implementation
+	control, err := r.getControl(commit)
+	if err != nil {
+		log.Error(err, "get control failed", "commit", klog.KObj(commit))
+		return ctrl.Result{}, err
+	}
+
+	// Handle deletion
+	if !commit.DeletionTimestamp.IsZero() {
+		args := &core.EnsureFuncArgs{Pod: pod, Commit: commit, NewStatus: newStatus}
```
furykerry, line 152: plz delete commit ScaleExpectations

pkg/controller/commit/job/job_template.go
```diff
@@ -0,0 +1,235 @@
...
+	// Map CommitSpec.TimeoutSeconds to Job's ActiveDeadlineSeconds
+	var activeDeadlineSeconds *int64
+	if g.Commit.Spec.TimeoutSeconds > 0 {
+		deadline := int64(g.Commit.Spec.TimeoutSeconds)
+		activeDeadlineSeconds = &deadline
+	}
+
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: MakeJobName(string(g.Commit.UID)),
```
furykerry: just use commit name (with some length limit) should be enough for generate name

pkg/controller/commit/job/config.go
```diff
@@ -0,0 +1,94 @@
...
+	// EnvAgentJobImagePullPolicy is the environment variable name for the agent job image pull policy.
+	EnvAgentJobImagePullPolicy = "AGENT_JOB_IMAGE_PULL_POLICY"
+)
+
+// EnvConfig reads configuration from environment variables.
+type EnvConfig struct{}
+
+func (c *EnvConfig) ContainerID() string   { return os.Getenv(EnvContainerID) }
+func (c *EnvConfig) CommitImage() string   { return os.Getenv(EnvCommitImage) }
+func (c *EnvConfig) AgentJobImage() string { return os.Getenv(EnvAgentJobImage) }
```
furykerry: is it possible to pass the commit job image through command line arguments ?

pkg/controller/commit/commit_event_handler.go
```diff
@@ -0,0 +1,96 @@
...
+	job, ok := obj.(*batchv1.Job)
+	if !ok {
+		return
+	}
+
+	commitName, ok := job.Labels[jobutil.LabelCommitName]
```
furykerry: use ownerreference as the commitName ?

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,261 @@
...
+	if len(jobList.Items) == 0 {
+		log.Info("Job not found, marking commit as failed", "commit", klog.KObj(commit))
+		r.Recorder.Eventf(commit, corev1.EventTypeWarning, "JobNotFound", "Commit job not found for commit %s", commit.Name)
+		now := metav1.Now()
+		args.NewStatus.Phase = agentsv1alpha1.CommitPhaseFailed
+		args.NewStatus.CompletionTime = &now
+		return 0, nil
+	}
+	if len(jobList.Items) > 1 {
+		log.Info("Multiple jobs found for commit, using the latest one", "commit", klog.KObj(commit), "count", len(jobList.Items))
```
furykerry, line 147: plz add ut for multiple jobs per commit case

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,240 @@
...
+	log.Info("Job completed", "job", klog.KObj(job), "commit", klog.KObj(commit), "success", success)
+	phase := agentsv1alpha1.CommitPhaseSucceeded
+	if !success {
+		phase = agentsv1alpha1.CommitPhaseFailed
+	}
+	if condition := r.getLatestJobPodExitCode(ctx, commit); condition != nil {
+		args.NewStatus.Conditions = append(args.NewStatus.Conditions, *condition)
+	}
+	args.NewStatus.Phase = phase
+	now := metav1.NewTime(time.Now())
```
furykerry: consider using metav1.Now() ， metav1.Now() truncates to second precision (as K8s API does), while metav1.NewTime(time.Now()) preserves nanosecond precision in the local object (though it's truncated on serialization).

pkg/controller/commit/commit_controller.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func (r *CommitReconciler) handleCommitDelete(ctx context.Context, args *core.EnsureFuncArgs, control core.CommitControl) (ctrl.Result, error) {
+	log := log.FromContext(ctx)
+	commit := args.Commit
+	log.V(3).Info("handleCommitDelete", "commit", klog.KObj(commit), "commitID", commit.Status.CommitID, "uid", commit.UID)
+	requeueAfter, err := control.EnsureCommitDeleted(ctx, args)
+	if err != nil {
+		log.Error(err, "handleCommitDelete failed", "commit", klog.KObj(commit))
+		return ctrl.Result{RequeueAfter: requeueAfter}, err
```
furykerry: Returns RequeueAfter alongside error， consider return ctrl.Result{}, err when err != nil.

pkg/controller/commit/core/common_control.go
```diff
@@ -0,0 +1,240 @@
...
+		args.NewStatus.StartTime = &now
+		args.NewStatus.Phase = agentsv1alpha1.CommitPhaseFailed
+		args.NewStatus.CompletionTime = &now
+		return 0, nil
+	}
+
+	// Create Job — transient errors return error for backoff retry, Commit stays Pending.
+	if err := r.Client.Create(ctx, job); err != nil {
+		if errors.IsAlreadyExists(err) {
+			log.Info("Job already exists, ignore", "job", klog.KObj(job), "commit", klog.KObj(commit))
```
furykerry, line 104: When Create returns IsAlreadyExists, the code logs "Job already exists, ignore" but then falls through to log "created Job" and sets a ScaleExpectations.ExpectScale() on a job that was not actually created by this reconcile. This records a false expectation.

Consider return early after the IsAlreadyExists branch, or at minimum don't log "created Job" and don't set the expectation:

pkg/controller/commit/commit_event_handler.go
```diff
@@ -0,0 +1,102 @@
...
+			Name:      commitName,
+		},
+	})
+}
+
+func (e *enqueueRequestForJob) Create(_ context.Context, evt event.TypedCreateEvent[client.Object], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	e.addEvent(q, evt.Object)
+}
+
+func (p *enqueueRequestForJob) Delete(_ context.Context, evt event.TypedDeleteEvent[client.Object], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
```
furykerry: Receiver names alternate between p (in addEvent, Delete, Generic, Update) and e (in Create). This is a style inconsistency.

pkg/controller/commit/job/utils.go
```diff
@@ -0,0 +1,120 @@
...
+		if cs.Name != AgentJobContainerName {
+			continue
+		}
+		if cs.State.Terminated != nil {
+			conditionValue, ok := commitJobExitCodeMap[cs.State.Terminated.ExitCode]
+			if !ok {
+				klog.InfoS("Unknown exit code, skipping condition", "containerID", cs.ContainerID, "exitCode", cs.State.Terminated.ExitCode)
+				return nil
+			}
+			klog.InfoS("Commit job container terminated",
```
furykerry: The job package uses klog.InfoS directly, while the controllers use log.FromContext(ctx). The job utils package is imported by the controller layer and should follow the controller logging conventions.

furykerry (commented):
/lgtm

PR #541, opened 2026-06-15
update envd-run.sh script

BH4AWS:

### DIFF
--- a/pkg/agent-runtime/envd-run.sh
+++ b/pkg/agent-runtime/envd-run.sh
@@ -2,24 +2,39 @@
 set -e  # Enable exit on error for the entire script
 
 # Parse arguments. Supported flags:
-#   --nolog   Suppress envd stdout/stderr (do not redirect to /proc/1/fd/1).
-#             Default behaviour pipes envd output to PID 1's stdout so that
-#             container log collectors pick it up.
-#   --help    Show usage information and exit.
+#   --nolog              Suppress envd stdout/stderr (do not redirect to /proc/1/fd/1).
+#                        Default behaviour pipes envd output to PID 1's stdout so that
+#                        container log collectors pick it up.
+#   -- <cmd> [args...]   Everything after "--" is collected as a command array and
+#                        executed directly (no shell interpretation). This follows the
+#                        kubectl exec pattern and avoids double-interpretation pitfalls.
+#                        Examples:
+#                        /bin/bash envd-run.sh -- /path/to/script arg1 arg2
+#                        /bin/bash envd-run.sh -- echo hello world
+#   --help               Show usage information and exit.
 NOLOG=false
-for arg in "$@"; do
-    case "$arg" in
+USER_INIT_ARGS=()
+while [ $# -gt 0 ]; do
+    case "$1" in
         --nolog)
             NOLOG=true
+            shift
+            ;;
+        --)
+            # Everything after -- becomes the init command (no shell interpretation)
+            shift
+            USER_INIT_ARGS=("$@")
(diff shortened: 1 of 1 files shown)


pkg/agent-runtime/envd-run.sh
```diff
@@ -2,24 +2,41 @@
...
 
 # Parse arguments. Supported flags:
-#   --nolog   Suppress envd stdout/stderr (do not redirect to /proc/1/fd/1).
-#             Default behaviour pipes envd output to PID 1's stdout so that
-#             container log collectors pick it up.
-#   --help    Show usage information and exit.
+#   --nolog              Suppress envd stdout/stderr (do not redirect to /proc/1/fd/1).
+#                        Default behaviour pipes envd output to PID 1's stdout so that
+#                        container log collectors pick it up.
+#   --user-init <arg>    User-provided init script path or inline command string.
```
furykerry: When --user-init is not an executable file, it's passed verbatim to sh -c, which subjects the string to a second round of shell interpretation. 

Consider follow the kubectl exec pattern — use -- separator and collect remaining args as an array, then exec them directly:

```
# Instead of taking a single string, support:
#   envd-run.sh -- /path/to/script arg1 arg2
#   envd-run.sh -- echo hello world
USER_INIT_ARGS=()
while [ $# -gt 0 ]; do
    case "$1" in
        --nolog)
            NOLOG=true
            shift
            ;;
        --)
            # Everything after -- becomes the init command
            shift
            USER_INIT_ARGS=("$@")
            shift $#
            ;;
        ...
    esac
done

# Execute: direct exec, no shell interpretation
if [ ${#USER_INIT_ARGS[@]} -gt 0 ]; then
        # If it's not a file, exec it directly — let the OS find it in PATH
        # DO NOT use sh -c to avoid double-interpretation
        "${USER_INIT_ARGS[@]}" || echo "Warning: --user-init exited with code $?, continuing..." >&2
fi
```

furykerry (commented):
plz change the commit message and describe what changes are made to envd-run.sh, current commit message conveys no useful message

pkg/agent-runtime/envd-run.sh
BH4AWS: fixed

pkg/agent-runtime/envd-run.sh
```diff
@@ -2,24 +2,55 @@
...
```
furykerry: if we have -- <cmd> , --user-init options sees redundant ,  consider remove them completely

BH4AWS: fixed

pkg/agent-runtime/envd-run.sh
```diff
@@ -180,3 +195,24 @@ if [ "$NOLOG" = "true" ]; then
...
+#       exec:
+#         command: ["/bin/bash", "envd-run.sh", "--", "/mnt/init/setup.sh", "arg1", "arg2"]
+#   or:
+#         command: ["/bin/bash", "envd-run.sh", "--", "echo", "hello", "world"]
+#
+# A non-zero exit code is logged as a warning but does NOT abort startup.
+# ---------------------------------------------------------------------------
+if [ ${#USER_INIT_ARGS[@]} -gt 0 ]; then
+    echo "[user-init] Running direct exec: ${USER_INIT_ARGS[*]}"
+    "${USER_INIT_ARGS[@]}" || echo "Warning: -- command exited with code $?, continuing..." >&2
```
furykerry: we shall not continue if user specified postStarthooks fails

PR #552, opened 2026-06-18
feat(identity): extend InjectAllCAIntoContainers to cover InitContainers

BH4AWS:
Previously InjectAllCAIntoContainers only iterated pod.Spec.Containers, which meant sidecars placed in InitContainers (e.g. csi-agent-sidecar injected by setCSIMountContainer) never received CA VolumeMount or EnvVar entries even when a CABundleSpec targeted them by name.

This commit extends the function to also iterate InitContainers. A new constant initContainerIndexBase (1<<16) is used as an index offset so that position-based selectors (OnlyMainContainer, MainContainerOrByName) that check idx==0 correctly skip init containers, while name-based selectors (ByContainerName, AllContainers) remain unaffected.

Covered by 5 new table-driven test cases validating:
- ByContainerName targets init container by name
- AllContainers covers both init and regular containers
- OnlyMainContainer skips all init containers
- Only init containers present (edge case)
- Idempotent skip when mount already exists

### DIFF
--- a/pkg/identity/ca_cert_injector.go
+++ b/pkg/identity/ca_cert_injector.go
@@ -80,14 +80,15 @@ const (
 // the override semantics.
 func init() {
 	RegisterCABundleSpec(CABundleSpec{
-		Name:              GatewayCABundleName,
-		SecretName:        GatewayCASecretName,
-		SecretDataKey:     GatewayCAKey,
-		VolumeName:        gatewayCAVolumeName,
-		MountPath:         gatewayCAMountPath,
-		SubPath:           GatewayCAKey,
-		ReadOnly:          true,
-		ContainerSelector: OnlyMainContainer(),
+		Name:                  GatewayCABundleName,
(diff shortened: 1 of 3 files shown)


pkg/identity/ca_cert_injector.go
```diff
@@ -95,6 +95,13 @@ func init() {
 	})
 }
 
+// initContainerIndexBase is added to the positional index of init containers
+// when evaluating ContainerSelector. This ensures that index-based selectors
+// such as OnlyMainContainer() (which checks idx == 0) never accidentally match
+// an init container. Name-based selectors like ByContainerName() ignore the
+// index altogether and remain unaffected.
+const initContainerIndexBase = 1 << 16
```
furykerry: plz extend CABundleSpec with  InitContainerSelector field, and do not hack with initContainerIndexBase

BH4AWS: done

PR #554, opened 2026-06-18
fix(features): disable CachePodLabelSelector and SandboxUpgradeResumeFromFailedStep by default

zmberg:
## Summary
- Disable `CachePodLabelSelector` feature gate by default (was enabled)
- Disable `SandboxUpgradeResumeFromFailedStep` feature gate by default (was enabled)

These two features are not yet stable enough for default-on behavior and should require explicit opt-in.

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -70,10 +70,10 @@ var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	SandboxClaimGate:                       {Default: true, PreRelease: featuregate.Alpha},
 	SandboxCreatePodRateLimitGate:          {Default: false, PreRelease: featuregate.Alpha},
 	SandboxCreatePodInjectConfigGate:       {Default: false, PreRelease: featuregate.Alpha},
-	CachePodLabelSelectorGate:              {Default: true, PreRelease: featuregate.Alpha},
+	CachePodLabelSelectorGate:              {Default: false, PreRelease: featuregate.Alpha},
 	SandboxInPlaceResourceResizeGate:       {Default: true, PreRelease: featuregate.Alpha},
 	SandboxMultiClusterNaming:              {Default: false, PreRelease: featuregate.Alpha},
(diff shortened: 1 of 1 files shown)


HARSHRAJ2789:
Ran both flips against current master (`fbadb53`) as well as the branch base, since this has been
open a while: `./pkg/controller/sandbox/...` passes with them applied, and the PR still merges
cleanly even though `2b1118d` is ten weeks back. `CachePodLabelSelector` reads fine to me. Its two
consumers at `pod_event_handler.go:75` and `main.go:286` both have a working disabled path and
nothing downstream reads the gate.

One consequence of flipping `SandboxUpgradeResumeFromFailedStep` is worth knowing before it lands.

`TestDetermineUpgradeResumeReason` opens with a comment saying it enables the gate so the switch
statement is exercised, then calls `Set("SandboxUpgradeResumeFromFailedStepGate=true")` at
`sandbox_controller_test.go:6261`. That string is not a registered gate name: `features.go:56`
registers the constant under `"SandboxUpgradeResumeFromFailedStep"`, with no `Gate` suffix. The
call returns an error and `_ =` drops it. Probing the same call on both defaults:

```
master     default=true    Set err = unrecognized feature gate: SandboxUpgradeResumeFromFailedStepGate    enabled after = true
this PR    default=false   Set err = unrecognized feature gate: SandboxUpgradeResumeFromFailedStepGate    enabled after = false
```

Today the test reaches the enabled path because the default carries it there. After this PR it
reaches the disabled path while the comment still says otherwise, and it passes in both cases, so
nothing reports the swap.

The suffix is easy to get wrong here, since `features.go` is not consistent about it:
`SandboxCreatePodRateLimitGate` at `:37` and `SandboxCreatePodInjectConfigGate` at `:40` really are
registered with `Gate` on the end, while `:44` and `:56` are not.

Two smaller ones nearby, both about restoring a default this PR changes. `:5946` sets the gate
false for its test and puts `=true` back in the defer, and `:5828` sets `=true` then restores
`=true`. Both write the old default back, so once this lands they hand the next test in the
package an enabled gate.

PR #555, opened 2026-06-18
feat(sidecarutils): merge postStart hooks using -- separator instead of skipping

BH4AWS:
…of skipping

When the user's main container already has a postStart Exec hook, the previous behavior was to skip injection entirely. This change merges the injected envd-run.sh command with the user's command using the "--" separator pattern, following the kubectl exec convention.

The merged result preserves each user command element as an individual argument (no shell interpretation), e.g.:
  injected: ["bash", "/mnt/envd/envd-run.sh"]
  user:     ["/path/to/script", "arg1", "arg2"]
  merged:   ["bash", "/mnt/envd/envd-run.sh", "--", "/path/to/script", "arg1", "arg2"]

For non-Exec handlers (HTTPGet/TCPSocket), the existing hook is preserved and injection is skipped since they cannot be merged.

- Add mergePostStartExecCommands helper with "--" separator logic
- Update setMainContainerConfigWhenInjectRuntimeSidecar to call merge
- Add TestMergePostStartExecCommands covering 8 boundary scenarios

### DIFF
--- a/pkg/utils/sidecarutils/sidecar_config_inject.go
+++ b/pkg/utils/sidecarutils/sidecar_config_inject.go
@@ -237,19 +237,30 @@ func setMainContainerConfigWhenInjectRuntimeSidecar(ctx context.Context, mainCon
 		config.MainContainer.Lifecycle.PostStart != nil &&
 		hasValidLifecycleHandler(config.MainContainer.Lifecycle.PostStart)
 
-	if mainContainerHasValidPostStart {
-		if configHasValidPostStart {
-			log.V(utils.DebugLogLevel).Info("conflicting postStart hooks detected, main container already has a postStart hook defined",
-				"existingHook", mainContainer.Lifecycle.PostStart,
-				"injectedHook", config.MainContainer.Lifecycle.PostStart)
+	if configHasValidPostStart {
+		if mainContainer.Lifecycle == nil {
+			mainContainer.Lifecycle = &corev1.Lifecycle{}
 		}
-	} else {
-		// Main container doesn't have valid postStart, apply config if available
-		if configHasValidPostStart {
-			// set main container lifecycle
-			if mainContainer.Lifecycle == nil {
-				mainContainer.Lifecycle = &corev1.Lifecycle{}
+		if mainContainerHasValidPostStart {
+			// The user has a postStart hook. Merge using "--" separator so that
+			// envd-run.sh executes the user's command after its own initialisation.
+			mergedCmd := mergePostStartExecCommands(config.MainContainer.Lifecycle.PostStart, mainContainer.Lifecycle.PostStart)
+			if mergedCmd != nil {
+				log.V(utils.DebugLogLevel).Info("merging user's postStart command into injected script",
+					"userCommand", mainContainer.Lifecycle.PostStart.Exec.Command,
+					"injectedCommand", config.MainContainer.Lifecycle.PostStart.Exec.Command)
+				mainContainer.Lifecycle.PostStart = &corev1.LifecycleHandler{
+					Exec: &corev1.ExecAction{Command: mergedCmd},
+				}
+			} else {
+				// Cannot merge non-Exec handlers (e.g. HTTPGet/TCPSocket); keep the
+				// existing hook and skip injection of the injected postStart.
+				log.Error(nil, "cannot merge non-Exec postStart hooks, keeping existing hook",
+					"existingHook", mainContainer.Lifecycle.PostStart,
+					"injectedHook", config.MainContainer.Lifecycle.PostStart)
 			}
+		} else {
+			// Main container doesn't have a valid postStart; apply the config directly.
 			mainContainer.Lifecycle.PostStart = config.MainContainer.Lifecycle.PostStart
 		}
 	}
@@ -265,6 +276,34 @@ func setMainContainerConfigWhenInjectRuntimeSidecar(ctx context.Context, mainCon
 		mainContainer.VolumeMounts = make([]corev1.VolumeMount, 0, len(config.MainContainer.VolumeMounts))
 	}
 	mainContainer.VolumeMounts = append(mainContainer.VolumeMounts, config.MainContainer.VolumeMounts...)
+}
+
+// mergePostStartExecCommands merges the injected postStart Exec command with the user's
+// postStart Exec command using the "--" separator pattern. Everything after "--" is passed
+// to envd-run.sh as individual arguments representing the user command, executed directly
+// without shell interpretation. This follows the kubectl exec convention.
+//
+// Precondition: the caller has already verified that both the injected config and the main
+// container have valid postStart hooks (configHasValidPostStart && mainContainerHasValidPostStart),
+// so neither handler is nil and both have a non-nil action. Consequently, the nil checks below
+// are defensive; the only realistic failure mode is a non-Exec handler, which returns nil.
+// Returns nil if either handler is non-Exec (cannot be merged).
+func mergePostStartExecCommands(injected, user *corev1.LifecycleHandler) []string {
+	if injected == nil || injected.Exec == nil || user == nil || user.Exec == nil {
+		return nil
+	}
+	// If the user's command is empty, there is nothing to append.
+	if len(user.Exec.Command) == 0 {
+		return injected.Exec.Command
+	}
+	// Use "--" separator following the kubectl exec pattern. Everything after "--"
+	// is passed as individual arguments to envd-run.sh which executes the command
+	// directly without shell interpretation.
+	merged := make([]string, 0, len(injected.Exec.Command)+1+len(user.Exec.Command))
+	merged = append(merged, injected.Exec.Command...)
+	merged = append(merged, "--")
+	merged = append(merged, user.Exec.Command...)
+	return merged
 }
 
 func findVolumeMountByName(volumeMounts []corev1.VolumeMount, name string) bool {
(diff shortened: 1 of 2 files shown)


pkg/utils/sidecarutils/sidecar_config_inject.go
```diff
@@ -237,19 +237,30 @@ func setMainContainerConfigWhenInjectRuntimeSidecar(ctx context.Context, mainCon
...
+				log.V(utils.DebugLogLevel).Info("merging user's postStart command into injected script",
+					"userCommand", mainContainer.Lifecycle.PostStart.Exec.Command,
+					"injectedCommand", config.MainContainer.Lifecycle.PostStart.Exec.Command)
+				mainContainer.Lifecycle.PostStart = &corev1.LifecycleHandler{
+					Exec: &corev1.ExecAction{Command: mergedCmd},
+				}
+			} else {
+				// Cannot merge non-Exec handlers (e.g. HTTPGet/TCPSocket); keep the
+				// existing hook and skip injection of the injected postStart.
+				log.V(utils.DebugLogLevel).Info("cannot merge non-Exec postStart hooks, keeping existing hook",
```
furykerry: this should be an warning not debug, kubernetes container cannot have multiple poststarthook of different kinds

pkg/utils/sidecarutils/sidecar_config_inject_test.go
```diff
@@ -601,16 +601,16 @@ func TestSetMainContainerConfigWhenInjectRuntimeSidecar(t *testing.T) {
...
 					},
 				},
 			},
-			expectedEnvCount:         1, // conflicting postStart only skips postStart injection, env still injected
-			expectedVolumeMountCount: 1, // conflicting postStart only skips postStart injection, volumeMounts still injected
+			expectedEnvCount:         1,
+			expectedVolumeMountCount: 1,
 			hasPostStart:             true,
-			postStartCommand:         []string{"echo", "old"}, // keeps existing, NOT overridden
+			postStartCommand:         []string{"bash", "/mnt/envd/envd-run.sh", "--", "echo", "old"}, // merged with -- separator
```
furykerry, line 614: test does not actually assert the merged command contents, plz compare the actual slices instead of simple length assertion

pkg/utils/sidecarutils/sidecar_config_inject_test.go
```diff
@@ -580,7 +580,7 @@ func TestSetMainContainerConfigWhenInjectRuntimeSidecar(t *testing.T) {
 			postStartCommand:         []string{"bash", "-c", "/mnt/envd/envd-run.sh"},
 		},
 		{
-			name: "container with existing lifecycle - conflicting postStart should skip all injection",
+			name: "container with existing lifecycle - merge postStart with -- separator",
```
furykerry, line 584: Add a case to TestSetMainContainerConfigWhenInjectRuntimeSidecar (or a dedicated test) where the main container already has an HTTPGet/TCPSocket post-start and the config injects an Exec hook, then assert the original non-Exec handler is preserved.

pkg/utils/sidecarutils/sidecar_config_inject.go
BH4AWS: In this function, I use klog as the logging client for all function line locations.

pkg/utils/sidecarutils/sidecar_config_inject_test.go
BH4AWS: The current review comment is correct; the current assertion only compares the slice length, not the actual content. We should use reflect.DeepEqual to compare the entire command slice.

pkg/utils/sidecarutils/sidecar_config_inject.go
```diff
@@ -237,19 +236,29 @@ func setMainContainerConfigWhenInjectRuntimeSidecar(ctx context.Context, mainCon
...
+				klog.V(utils.DebugLogLevel).InfoS("merging user's postStart command into injected script",
+					"userCommand", mainContainer.Lifecycle.PostStart.Exec.Command,
+					"injectedCommand", config.MainContainer.Lifecycle.PostStart.Exec.Command)
+				mainContainer.Lifecycle.PostStart = &corev1.LifecycleHandler{
+					Exec: &corev1.ExecAction{Command: mergedCmd},
+				}
+			} else {
+				// Cannot merge non-Exec handlers (e.g. HTTPGet/TCPSocket); keep the
+				// existing hook and skip injection of the injected postStart.
+				klog.Warningf("cannot merge non-Exec postStart hooks, keeping existing hook, existingHook: %v, injectedHook: %v",
```
furykerry: plz use structured logging

pkg/utils/sidecarutils/sidecar_config_inject.go
```diff
@@ -237,19 +236,29 @@ func setMainContainerConfigWhenInjectRuntimeSidecar(ctx context.Context, mainCon
...
-		if configHasValidPostStart {
-			// set main container lifecycle
-			if mainContainer.Lifecycle == nil {
-				mainContainer.Lifecycle = &corev1.Lifecycle{}
+		if mainContainerHasValidPostStart {
+			// The user has a postStart hook. Merge using "--" separator so that
+			// envd-run.sh executes the user's command after its own initialisation.
+			mergedCmd := mergePostStartExecCommands(config.MainContainer.Lifecycle.PostStart, mainContainer.Lifecycle.PostStart)
+			if mergedCmd != nil {
+				klog.V(utils.DebugLogLevel).InfoS("merging user's postStart command into injected script",
```
furykerry: plz use logf not klog in controller code

pkg/utils/sidecarutils/sidecar_config_inject.go
```diff
@@ -225,9 +226,7 @@ func checkInjectionConflicts(pod *corev1.Pod, config SidecarInjectConfig) error
 	return nil
 }
 
-func setMainContainerConfigWhenInjectRuntimeSidecar(ctx context.Context, mainContainer *corev1.Container, config SidecarInjectConfig) {
-	log := logf.FromContext(ctx)
-
+func setMainContainerConfigWhenInjectRuntimeSidecar(_ context.Context, mainContainer *corev1.Container, config SidecarInjectConfig) {
```
furykerry: _ signals the context is unused, which is an anti-pattern when the caller (line 162) explicitly passes a real context with values, consider restore ctx context.Context and use it for logging

pkg/utils/sidecarutils/sidecar_config_inject.go
```diff
@@ -267,6 +276,31 @@ func setMainContainerConfigWhenInjectRuntimeSidecar(ctx context.Context, mainCon
 	mainContainer.VolumeMounts = append(mainContainer.VolumeMounts, config.MainContainer.VolumeMounts...)
 }
 
+// mergePostStartExecCommands merges the injected postStart Exec command with the user's
+// postStart Exec command using the "--" separator pattern. Everything after "--" is passed
+// to envd-run.sh as individual arguments representing the user command, executed directly
+// without shell interpretation. This follows the kubectl exec convention.
+// Returns nil if either handler is non-Exec (cannot be merged).
+func mergePostStartExecCommands(injected, user *corev1.LifecycleHandler) []string {
+	if injected == nil || injected.Exec == nil || user == nil || user.Exec == nil {
```
furykerry, line 292: the func is only called within configHasValidPostStart & mainContainerHasValidPostStart condition, so injected == nil and user == nil is unlikely, consider comment the precondition in func comments

pkg/utils/sidecarutils/sidecar_config_inject.go
```diff
@@ -267,6 +276,31 @@ func setMainContainerConfigWhenInjectRuntimeSidecar(ctx context.Context, mainCon
...
+		return nil
+	}
+	// If the user's command is empty, there is nothing to append.
+	if len(user.Exec.Command) == 0 {
+		merged := make([]string, len(injected.Exec.Command))
+		copy(merged, injected.Exec.Command)
+		return merged
```
furykerry: if user.Exec.Command is empty, there is nothing to append, consider just return injected.Exec.Command

pkg/utils/sidecarutils/sidecar_config_inject.go
BH4AWS: using log provider

pkg/utils/sidecarutils/sidecar_config_inject.go
BH4AWS: done

pkg/utils/sidecarutils/sidecar_config_inject.go
BH4AWS: done
```
func setMainContainerConfigWhenInjectRuntimeSidecar(ctx context.Context, mainContainer *corev1.Container, config SidecarInjectConfig) {
	log := logf.FromContext(ctx)
...
}
```

pkg/utils/sidecarutils/sidecar_config_inject.go
BH4AWS: // Precondition: the caller has already verified that both the injected config and the main
// container have valid postStart hooks (configHasValidPostStart && mainContainerHasValidPostStart),
// so neither handler is nil and both have a non-nil action. Consequently, the nil checks below
// are defensive; the only realistic failure mode is a non-Exec handler, which returns nil.

pkg/utils/sidecarutils/sidecar_config_inject.go
BH4AWS: done

BH4AWS: done

PR #557, opened 2026-06-20
fix(sandbox): avoid triggering in-place update for metadata-only changes

furykerry:

This PR optimizes the in-place update flow so that metadata-only changes (labels/annotations) bypass the full InplaceUpdate condition, preventing unnecessary Ready=False transitions for sandbox claims.

**Key changes:**

1. **Metadata-only update detection** (`common_inplace_update_handler.go`): New `isMetadataOnlyChange` function detects when only labels/annotations differ between the pod and sandbox template (no image or resource changes). When detected, the controller patches pod metadata directly without setting the InplaceUpdate condition, avoiding blocked sandbox readiness.

2. **Export `ResourcesEqual`** (`inplace_update.go`): Renamed `resourcesEqual` to exported `ResourcesEqual` and replaced all `reflect.DeepEqual` usages with it in `DefaultGeneratePatchBodyFunc`, `DefaultGenerateResizeSubresourceBody`, and `isPodResourceResizeCompleted`. The subset-based comparison ignores system-injected resources (e.g., ephemeral-storage from LimitRanger) so they don't trigger false in-place updates.

3. **Checkpoint utils consistency** (`checkpoint/utils.go`): Uses exported `ResourcesEqual` instead of `reflect.DeepEqual` for consistent resource comparison.

4. **`MergePodLabels` utility** (`interface.go`, `create.go`): New `MergePodLabels(sandbox, labels)` function centralizes pod label propagation. `create.go` refactored to use it.

5. **Claim validation message** (`claim.go`): Updated error message to "requires at least one of image or resources to be set".

NONE

1. Run unit tests:
   ```bash
   go test ./pkg/controller/sandbox/core/ -run TestIsMetadataOnlyChange -v
   go test ./pkg/utils/inplaceupdate/ -run TestIsPodResourceResizeCompleted -v
   go test ./pkg/sandbox-manager/infra/ -run TestMergePodLabels -v
   go test ./pkg/sandbox-manager/infra/sandboxcr/ -run TestInfra_ClaimSandbox -v
   ```
2. Verify that a sandbox claim with only label/annotation changes does not set the InplaceUpdate condition
3. Verify that a sandbox claim with image or resource changes still goes through the full in-place update flow

- `isMetadataOnlyChange` uses subset-based `ResourcesEqual` comparison — extra resources injected into the pod by admission webhooks are ignored so metadata-only changes are not mistakenly treated as in-place updates
- `isPodResourceResizeCompleted` now uses `ResourcesEqual` (subset-based) instead of `reflect.DeepEqual` (exact match), consistent with the rest of the in-place update code
- `MergePodLabels` tests are co-located in `interface_test.go` using a `mockSandboxForLabels` mock that implements the full `Sandbox` interface
- `TestIsPodResourceResizeCompleted` is table-driven with 7 cases including edge cases for subset comparison

---

- Minor comment fix in `pkg/cache/tasks.go` (added trailing period)
- Comment addition in `pkg/sandbox-manager/config/infra.go` clarifying metadata handling


pkg/utils/inplaceupdate/inplace_update.go
```diff
@@ -596,7 +596,7 @@ func isPodResourceResizeCompleted(pod *corev1.Pod) bool {
 		if !ok || status.Resources == nil {
 			return false
 		}
-		if !resourcesEqual(c.Resources, *status.Resources) {
+		if !reflect.DeepEqual(c.Resources, *status.Resources) {
```
zmberg: Why was this changed?

furykerry: fixed

PR #561, opened 2026-06-22
Feat/sandbox gateway auth

chengzhycn:

Add inbound request authentication to sandbox-gateway via `X-Access-Token` header validation.

**Core Changes:**
- Add `AccessToken` field to `Route` struct, extracted from the Sandbox CR annotation `agents.kruise.io/runtime-access-token`
- Implement token validation in the Envoy Go HTTP filter's `DecodeHeaders` method using `crypto/subtle.ConstantTimeCompare` to prevent timing attacks
- Return HTTP 401 when the token is missing or invalid
- Keep `X-Access-Token` header forwarded to upstream envd for downstream re-validation (layered auth)

**Configuration:**
- Add `enable-auth` boolean config field to the filter `Config`. Authentication is **disabled by default** for backward compatibility, and must be explicitly enabled via Envoy filter configuration.

**Testing:**
- Unit tests covering: valid token, invalid token, missing token, no-token-configured (backward compat), and auth-disabled scenarios
- Unit test for `Route.AccessToken` extraction from Sandbox annotations
- E2E tests (`test/e2b/test_gateway_auth.py`): 4 scenarios testing valid/missing/invalid token and host-based routing with auth
- Updated existing gateway E2E tests to include access token header
- Updated `hack/run-e2b-e2e-test.sh` to skip gateway auth tests when not running in gateway mode

NONE

1. **Unit tests:**
   ```bash
   go test ./pkg/sandbox-gateway/filter/ -run "TestDecodeHeadersAccessToken|TestDecodeHeadersAuthDisabled" -v
   go test ./pkg/utils/proxyutils/ -run "TestGetRouteFromSandbox" -v
   ```

2. **E2E tests** (requires cluster with gateway deployed and `enable-auth: true` in filter config):
   ```bash
   WITH_GATEWAY=true ./hack/run-e2b-e2e-test.sh
   ```

3. **Manual verification:**
   - Deploy sandbox-gateway with `"enable-auth": true` in filter config
   - Create a sandbox (it should get `agents.kruise.io/runtime-access-token` annotation)
   - Request without `X-Access-Token` → expect HTTP 401
   - Request with wrong token → expect HTTP 401
   - Request with correct token → expect forwarded to upstream

- Token comparison uses `crypto/subtle.ConstantTimeCompare` to prevent timing side-channel attacks
- `EnableAuth` defaults to `false` — no behavioral change for existing deployments until explicitly enabled
- The `X-Access-Token` header is intentionally **not** stripped before forwarding, because envd performs a second validation
- When `Route.AccessToken` is empty (sandbox has no token annotation), auth is skipped regardless of `EnableAuth`, maintaining compatibility with legacy sandboxes

### DIFF
--- a/config/sandbox-gateway/configmap.yaml
+++ b/config/sandbox-gateway/configmap.yaml
@@ -79,6 +79,7 @@ data:
                               sandbox-header-name: "e2b-sandbox-id"
                               sandbox-port-header: "e2b-sandbox-port"
                               default-port: "49983"
+                              enable-auth: true
                       - name: envoy.filters.http.router
                         typed_config:
                           "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
--- a/config/sandbox-gateway/configmap.yaml
+++ b/config/sandbox-gateway/configmap.yaml
@@ -79,7 +79,7 @@ data:
                               sandbox-header-name: "e2b-sandbox-id"
                               sandbox-port-header: "e2b-sandbox-port"
                               default-port: "49983"
-                              enable-auth: true
+                              enable-auth: false
                       - name: envoy.filters.http.router
                         typed_config:
                           "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
--- a/pkg/sandbox-gateway/filter/config_test.go
+++ b/pkg/sandbox-gateway/filter/config_test.go
@@ -148,38 +148,26 @@ func TestDefaultConfig(t *testing.T) {
 	}
 }
 
-// helperTypedStructAny creates an *anypb.Any wrapping a v3.TypedStruct with given fields.
-func helperTypedStructAny(t *testing.T, fields map[string]interface{}) *anypb.Any {
-	t.Helper()
-	ts := &v3.TypedStruct{}
-	if fields != nil {
-		s, err := structpb.NewStruct(fields)
-		if err != nil {
-			t.Fatalf("failed to create structpb.Struct: %v", err)
-		}
-		ts.Value = s
-	}
-	a, err := anypb.New(ts)
-	if err != nil {
-		t.Fatalf("failed to marshal TypedStruct to Any: %v", err)
-	}
-	return a
-}
-
 func TestConfigParserParse(t *testing.T) {
 	parser := &ConfigParser{}
 
 	tests := []struct {
 		name              string
-		any               *anypb.Any
+		input             *anypb.Any
 		wantErr           bool
 		wantSandboxHeader string
 		wantHostHeader    string
 		wantPortHeader    string
 		wantDefaultPort   string
+		wantEnableAuth    bool
 	}{
 		{
-			name:              "nil value in TypedStruct returns defaults",
+			name: "nil value in TypedStruct returns defaults",
+			input: func() *anypb.Any {
+				ts := &v3.TypedStruct{}
+				a, _ := anypb.New(ts)
+				return a
+			}(),
 			wantErr:           false,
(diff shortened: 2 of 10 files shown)


chengzhycn:
@furykerry Should I add this feature to sandbox-manager as well?

config/sandbox-gateway/configmap.yaml
```diff
@@ -79,6 +79,7 @@ data:
                               sandbox-header-name: "e2b-sandbox-id"
                               sandbox-port-header: "e2b-sandbox-port"
                               default-port: "49983"
+                              enable-auth: true
```
furykerry: per code comment， enable-auth should be false for backward compatibility, consider change this value to false

pkg/sandbox-gateway/filter/filter.go
```diff
@@ -104,6 +111,24 @@ func (f *sandboxFilter) DecodeHeaders(header api.RequestHeaderMap, endStream boo
 		return api.LocalReply
 	}
 
+	// Authenticate the request if auth is enabled and the sandbox has an access token configured.
+	// When EnableAuth is false (default), the gateway skips token validation for backward compatibility.
+	if f.config.EnableAuth && route.AccessToken != "" {
+		requestToken := headers[accessTokenHeader]
```
furykerry: HTTP headers are case-insensitive by spec. Using a case-sensitive map lookup for a security-critical authentication check is fragile and could cause valid tokens to be rejected depending on the client's header casing.  Consider use header.Get(accessTokenHeader) instead of the flat map lookup. This uses Envoy's built-in case-insensitive header matching, e.g. 

```
// Recommended (case-insensitive Envoy API):
requestToken, _ := header.Get(accessTokenHeader)
```

test/e2b/test_gateway_auth.py
```diff
@@ -0,0 +1,185 @@
...
+    Args:
+        sandbox_id: The sandbox ID in namespace--name format (e.g., default--my-sandbox).
+
+    Returns:
+        The access token string, or empty string if not set.
+    """
+    # sandbox_id format is "namespace--name", extract the name part
+    name = sandbox_id.split("--")[1] if "--" in sandbox_id else sandbox_id
+    result = subprocess.run(
+        ["kubectl", "get", "sandbox", name, "-o", "json"],
```
furykerry: parse the namespace from sandbox id and pass to kubectl ?

pkg/sandbox-gateway/filter/config_test.go
```diff
@@ -230,6 +238,9 @@ func TestConfigParserParse(t *testing.T) {
 		"sandbox-port-header": "x-port",
 		"default-port":        "8080",
 	})
+	tests[4].any = helperTypedStructAny(t, map[string]interface{}{
```
furykerry: is it possible to assign test value .any directly in the tests assignment? for example 

```
	tests := []struct {
		name              string
		any               *anypb.Any
		wantErr           bool
		wantSandboxHeader string
		wantHostHeader    string
		wantPortHeader    string
		wantDefaultPort   string
	}{
		{
			name:              "nil value in TypedStruct returns defaults",
            any:  func() *anypb.Any {
		            ts := &v3.TypedStruct{}
		            a. , _ := anypb.New(ts)
		            return a
	        }()
			wantErr:           false,
			wantSandboxHeader: DefaultSandboxHeaderName,
			wantHostHeader:    DefaultHostHeaderName,
			wantPortHeader:    DefaultSandboxPortHeader,
			wantDefaultPort:   DefaultSandboxPort,
		},
```

PR #563, opened 2026-06-22
fix(e2b): stabilize pagination for duplicate timestamps

### LINKED ISSUE #439: [BUG] Listing sandboxes can silently hide some of them when they share the same claim time

zhuangzhewei09:
#### What this PR does

Fixes pagination for `GET /v2/sandboxes` and snapshot listing when multiple objects share the same timestamp sort key.

The previous cursor only stored the timestamp of the last returned item. Since claim timestamps and checkpoint creation timestamps can collide, the next page searched for the first item strictly after that timestamp and skipped remaining objects with the same timestamp.

This PR adds an optional stable cursor to the internal paginator. When a unique key is configured, pagination now sorts and resumes by `(sort key, unique key)`, encodes the cursor as an opaque base64 JSON token, and keeps raw sort-key tokens compatible for existing clients.

#### Changes

- Add optional `GetUniqueKey` support to `pkg/utils/pagination.Paginator`.
- Use sandbox ID as the tiebreaker for sandbox listing.
- Use Kubernetes object identity (`namespace/name`) as the tiebreaker for snapshot listing while keeping the response `SnapshotID` unchanged.
- Add regression coverage for duplicate sandbox claim timestamps, duplicate checkpoint creation timestamps, and duplicate checkpoint IDs across different checkpoint objects.

#### Tests

```bash
go test ./pkg/utils/pagination
go test ./pkg/sandbox-manager -run 'TestSandboxManager_ListSandboxes|TestSandboxManager_ListCheckpoints'
go test ./pkg/servers/e2b -run 'TestListSandboxes_Pagination|TestListSandboxes_PaginationDuplicateClaimTime|TestListSnapshots'
go test ./pkg/utils/pagination ./pkg/sandbox-manager ./pkg/servers/e2b
git diff --check
```

Fixes #439

### DIFF
--- a/pkg/servers/e2b/list.go
+++ b/pkg/servers/e2b/list.go
@@ -143,6 +143,9 @@ func (sc *Controller) ListSandboxes(r *http.Request) (web.ApiResponse[[]*models.
 		GetKey: func(sbx infra.Sandbox) string {
 			return sbx.GetAnnotations()[agentsv1alpha1.AnnotationClaimTime]
 		},
+		GetUniqueKey: func(sbx infra.Sandbox) string {
+			return sbx.GetSandboxID()
+		},
 	})
 	var headers map[string]string
 	if nextToken != "" {
@@ -284,6 +287,9 @@ func (sc *Controller) listCheckpointsForUser(ctx context.Context, user *models.C
 		NextToken: nextTokenParam,
 		GetKey: func(cp infra.CheckpointInfo) string {
 			return cp.CreationTimestamp // Sort by creation timestamp
+		},
+		GetUniqueKey: func(cp infra.CheckpointInfo) string {
+			return fmt.Sprintf("%s/%s", cp.Namespace, cp.Name)
 		},
 		Filter: filter,
 	})
--- a/pkg/servers/e2b/list_test.go
+++ b/pkg/servers/e2b/list_test.go
@@ -519,6 +519,151 @@ func TestListSandboxes_Pagination(t *testing.T) {
 	}
 }
 
+func TestListSandboxes_PaginationDuplicateClaimTime(t *testing.T) {
+	tests := []struct {
+		name         string
+		sandboxCount int
+		limit        int
+		claimTime    string
+	}{
+		{
+			name:         "same claim time across page boundary",
+			sandboxCount: 5,
+			limit:        2,
+			claimTime:    "2024-01-01T00:00:01Z",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			templateName := "pagination-duplicate-template"
+			controller, fc, teardown := Setup(t)
+			defer teardown()
+			user := &models.CreatedTeamAPIKey{
+				ID:   keys.AdminKeyID,
+				Key:  InitKey,
+				Name: "admin",
+			}
+
+			tmpl := agentsv1alpha1.EmbeddedSandboxTemplate{
+				Template: &corev1.PodTemplateSpec{
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{
+								Name:  "main",
+								Image: "test-image",
+							},
+						},
+					},
+				},
+			}
+			sbs := &agentsv1alpha1.SandboxSet{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      templateName,
+					Namespace: Namespace,
+					UID:       types.UID(uuid.NewString()),
+				},
+				Spec: agentsv1alpha1.SandboxSetSpec{
+					EmbeddedSandboxTemplate: tmpl,
+				},
+			}
+			require.NoError(t, fc.Create(t.Context(), sbs))
+			defer func() {
+				_ = fc.Delete(context.Background(), sbs)
+			}()
+
+			expectedSandboxIDs := make([]string, 0, tt.sandboxCount)
+			now := metav1.Now()
+			for i := 0; i < tt.sandboxCount; i++ {
+				sbxName := fmt.Sprintf("%s-%d", templateName, i)
+				sbx := &agentsv1alpha1.Sandbox{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      sbxName,
+						Namespace: Namespace,
+						Labels: map[string]string{
+							agentsv1alpha1.LabelSandboxTemplate:  templateName,
+							agentsv1alpha1.LabelSandboxIsClaimed: "true",
+						},
+						Annotations: map[string]string{
+							agentsv1alpha1.AnnotationClaimTime: tt.claimTime,
+							agentsv1alpha1.AnnotationOwner:     user.ID.String(),
+						},
+						UID:               types.UID(uuid.NewString()),
+						CreationTimestamp: now,
+					},
+					Spec: agentsv1alpha1.SandboxSpec{
+						EmbeddedSandboxTemplate: tmpl,
+					},
+					Status: agentsv1alpha1.SandboxStatus{
+						Phase: agentsv1alpha1.SandboxRunning,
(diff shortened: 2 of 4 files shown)


zhuangzhewei09:
Hello @furykerry @AiRanthem @zmberg ,
I wanted to politely follow up on this PR. The CI checks are passing, and I believe it is ready for review. I would appreciate any feedback whenever you have the opportunity.
Thank you!

AiRanthem:
Thanks for working on this. The overall stable-cursor direction looks right to me.

For `GET /v2/sandboxes`, this should fix the duplicate `claim-timestamp` pagination skip issue: the paginator now sorts and resumes by `(sort key, unique key)`, the cursor is encoded as an opaque base64url JSON token, and sandbox listing uses `sandboxID` as the stable tiebreaker. That addresses the core problem where multiple sandboxes share the same timestamp and the next page previously skipped the remaining objects with the same key.

One thing I think should be fixed before merge: snapshot listing currently uses `cp.CheckpointID` as `GetUniqueKey`. `CheckpointID` is not guaranteed to be unique across all list-visible checkpoint objects; there are existing scenarios where the same checkpoint ID can appear in different namespaces, and an admin listing can see both. If two checkpoints have the same `CreationTimestamp` and the same `CheckpointID`, the cursor key is still not unique, so the same skip bug can remain for snapshots.

Could we change the snapshot tiebreaker to the Kubernetes object identity instead, for example:

```go
GetUniqueKey: func(cp infra.CheckpointInfo) string {
    return cp.Namespace + "/" + cp.Name
}
```

or include `namespace/name` as a suffix if you still want `CheckpointID` in the token?

It would also be good to add a regression test for snapshots where multiple checkpoints have the same creation timestamp and the same checkpoint ID but different `namespace/name`, with pagination `limit` crossing that duplicate-key boundary. The PR description currently says there is a `namespace/name` fallback, but the current code appears to return only `CheckpointID`.

After that adjustment, I think this PR should cover the duplicate timestamp pagination issue much more robustly.

zhuangzhewei09:
> Thanks for working on this. The overall stable-cursor direction looks right to me.
> 
> For `GET /v2/sandboxes`, this should fix the duplicate `claim-timestamp` pagination skip issue: the paginator now sorts and resumes by `(sort key, unique key)`, the cursor is encoded as an opaque base64url JSON token, and sandbox listing uses `sandboxID` as the stable tiebreaker. That addresses the core problem where multiple sandboxes share the same timestamp and the next page previously skipped the remaining objects with the same key.
> 
> One thing I think should be fixed before merge: snapshot listing currently uses `cp.CheckpointID` as `GetUniqueKey`. `CheckpointID` is not guaranteed to be unique across all list-visible checkpoint objects; there are existing scenarios where the same checkpoint ID can appear in different namespaces, and an admin listing can see both. If two checkpoints have the same `CreationTimestamp` and the same `CheckpointID`, the cursor key is still not unique, so the same skip bug can remain for snapshots.
> 
> Could we change the snapshot tiebreaker to the Kubernetes object identity instead, for example:
> 
> ```go
> GetUniqueKey: func(cp infra.CheckpointInfo) string {
>     return cp.Namespace + "/" + cp.Name
> }
> ```
> 
> or include `namespace/name` as a suffix if you still want `CheckpointID` in the token?
> 
> It would also be good to add a regression test for snapshots where multiple checkpoints have the same creation timestamp and the same checkpoint ID but different `namespace/name`, with pagination `limit` crossing that duplicate-key boundary. The PR description currently says there is a `namespace/name` fallback, but the current code appears to return only `CheckpointID`.
> 
> After that adjustment, I think this PR should cover the duplicate timestamp pagination issue much more robustly.

Thanks for the review. I updated the snapshot pagination cursor according to your suggestion.

Snapshot listing now uses the Kubernetes object identity (`namespace/name`) as the stable tiebreaker instead of `CheckpointID`, while the response `SnapshotID` remains unchanged. I also added a regression test covering multiple checkpoints with the same `CreationTimestamp` and the same `CheckpointID` but different `namespace/name`, with pagination crossing that duplicate-key boundary.

@AiRanthem

AiRanthem:
Fixes #581

PR #564, opened 2026-06-22
fix(updateops): handle deletion before mutual-exclusion check

zmberg:
## Summary
- Move the `DeletionTimestamp` check before the mutual-exclusion check in `SandboxUpdateOps` reconciler, so that a deleting ops is not blocked from finalizer removal by another active ops in the same namespace.
- Previously, when another `SandboxUpdateOps` was in `Updating` phase, the controller would skip reconciliation entirely — including deletion handling. This caused finalizer removal to be delayed until the active ops completed (up to hours in production).

## Test plan
- [x] Unit test `TestReconcile_DeletionNotBlockedByActiveOps`: verifies deletion proceeds while another ops is Updating
- [x] E2E test: creates two ops in the same namespace, verifies the completed one can be deleted promptly while the other is stuck in Updating

### DIFF
--- a/pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
+++ b/pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
@@ -103,7 +103,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	// 2. Check if another SandboxUpdateOps in the same namespace is already Updating
+	// 2. Handle deletion: clean up sandbox labels.
+	// This must run before the mutual-exclusion check so that a deleting ops
+	// is not blocked by an active one — it only removes the finalizer and
+	// never performs updates on sandboxes.
+	if !ops.DeletionTimestamp.IsZero() {
+		return r.handleDeletion(ctx, ops)
+	}
+
+	// 3. Check if another SandboxUpdateOps in the same namespace is already Updating
 	// TODO: This is a short-term solution to prevent concurrent ops in the same namespace.
 	//  A more robust approach would be using a webhook to reject creation when an active ops exists,
 	//  or implementing a queue/priority-based scheduling mechanism.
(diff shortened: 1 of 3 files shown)


HARSHRAJ2789:
Traced the deadlock this closes and applied the controller change to current master to check it
still fits. The reorder is right, and the severity is worth putting a number on.

`:123` returns `ctrl.Result{}, nil` with no requeue, so a deleting ops that loses the
mutual-exclusion race never reaches `handleDeletion` and never sheds its finalizer. Nothing brings
it back quickly either. `For(&SandboxUpdateOps{})` only enqueues it when it changes, and it changes
nothing while blocked, while `sandbox_event_handler.go:44` enqueues the ops named in a sandbox's
own label, so an active ops's sandboxes wake that ops and leave this one asleep. No `SyncPeriod` is set
for this manager, which leaves controller-runtime's ten hour default as the recovery path, and a
resync that lands while the other ops is still `Updating` blocks again for another cycle. A user
deleting a SandboxUpdateOps can therefore watch it sit in Terminating for hours.

The controller change applies to current master unchanged and `./pkg/controller/sandboxupdateops/`
passes with it. The conflict is confined to `test/e2e/sandboxupdateops_test.go`, so the rebase is a
test-file merge.

One thing about the comment the change carries. It says deletion handling "only removes the
finalizer and never performs updates on sandboxes", and `handleDeletion` does both of the things
that sentence rules out: it patches every matching sandbox at `:595`, and at `:588` it can clear
`spec.upgradePolicy` for sandboxes whose upgrade already succeeded, which is a spec write.

The reorder still looks safe to me, for a stronger reason than the comment gives.
`LabelSandboxUpdateOps` is a single key, so a sandbox carries exactly one ops name, and
`handleDeletion` lists on its own name at `:567` and rechecks it at `:578`. The sandboxes it writes
to are disjoint from the active ops's set by construction, which is what makes running it ahead of
the exclusion check sound. Worth saying that in the comment instead, since the current wording
would let a later reader conclude the ordering is safe because nothing is written, and then break
it by adding a write.

`TestReconcile_DeletionNotBlockedByActiveOps` covers the case well: it asserts the deleting ops
reaches NotFound, the label is gone from the sandbox, and the active ops is untouched in
`Updating`.

PR #566, opened 2026-06-23
feature(sandbox-manager): refine paused retention timeout handling

AiRanthem:
## Summary
- Compute create-time auto-pause retention from the persisted reserve-paused value instead of storing a parsed duration on request extensions.
- Split timeout math into a pure helper so resume placeholder calculation has no annotation side effects.
- Update E2B tests around resume placeholder/default retention behavior.

## Test Plan
- GOCACHE=/private/tmp/go-build-cache go test ./pkg/pausedretention
- GOCACHE=/private/tmp/go-build-cache go test ./pkg/controller/sandbox -run 'TestSandboxReconciler_(HandleShutdownTimeout|HandlePauseTimeout|AutoPauseReservePausedRetention)'\n- GOCACHE=/private/tmp/go-build-cache go test ./pkg/servers/e2b/models\n- GOCACHE=/private/tmp/go-build-cache go test ./pkg/servers/e2b -run 'Test.*(Resume|Pause|Timeout|ReservePaused)'\n- git diff --check


pkg/pausedretention/retention.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"fmt"
+	"time"
+
+	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
+	"github.com/openkruise/agents/pkg/utils/timeout"
+)
+
+func ParseReservePausedSandboxFor(raw string) (time.Duration, error) {
+	if raw == timeout.ReservePausedSandboxForDefaultValue {
```
furykerry: why introduce an "default" value explicitly instead of leaving the annotation empty.

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -558,35 +561,115 @@ func (r *SandboxReconciler) checkTimers(ctx context.Context, box *agentsv1alpha1
...
+	modified.Spec.Paused = true
+
+	// If the sandbox has a paused-retention policy, extend ShutdownTime so the
+	// sandbox is preserved for the configured duration after being paused.
+	if retention, managed := r.resolveRetentionAnnotationOrDefault(box); managed {
+		if box.Spec.ShutdownTime != nil {
+			newShutdown := metav1.NewTime(pausedretention.PausedShutdownTime(now.Time, retention))
+			modified.Spec.ShutdownTime = &newShutdown
+			// Clear PauseTime to the same value so it won't re-trigger on the next reconcile.
+			modified.Spec.PauseTime = &newShutdown
```
furykerry, line 674: Consider setting PauseTime = nil instead, which is the more natural "clear" semantics and also prevents re-triggering (since handlePauseTimeout checks box.Spec.PauseTime == nil)

pkg/servers/e2b/pause_resume.go
```diff
@@ -84,12 +97,57 @@ func resumeSandboxErrorCode(err error) int {
...
+		return timeout.DefaultReservePausedSandboxFor, defaultReservePausedForValue()
+	}
+	if managed {
+		return retention, nil
+	}
+	return timeout.DefaultReservePausedSandboxFor, defaultReservePausedForValue()
+}
+
+// defaultReservePausedForValue copies ReservePausedSandboxForDefaultValue for safety.
+func defaultReservePausedForValue() *string {
```
furykerry: Consider using ptr.To(timeout.ReservePausedSandboxForDefaultValue) from k8s.io/utils/ptr (already imported in test files) for consistency with the rest of the codebase.

api/v1alpha1/annotations.go
```diff
@@ -21,6 +21,8 @@ package v1alpha1
 const (
 	AnnotationRuntimeURL         = InternalPrefix + "runtime-url"
 	AnnotationRuntimeAccessToken = InternalPrefix + "runtime-access-token"
+	// AnnotationReservePausedSandboxFor stores the internal paused-retention policy parsed by pkg/pausedretention.
```
furykerry: the comments is quite confusing, from the code these annotation actually save the retention duration, consider rename to reserve-paused-sandbox-duration， and rename x-e2b-kruise-reserve-paused-sandbox-for  to x-e2b-kruise-reserve-paused-sandbox-duration

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -558,35 +561,115 @@ func (r *SandboxReconciler) checkTimers(ctx context.Context, box *agentsv1alpha1
...
+	// actually act in the same loop. If PauseTime is nil or still in the
+	// future, we must proceed with deletion.
+	if _, hasRetention := box.Annotations[agentsv1alpha1.AnnotationReservePausedSandboxFor]; hasRetention &&
+		!box.Spec.Paused &&
+		pauseTimeReached(box.Spec.PauseTime, now) {
+		return ctrl.Result{}, false, nil
+	}
+
+	klog.InfoS("sandbox shutdown time reached, deleting", "sandbox", klog.KObj(box), "shutdownTime", box.Spec.ShutdownTime)
+	return ctrl.Result{}, true, r.Delete(ctx, box)
```
furykerry: the first return param is always ctrl.Result, consider remove it

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -558,35 +561,115 @@ func (r *SandboxReconciler) checkTimers(ctx context.Context, box *agentsv1alpha1
...
+	if result, done, err := r.handlePauseTimeout(ctx, box, now); done {
+		return result, true, err
+	}
+	return ctrl.Result{RequeueAfter: r.calcTimeoutRequeue(box, now)}, false, nil
+}
+
+// handlePauseTimeout triggers auto-pause when PauseTime has been reached.
+// It returns (result, true, err) when the caller should return immediately,
+// or (_, false, nil) when reconciliation should continue.
+func (r *SandboxReconciler) handlePauseTimeout(ctx context.Context, box *agentsv1alpha1.Sandbox, now metav1.Time) (ctrl.Result, bool, error) {
```
furykerry, line 652: the first return param is always ctrl.Result, consider remove it

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -558,35 +561,115 @@ func (r *SandboxReconciler) checkTimers(ctx context.Context, box *agentsv1alpha1
...
+		return retention, managed
+	}
+	raw := box.Annotations[agentsv1alpha1.AnnotationReservePausedSandboxFor]
+
+	klog.ErrorS(err, "invalid reserve paused sandbox annotation, using default",
+		"sandbox", klog.KObj(box),
+		"annotation", agentsv1alpha1.AnnotationReservePausedSandboxFor,
+		"value", raw)
+	if r.recorder != nil {
+		r.recorder.Eventf(box, corev1.EventTypeWarning, "InvalidReservePausedSandboxFor",
```
furykerry: if annotation format is incorrect, the controller will reconcile forever and keep emiting warning event, consider record the warning in a sandbox condition, or simply records warning in logs

pkg/pausedretention/retention.go
AiRanthem: Addressed locally: replaced the explicit default sentinel with forever. It still maps to the built-in 100-year retention window, and absent annotation remains policy-free for the controller path.

pkg/controller/sandbox/sandbox_controller.go
AiRanthem: Kept PauseTime aligned with ShutdownTime instead of clearing it. The next connect/resume path uses PauseTime to preserve auto-pause mode, so setting it nil would avoid re-triggering but lose that mode. Added a comment to make this explicit.

pkg/servers/e2b/pause_resume.go
AiRanthem: Addressed locally: replaced the hand-written pointer copy with ptr.To(timeout.ReservePausedSandboxForForeverValue).

api/v1alpha1/annotations.go
AiRanthem: Addressed locally: renamed the internal annotation, E2B metadata key, header, docs, and tests to reserve-paused-sandbox-duration / x-e2b-kruise-reserve-paused-sandbox-duration.

pkg/controller/sandbox/sandbox_controller.go
AiRanthem: Addressed locally for handleShutdownTimeout: removed the redundant ctrl.Result return and made the caller supply ctrl.Result{}.

pkg/controller/sandbox/sandbox_controller.go
AiRanthem: Kept this one because handlePauseTimeout can return ctrl.Result{Requeue: true} on patch conflict. I only removed the redundant ctrl.Result from handleShutdownTimeout.

pkg/controller/sandbox/sandbox_controller.go
AiRanthem: Addressed locally: invalid paused-retention annotations now only log and fail open to the default retention. Removed the InvalidReservePausedSandboxFor warning event and the related test assertions.

PR #567, opened 2026-06-23
fix: register scaleUpExpectation before creating sandbox

ashnaaseth2325-oss:
## SUMMARY

This PR fixes a race condition in the SandboxSet controller where scale-up expectations could be registered after a sandbox creation event was already observed. In certain timing scenarios, this left an unresolved expectation that blocked further scale-up operations until the expectation timeout expired.

The change updates `pkg/controller/sandboxset/sandboxset_controller.go` by aligning `createSandbox()` with the expectation handling pattern already used elsewhere in the codebase.

## FIX

```go
// Before
if err := r.Create(ctx, sbx); err != nil {
    return nil, err
}
scaleUpExpectation.ExpectScale(GetControllerKey(sbs), expectations.Create, sbx.Name)

// After
scaleUpExpectation.ExpectScale(GetControllerKey(sbs), expectations.Create, sbx.Name)
if err := r.Create(ctx, sbx); err != nil {
    scaleUpExpectation.ObserveScale(GetControllerKey(sbs), expectations.Create, sbx.Name)
    return nil, err
}
```

### DIFF
--- a/pkg/controller/sandboxset/sandboxset_controller.go
+++ b/pkg/controller/sandboxset/sandboxset_controller.go
@@ -383,12 +383,13 @@ func (r *Reconciler) createSandbox(ctx context.Context, sbs *agentsv1alpha1.Sand
 	if err := ctrl.SetControllerReference(sbs, sbx, r.Scheme); err != nil {
 		return nil, err
 	}
+	scaleUpExpectation.ExpectScale(GetControllerKey(sbs), expectations.Create, sbx.Name)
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
After investigating the CI and E2E failures, the proposed change introduces unresolved create expectations (`createDirty=1`) that cause replica overcounting and reconciliation stalls. The current implementation appears to rely on different expectation semantics than initially assumed, so I'm closing this PR until the expectation lifecycle can be analyzed further.

PR #568, opened 2026-06-23
feat: add RRSA storage authentication support for on-demand CSI mounts

BH4AWS:
Add credential provider fields (CredProviderName, RoleName) to CSIMountConfig and implement the full storage-auth annotation injection pipeline, enabling pod-level RRSA (RAM Roles for Service Accounts) authentication for CSI storage mounts without traditional Secrets.

Key changes:
- api/v1alpha1: add CredProviderName and RoleName fields to CSIMountConfig
- pkg/identity: define AnnotationStorageAuth constant and StorageAuthItem type; propagate storage-auth annotation to sandbox token metadata
- pkg/utils/csiutils: skip Secret lookup when volumeAttributes contains "authType: agent-identity"; add BuildStorageAuthItems() with pluggable StorageAuthItemEnricher hook for inner extension
- pkg/controller/sandboxclaim/core: inject storage-auth annotation via Modifier closure in buildClaimOptions when CredProviderName is set
- pkg/servers/e2b: inject storage-auth annotation via Modifier closure in createSandboxWithClaim when CredProviderName is set

Tests:
- storages_provider_test.go: 11 new cases covering BuildStorageAuthItems, agent-identity Secret skip, and CredProviderName passthrough
- sandbox_token_helper_test.go: 5 new cases for storage-auth annotation propagation to token metadata
- common_control_test.go: 2 new cases for Modifier injection behavior
- create_test.go: 1 new case for CredProviderName request structure


pkg/identity/sandbox_token_helper.go
```diff
@@ -71,6 +71,13 @@ func IssueSandboxToken(ctx context.Context, sbx *agentsv1alpha1.Sandbox) (*Token
 			metadata[k] = v
 		}
 	}
+	// Allow enterprise implementations to extract additional security metadata
+	// (e.g., storage-auth annotations for RRSA-based credential scoping) via hook.
+	if ExtractStorageAuthMetadata != nil {
```
furykerry: plz do not extract metadata here, pass the sandbox object to IssueToken of various token provider

PR #569, opened 2026-06-24
fix: handle edge cases in sandbox reuse lifecycle

zmberg:
## Summary
- Fail reuse immediately when Pod enters terminal phase (Succeeded/Failed) instead of waiting for timeout
- Add polling-based requeue during reuse in-progress to guarantee timeout fires even without external events
- Prioritize reuse trigger over Pod terminal detection in Running phase to prevent sandbox stuck in Failed with dirty claim metadata
- Preserve `AnnotationReuseEnabled` when clearing internal annotations in SandboxSet
- Add `ReuseCount` printer column to Sandbox CRD

## Test plan
- [x] Unit tests added for terminal phase detection (PodSucceeded/PodFailed)
- [x] Existing reuse tests updated for polling requeue behavior

### DIFF
--- a/pkg/controller/sandbox/core/reuse.go
+++ b/pkg/controller/sandbox/core/reuse.go
@@ -93,22 +93,9 @@ func (r *SandboxReuseControl) ensureSandboxReused(ctx context.Context, args Ensu
 func (r *SandboxReuseControl) doReuse(ctx context.Context, args EnsureFuncArgs) (time.Duration, error) {
 	box, newStatus := args.Box, args.NewStatus
 
-	poolName := box.Labels[agentsv1alpha1.LabelSandboxPool]
-	if poolName == "" {
-		return 0, fmt.Errorf("sandbox %s has no sandbox-pool label", box.Name)
-	}
-	sbs := &agentsv1alpha1.SandboxSet{}
-	if err := r.client.Get(ctx, client.ObjectKey{Namespace: box.Namespace, Name: poolName}, sbs); err != nil {
-		if apierrors.IsNotFound(err) {
-			return 0, fmt.Errorf("SandboxSet %s not found", poolName)
-		}
-		return 0, &RetriableError{Err: fmt.Errorf("failed to get SandboxSet %s: %w", poolName, err)}
-	}
-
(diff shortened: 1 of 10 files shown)


pkg/controller/sandbox/core/reuse.go
```diff
@@ -93,22 +93,9 @@ func (r *SandboxReuseControl) ensureSandboxReused(ctx context.Context, args Ensu
...
-		return 0, &RetriableError{Err: fmt.Errorf("failed to get SandboxSet %s: %w", poolName, err)}
-	}
-
-	// A nil Pod during reuse is an abnormal state — the sandbox should always
-	// have a running Pod. Fail immediately rather than waiting indefinitely.
-	if args.Pod == nil {
-		return 0, fmt.Errorf("pod not found during reuse")
+	sbs, err := r.validateReusePreconditions(ctx, args)
+	if err != nil {
+		return 0, err
```
furykerry, line 98: if validation fail, shall we delete the sandbox directly ?

pkg/controller/sandbox/core/reuse.go
```diff
@@ -151,14 +137,63 @@ func (r *SandboxReuseControl) doReuse(ctx context.Context, args EnsureFuncArgs)
...
+			box.Labels[agentsv1alpha1.LabelTemplateHash], sbs.Name, sbs.Status.UpdateRevision)
+	}
+
+	// A nil Pod during reuse is an abnormal state — the sandbox should always
+	// have a running Pod. Fail immediately rather than waiting indefinitely.
+	if args.Pod == nil {
+		return nil, fmt.Errorf("pod not found during reuse")
+	}
+
+	// If the Pod has already entered a terminal phase (Succeeded/Failed), the
```
furykerry, line 177: shall we check the following condition 
1. whether sandbox is paused ?
2. whether sandbox has pvc ?

pkg/controller/sandbox/core/reuse.go
zmberg: yes, handleReuseFailed will handle that uniformly.

PR #572, opened 2026-06-24
fix: nil guard SandboxConditionPaused in calculateStatus

ashnaaseth2325-oss:
## SUMMARY

This PR adds nil checks before accessing `SandboxConditionPaused` in the Sandbox controller's status calculation logic. It prevents a potential nil pointer dereference when a Sandbox is in the `Paused` phase but the corresponding paused condition is absent.

The change is limited to `pkg/controller/sandbox/sandbox_controller.go`, specifically the `calculateStatus()` logic for the `SandboxPaused` state.

## FIX

```go
// Before
if cond.Status == metav1.ConditionTrue && !box.Spec.Paused {
    ...
} else if !box.Spec.Paused && cond.Status == metav1.ConditionFalse {
    ...
}

// After
if cond != nil && cond.Status == metav1.ConditionTrue && !box.Spec.Paused {
    ...
} else if cond != nil && !box.Spec.Paused && cond.Status == metav1.ConditionFalse {
    ...
}
```

### DIFF
--- a/pkg/controller/sandbox/sandbox_controller.go
+++ b/pkg/controller/sandbox/sandbox_controller.go
@@ -510,7 +510,7 @@ func (r *SandboxReconciler) calculateStatus(ctx context.Context, args core.Ensur
 
 		cond := utils.GetSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionPaused))
 		// sandbox will only enter the resuming state after successful paused
-		if cond.Status == metav1.ConditionTrue && !box.Spec.Paused {
+		if cond != nil && cond.Status == metav1.ConditionTrue && !box.Spec.Paused {
 			// delete paused condition
 			utils.RemoveSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionPaused))
 			newStatus.Phase = agentsv1alpha1.SandboxResuming
@@ -521,7 +521,7 @@ func (r *SandboxReconciler) calculateStatus(ctx context.Context, args core.Ensur
 				LastTransitionTime: metav1.Now(),
 			}
 			utils.SetSandboxCondition(newStatus, rCond)
-		} else if !box.Spec.Paused && cond.Status == metav1.ConditionFalse {
+		} else if cond != nil && !box.Spec.Paused && cond.Status == metav1.ConditionFalse {
 			klog.InfoS("sandbox pause not completed, cannot enter resume state temporarily", "sandbox", klog.KObj(box))
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -396,7 +396,7 @@ func (r *SandboxReconciler) calculateStatus(ctx context.Context, args core.Ensur
 	case agentsv1alpha1.SandboxPaused:
 		cond := utils.GetSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionPaused))
 		// sandbox will only enter the resuming state after successful paused
-		if cond.Status == metav1.ConditionTrue && !box.Spec.Paused {
+		if cond != nil && cond.Status == metav1.ConditionTrue && !box.Spec.Paused {
```
zmberg: **[Issue]** The PR correctly adds `cond != nil` defensive checks to prevent nil pointer panics. However, when `cond == nil`, neither branch is entered and `calculateStatus` silently skips processing. This edge case lacks observability.

Suggest adding a `cond == nil` branch with a log message (e.g., `klog.InfoS` recording that the sandbox is in `Paused` phase but the paused condition is absent) to make this scenario traceable in production. The existing `EnsureSandboxPaused` mechanism will naturally handle condition initialization, so no control flow change is needed.

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -407,7 +407,7 @@ func (r *SandboxReconciler) calculateStatus(ctx context.Context, args core.Ensur
 				LastTransitionTime: metav1.Now(),
 			}
 			utils.SetSandboxCondition(newStatus, rCond)
-		} else if !box.Spec.Paused && cond.Status == metav1.ConditionFalse {
+		} else if cond != nil && !box.Spec.Paused && cond.Status == metav1.ConditionFalse {
 			klog.InfoS("sandbox pause not completed, cannot enter resume state temporarily", "sandbox", klog.KObj(box))
 		}
```
zmberg: **[Suggestion]** Suggest adding test cases to cover the `cond == nil` scenario — verifying the controller's behavior when the `SandboxPaused` phase lacks the paused condition — to prevent regressions. Existing `Paused`-related test cases in `TestCalculateStatus` can serve as reference.

zmberg (commented):
## Code Review

Thanks for the fix! The nil guard correctly prevents the panic. A couple of suggestions below:

- **Issue**: 1
- **Suggestion**: 1

PR #573, opened 2026-06-24
fix: avoid misleading wait hook change logs

### LINKED ISSUE #571: Misleading "object with wait hook changed" log in WaitReconciler
### Bug Description
In the `WaitReconciler[T client.Object]` implementation, the log message `"object with wait hook changed"` is printed unconditionally in the `Reconcile` method every time the watched object changes, regardless of whether a wait hook actually exists for that object.

### Affected Code
`pkg/cache/controllers/cache_controllers.go`:
```go
func (r *WaitReconciler[T]) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
    // ...
    log.V(utils.DebugLogLevel).Info("object with wait hook changed", "resourceVersion", obj.GetResourceVersion())
    r.checkWaitHooks(waitKey, obj)
    return ctrl.Result{}, nil
}
```

### Impact
When `CacheSandboxWaitReconciler` is registered to watch `Sandbox` resources, any state update or sync to *any* sandbox triggers this reconciler. The log will spam `"object with wait hook changed"` even if `loadWaitHook` inside `checkWaitHooks` returns `false`. This misleads users into thinking that the sandbox is stuck in a wait job or hook, when in fact it's just a routine status update of the object.

### Suggested Fix
Move the log message inside `checkWaitHooks` or conditionally check if the hook exists before logging:
```go
func (r *WaitReconciler[T]) checkWaitHooks(key string, obj T) {
	entry, ok := r.loadWaitHook(key)
	if !ok {
		return
	}
	// Log here when we actually know the hook exists
	// klog.FromContext(ctx).V(utils.DebugLogLevel).Info("object with wait hook changed", "resourceVersion", obj.GetResourceVersion())
	
	satisfied, err := entry.Check(obj)
	if satisfied || err != nil {
		entry.Close()
	}
}
```

Jayant-kernel:

This PR avoids a misleading debug log in the generic `WaitReconciler`.

Previously, `WaitReconciler.Reconcile` logged `object with wait hook changed` for every observed object update, even when no wait hook existed for that object. The log now happens only after the matching wait hook is found, immediately before the hook check runs.

A regression test covers both cases:

- no matching wait hook: no `object with wait hook changed` log is emitted;
- matching wait hook: the log is emitted once.

Fixes #571

The focused test is:

```bash
go test -count=1 ./pkg/cache/controllers -run 'TestCacheSandboxWaitReconciler_(Reconcile|LogsObjectChangedOnlyWithWaitHook)$'
```

I could not run the Go test locally because the local Windows environment blocks the available Go binary from executing. The change is limited to the wait-hook log path and its regression test; GitHub CI should run the package tests.

No API, CRD, generated code, or behavior outside debug logging is changed.

### DIFF
--- a/pkg/cache/controllers/cache_controller_sandbox_wait_test.go
+++ b/pkg/cache/controllers/cache_controller_sandbox_wait_test.go
@@ -21,11 +21,13 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/go-logr/logr"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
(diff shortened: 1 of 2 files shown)


Jayant-kernel:
@AiRanthem  review my pr

AiRanthem:
@Jayant-kernel plz fix your UT

AiRanthem:
Your unit tests are too messy. Please keep AI-generated code concise and high-quality.

Jayant-kernel:
@AiRanthem 
Sorry for the mess , i cleaned the unit test

PR #574, opened 2026-06-24
fix: fence sandbox creates after api key deletion

### LINKED ISSUE #570: [BUG] Delete API-key can leave inaccessible sandboxes from in-flight creates
## Problem

Deleting an API key does not fully fence create requests that already passed `CheckApiKey` before the delete completed.

A request can pass authentication with the soon-to-be-deleted key, continue through sandbox creation, and persist a sandbox whose owner annotation is still the deleted API-key ID. After the key is deleted, normal API-key authentication can no longer load that key, so the created sandbox is inaccessible through the E2B API-key path and cannot be managed by that key.

There is also quota cleanup drift around this race. `DeleteAPIKey` removes the key from storage and then starts quota cleanup for that key. A concurrent in-flight create can acquire quota after cleanup, which can write Redis live-set and usage residue for the deleted key. Anti-drift enumerates limited keys from key storage, so after deletion it no longer lists this key and will not visit that residue again.

## Impact

- Sandbox resource leak: a sandbox can remain owned by a deleted API key and be unreachable through normal API-key access.
- Redis dead memory: quota live-set and usage entries for the deleted key can be recreated after cleanup.
- Other API keys are not affected for quota calculation because quota accounting is keyed by API-key ID.

## Follow-up directions

Possible fixes include:

- add a tombstone or short TTL deny-list for deleted API-key IDs and reject or cleanup late creates;
- drain or deny in-flight create requests after API-key deletion starts;
- redefine API-key delete semantics so created sandboxes remain administratively discoverable and cleanup can converge;
- teach quota cleanup or anti-drift to handle deleted-key residue for a bounded period.

Jayant-kernel:

This PR prevents sandbox create requests from leaving resources owned by an API key that has just been deleted.

The change adds a short-lived deleted API-key fence in the E2B controller. When an API key is deleted, its ID is recorded in the fence. `CreateSandbox` checks that fence before creating a sandbox and checks it again after claim/clone returns. If the key is deleted while creation is in flight, the newly created sandbox is cleaned up and the request returns unauthorized instead of leaving an inaccessible sandbox behind.

The tests cover:

- recording the deleted API-key ID after a successful delete;
- rejecting create requests for a tombstoned API key before any sandbox is created; and
- cleaning up a sandbox when the API key is deleted while create is in progress.

Fixes #570

- `gofmt` on the changed Go files
- `git diff --check`

The focused Go test command could not be run locally because the local Go executable is blocked by the machine's application-control policy. The added tests are intended to run in repository CI:

```bash
go test ./pkg/servers/e2b -run 'Test(DeleteAPIKeyPermissionMiddleware|CreateSandboxRejectsDeletedAPIKeyTombstone|CreateSandboxCleansUpWhenAPIKeyDeletedDuringCreate)$'
```

The fence is intentionally bounded in memory. It closes the local race window around deletion and late create completion without changing persistent API-key storage semantics.

### DIFF
--- a/pkg/servers/e2b/api_key.go
+++ b/pkg/servers/e2b/api_key.go
@@ -168,6 +168,7 @@ func (sc *Controller) DeleteAPIKey(r *http.Request) (web.ApiResponse[struct{}], 
 				Message: fmt.Sprintf("Failed to delete API key: %v", err),
 			}
 		}
+		sc.deletedAPIKeys.Add(key.ID.String())
 		if key.QuotaSpec != nil && key.QuotaSpec.IsLimited() {
 			// Deleted-key quota cleanup is bounded best-effort. If Redis stays
 			// unavailable past the retry window, the leftover q:live/q:sum keys become
--- a/pkg/servers/e2b/api_key_test.go
+++ b/pkg/servers/e2b/api_key_test.go
@@ -660,6 +660,7 @@ func TestDeleteAPIKeyPermissionMiddleware(t *testing.T) {
 			}
 			require.Nil(t, apiError)
 			assert.Equal(t, tt.expectCode, resp.Code)
+			assert.True(t, controller.deletedAPIKeys.Contains(tt.targetID))
 		})
 	}
 }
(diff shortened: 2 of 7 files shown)


Jayant-kernel:
@AiRanthem
review my pr

PR #585, opened 2026-06-28
feat(gateway): implement wake-on-traffic for paused sandboxes

furykerry:

Implement **wake-on-traffic** for paused sandboxes: the sandbox-gateway automatically resumes a paused sandbox when inbound traffic arrives, by calling `sandboxcr.Sandbox.Resume()` directly — no HTTP call to sandbox-manager, no systemtoken.

**How it works:**

1. At sandbox creation (E2B `CreateSandbox`), when `autoResume=true` + `autoPause=true`, two annotations are set on the Sandbox CR: `agents.kruise.io/wake-on-traffic=true` and `agents.kruise.io/wake-timeout-seconds=<N>`.
2. The route registry propagates `WakeOnTraffic` from the annotation into the `Route` struct.
3. When the Envoy Go filter receives traffic for a paused sandbox with `WakeOnTraffic=true`, it invokes the `Waker` inline. The Waker calls `sandboxcr.AsSandbox(sbx, cache).Resume()`, which reuses the entire connect-path (spec patch, wait-for-running, concurrent dedup, conflict retries), then syncs the route to the local registry and peer gateways.
4. After wake, the filter re-reads the registry and forwards the request normally.

**Key design decisions:**

- Reuses `sandboxcr.Sandbox.Resume()` — no reimplementation of spec patching, wait, or dedup.
- `ShutdownTime` is preserved during wake to prevent timed sandboxes from losing their auto-delete deadline.
- `atomic.Pointer[Waker]` for race-safe singleton access.
- `SyncRouteWithPeers` extracted as a package-level function so the gateway can sync routes without a full `proxy.Server` instance.
- Feature is opt-in: gated on both filter config (`enable-wake-on-traffic`) and per-sandbox annotation.

NONE

1. Run unit tests:
   ```
   go test ./pkg/sandbox-gateway/wake/ ./pkg/sandbox-gateway/filter/ ./pkg/utils/proxyutils/ ./pkg/proxy/
   ```
2. Deploy sandbox-gateway with `enable-wake-on-traffic: true` in the Envoy filter config.
3. Create a sandbox with `autoPause=true, autoResume=true`.
4. Wait for the sandbox to auto-pause (or pause it manually).
5. Send traffic through the gateway — the sandbox should be automatically resumed and the request forwarded.
6. Run the E2E test: `pytest test/e2b/test_wake_on_traffic.py`

- **`pkg/sandbox-gateway/wake/wake.go`** — the core of the feature. Delegates entirely to `sandboxcr.Sandbox.Resume()`. Preserves `ShutdownTime` from the informer cache to avoid resource leaks (timed sandboxes losing their auto-delete deadline).
- **`pkg/sandbox-gateway/filter/filter.go`** — the wake is triggered inline in `DecodeHeaders`, blocking the Envoy worker thread. The timeout is configurable via `wake-timeout-seconds` (default 60s).
- **`pkg/proxy/routes.go`** — `SyncRouteWithPeers` extracted from a `Server` method to a package-level function taking `peers.Peers` and `Route`. The original method delegates to it.
- **RBAC** — the gateway's ClusterRole already had `update`/`patch` on `sandboxes`, which is what `Resume()`'s `retryUpdate` needs. Only `sandboxsets` read access was added.

### DIFF
--- a/api/v1alpha1/annotations.go
+++ b/api/v1alpha1/annotations.go
@@ -27,6 +27,16 @@ const (
 	// no Sandbox or Checkpoint still references it before performing the actual
 	// deletion.
 	AnnotationCleanupCandidate = InternalPrefix + "cleanup-candidate"
+
+	// AnnotationWakeOnTraffic enables wake-on-traffic for a paused sandbox.
+	// When set to "true", the sandbox-gateway will attempt to resume the sandbox
+	// by patching Spec.Paused=false when traffic arrives.
+	AnnotationWakeOnTraffic = InternalPrefix + "wake-on-traffic"
+
+	// AnnotationWakeTimeoutSeconds stores the auto-pause timeout (in seconds) to
+	// apply when the sandbox is woken by traffic. The gateway reads this to set
+	// ResumeOptions.Timeout.PauseTime, re-arming auto-pause after wake.
+	AnnotationWakeTimeoutSeconds = InternalPrefix + "wake-timeout-seconds"
 )
 
 // E2B annotations
--- a/config/sandbox-gateway/configmap.yaml
+++ b/config/sandbox-gateway/configmap.yaml
@@ -80,6 +80,8 @@ data:
                               sandbox-port-header: "e2b-sandbox-port"
                               default-port: "49983"
                               enable-auth: false
+                              enable-wake-on-traffic: true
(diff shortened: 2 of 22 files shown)


furykerry:
Recreating PR with squashed commits for a cleaner history.

PR #586, opened 2026-06-28
feat(gateway): implement wake-on-traffic for paused sandboxes

furykerry:

This PR implements **wake-on-traffic** for paused sandboxes in the sandbox-gateway. When a paused sandbox receives incoming traffic, the gateway automatically resumes it and forwards the suspended request once the sandbox is running again. Design: `docs/proposals/20260627-wake-on-traffic-via-spec-patch.md`.

**Key changes:**

1. **Wake-on-traffic in filter** (`pkg/sandbox-gateway/filter/filter.go`): When a paused sandbox with wake-on-traffic enabled receives traffic, the filter launches wake in a goroutine with a detached context whose deadline it derives from the filter config, and returns `api.Running` to suspend Envoy request processing. `wakeAndContinue` calls `Continue(api.Continue)` on success or a generic `SendLocalReply(503)` on failure (error detail is logged server-side only, never echoed to callers). A mutex-protected completion state machine prevents double-reply when the async goroutine and `Destroy` race. Wake eligibility is encapsulated in `shouldWakeSandbox()`, which first enforces a **UID fence**: the route's sandbox UID must match the informer's sandbox UID, failing closed on a missing object or mismatch, so a stale route authenticated against a deleted sandbox can never wake or reach a same-name recreated sandbox. The normal Running path and the async wake completion path share `applyUpstreamOverrides()`, so the request that triggered the wake gets identical `envoy.lb.original_dst`, runtime mTLS metadata, and route-cache handling as a request that arrived on a Running sandbox.

2. **New `wake` package** (`pkg/sandbox-gateway/wake/`): `Waker.Wake()` delegates to the sandbox-manager infra `sandboxcr.Sandbox.Resume()` and then syncs the route locally and to peer gateways. The caller owns the wait deadline; `defaultWakeTimeout` is only the fallback timeout used when the sandbox carries no `wake-timeout-seconds` annotation, and `Wake` rejects a non-positive value explicitly. The wake path mirrors the E2B timeout-mode logic: only auto-pause sandboxes get a fresh `PauseTime`, and their `ShutdownTime` is reset to `now + ForeverReservePausedSandboxDuration` (100y) so a retained stale/soon-to-expire ShutdownTime cannot auto-delete a sandbox that was just woken by traffic — the next pause recomputes ShutdownTime from the paused-retention policy. Never-timeout and shutdown-only sandboxes keep their semantics. Dedup comes from Resume's first-writer-wins retry plus refcounted `NewSandboxResumeTask`, so no singleflight is needed.

3. **Resume persists via optimistic merge patch** (`pkg/sandbox-manager/infra/sandboxcr/sandbox.go`): All `Resume` operations now persist via `retryPatch` using `MergeFromWithOptimisticLock` instead of a full Update. The gateway informer cache strips `ManagedFields` and `spec.embeddedSandboxTemplate` (see #724), so a full Update of the cached object would have erased the embedded template on the API server; the patch body carries only the fields Resume changes plus the resourceVersion precondition.

4. **Shared resume timeout floor** (`pkg/utils/timeout/timeout.go`): `ApplyResumeTimeoutFloor` + `DefaultMinResumeTimeoutSeconds` (300s) floor the timeout carried by a Resume of a paused sandbox so the fresh deadline cannot expire mid-resume and re-pause the sandbox. The E2B Connect/Resume handlers reuse it with their tunable `--e2b-min-resume-timeout` value (behavior unchanged), and the gateway wake path applies the same floor to the `wake-timeout-seconds` value (the create API allows values as short as 30s).

5. **autoResume API support** (`pkg/servers/e2b/models/sandbox.go`, `pkg/servers/e2b/create.go`): `"autoResume": {"enabled": true}` in the E2B create request stamps `agents.kruise.io/wake-on-traffic=true` at creation time (the E2B SDK maps `lifecycle={"auto_resume": True}` to this field). The `wake-timeout-seconds` annotation is only stamped when the request also configures auto-pause — for shutdown-only lifecycles the gateway wake path never uses it, so it would be misleading metadata. Create also deletes any inherited wake annotations before merging template metadata, and both wake annotations are added to `AnnotationsClearedOnRecycle`, so every delivery's wake policy is decided solely by the current request — nothing leaks from a previous delivery on a recycled CR.

6. **New annotations** (`api/v1alpha1/annotations.go`): `agents.kruise.io/wake-on-traffic` and `agents.kruise.io/wake-timeout-seconds`.

7. **Route propagation** (`pkg/sandboxroute/`, `pkg/sandbox-gateway/server/server.go`, `pkg/proxy/routes.go`): `Route` gains `WakeOnTraffic`, derived from the annotation in `RouteFromSandbox`; the route's existing `UID` field is what the filter's UID fence compares against the informer object. `handleRefresh` retains routes for all states except `dead` and empty string so the filter has full visibility for wake decisions. `SyncRouteWithPeers` becomes a package-level function so the gateway Waker can sync routes without constructing a full `proxy.Server`.

8. **Gateway cache sandbox-only mode** (`pkg/cache/`): `NewCache`/`SetupCacheControllersWithManager`/`AddIndexesToCache` take a `sandboxOnly` flag; when true only Sandbox-related controllers and indexes are registered (Checkpoint, PVC, and SandboxSet are skipped), reducing the gateway's memory usage and API server load.

9. **Configuration** (`pkg/sandbox-gateway/filter/config.go`, `config/sandbox-gateway/configmap.yaml`): filter config fields `enable-wake-on-traffic` (bool) and `wake-timeout-seconds` (int, default 60). Wake configuration is **listener-level only**: per-route overrides are not supported and this is documented on both fields and in `Merge` (a per-route config can neither disable wake nor unintentionally reset the listener timeout).

10. **E2B health route** (`pkg/servers/e2b/routes.go`): `GET /health` is also registered under the `/kruise/api` prefix so health checks forwarded through the sandbox-gateway work.

11. **New E2E test** (`test/e2b/test_wake_on_traffic.py`): creates a sandbox with `lifecycle={"on_timeout": "pause", "auto_resume": True}`, pauses it, and wakes it with **exactly one request** whose client timeout covers the full wake budget — the filter holds that very request while the sandbox resumes and continues it when Ready, so the response proves both the wake and the completion path. The response is asserted to not be any gateway local reply (fingerprinted by the filter's local-reply bodies) and not an envd 401, ruling out success from unrelated retries. The sandbox CR is resolved via the `agents.kruise.io/sandbox-id` label so pool-claimed sandboxes work. Wired into `hack/run-e2b-e2e-test.sh`.

NONE

1. Run unit tests:
   ```bash
   go test ./pkg/sandbox-gateway/... ./pkg/cache/... ./pkg/servers/e2b/... ./pkg/utils/timeout/... ./pkg/sandbox-manager/infra/sandboxcr/...
   ```
2. Run E2E test:
   ```bash
   pytest test/e2b/test_wake_on_traffic.py
   ```
3. Manual verification:
   - Create a sandbox with `lifecycle={"on_timeout": "pause", "auto_resume": True}` (SDK) or `"autoResume": {"enabled": true}` (raw API)
   - Verify the sandbox is annotated with `agents.kruise.io/wake-on-traffic=true`
   - Pause it via `POST /sandboxes/{id}/pause`
   - Send traffic to the paused sandbox
   - Verify the sandbox is automatically resumed and the request is forwarded

- **Async wake**: `DecodeHeaders` returns `api.Running` when wake is triggered, launching `wakeAndContinue` in a goroutine with a detached context whose deadline is the sole wake deadline (derived from `wake-timeout-seconds` filter config). On completion it calls `Continue(api.Continue)` or `SendLocalReply`; `Destroy()` cancels the in-flight wake. A mutex-protected completion state machine prevents double-reply races. Wake failure replies stay generic — Kubernetes API error strings are logged server-side, not returned to callers.
- **UID fence**: `shouldWakeSandbox` checks `Waker.SandboxUIDMatches` before any wake decision; a route whose recorded sandbox UID does not match the informer object (or a missing object / empty UID) fails closed. This blocks the attack where a stale route, authenticated against a deleted sandbox, wakes or accesses a same-name recreated sandbox.
- **No singleflight**: dedup is handled by Resume's first-writer-wins `retryPatch` and refcounted wait via `NewSandboxResumeTask`, avoiding the footgun where a canceled caller context would cut short a shared Resume for all coalesced callers.
- **Optimistic patch vs gateway cache stripping**: since #724 strips `spec.embeddedSandboxTemplate` from the gateway informer cache, Resume persists via `MergeFromWithOptimisticLock` patch; a full Update would have erased the template. First-writer-wins, conflict retry, trace injection, and resourceVersion expectations are preserved.
- **Resume timeout floor**: the gateway wake path reuses the same floor (default 300s) as E2B Connect/Resume so a short `wake-timeout-seconds` (create API minimum is 30s) cannot expire mid-resume and trigger a controller re-pause.
- **Timeout-mode preservation during wake**: only auto-pause sandboxes receive a fresh `PauseTime`; never-timeout and shutdown-only sandboxes are resumed without injecting one. Auto-pause wake additionally resets `ShutdownTime` to the forever retention horizon (100y) so a retained ShutdownTime cannot delete a just-woken sandbox; the next pause recomputes it per the paused-retention policy.
- **Wake annotation lifecycle**: wake-on-traffic can be enabled via the E2B create API (`autoResume.enabled`) or manually via `kubectl annotate`; `wake-timeout-seconds` is only stamped by the create API for auto-pause sandboxes. Create strips inherited wake annotations before merging template metadata, and recycle clears both annotations (`AnnotationsClearedOnRecycle`), so a recycled CR never carries the previous delivery's wake policy. The E2E test verifies both annotations persist across the wake.
- **Listener-level config**: wake config is not supported at the per-route level (`typed_per_filter_config`); `Merge` only lets an explicit true / positive value override the listener value. This is documented in `config.go` rather than fixed, matching deployments that only render `plugin_config`.
- **handleRefresh**: only `dead` and empty-state routes are deleted from the registry; all other states are retained for full filter visibility.

### Ⅴ. Test coverage

Measured with `go test -cover` on the final diff:

| Package | Coverage |
|---------|----------|
| `pkg/sandbox-gateway/filter` | 88.8% |
| `pkg/sandbox-gateway/wake` | 77.8% |
| `pkg/servers/e2b` | 88.5% |
| `pkg/proxy` | 89.7% |

### DIFF
--- a/config/sandbox-gateway/rbac.yaml
+++ b/config/sandbox-gateway/rbac.yaml
@@ -8,7 +8,7 @@ metadata:
 rules:
   - apiGroups: ["agents.kruise.io"]
     resources: ["sandboxes"]
-    verbs: ["get", "list", "watch"]
+    verbs: ["get", "list", "watch", "update", "patch"]
   - apiGroups: ["agents.kruise.io"]
     resources: ["sandboxes/status"]
     verbs: ["get"]
--- a/docs/proposals/20260627-wake-on-traffic-via-spec-patch.md
+++ b/docs/proposals/20260627-wake-on-traffic-via-spec-patch.md
@@ -1,0 +1,105 @@
+# Wake-on-Traffic via Direct Spec Patch (No SystemToken)
+
+## Motivation
+
+PR #495 implements wake-on-traffic by having sandbox-gateway call sandbox-manager's
+`/connect` HTTP API using a scoped system credential (systemtoken). This introduces
+cross-component HTTP coupling, a new credential type, and credential lifecycle
+management. This proposal replaces that mechanism with a direct Go function call,
+eliminating the systemtoken entirely while reusing the exact same resume logic.
+
+## Design
+
+The sandbox-gateway directly calls `sandboxcr.AsSandbox(sbx, cache).Resume(ctx, opts)`
+— a Go function call, not an HTTP call. This reuses the entire sandbox-manager connect
+path: spec patch (`Spec.Paused = false`), wait-for-running (`NewSandboxResumeTask().Wait()`),
+concurrent dedup (first-writer-wins via `retryUpdate`), conflict retries, and post-resume
+refresh (`InplaceRefresh`).
+
+```
+Traffic -> Envoy Filter -> Registry lookup
+  |-> sandbox Running: forward (current behavior)
+  |-> sandbox Paused + WakeOnTraffic enabled:
+       |-> sandboxcr.AsSandbox(sbx, cacheProvider).Resume(ctx, opts)
+       |     ^-- reuses existing sandbox-manager connect implementation:
+       |         1. refreshFromAPIReader (fresh fetch)
+       |         2. IsSandboxResumable check
+       |         3. NewSandboxResumeTask (pre-acquired wait task)
+       |         4. retryUpdate: patches Spec.Paused=false + setTimeout
+       |         5. resumeTask.Wait() -- blocks until Ready condition True
+       |         6. InplaceRefresh + expectations
+       |     Concurrent dedup: first-writer-wins via retryUpdate
+       |-> syncRoute: update local registry + sync to peer gateways
+       |-> Forward request or return 503 on timeout/failure
+  |-> sandbox Paused + WakeOnTraffic disabled: return 502 (current behavior)
+```
+
+## Key Components
+
+### Annotation Lifecycle
+
+- `AnnotationWakeOnTraffic` ("agents.kruise.io/wake-on-traffic"): Set to "true"
+  in one of two ways:
+  1. **Via the E2B create API**: pass `"autoResume": {"enabled": true}` in the
+     request body. The sandbox-manager's `basicSandboxCreateModifier` writes the
+     annotation at sandbox creation time.
+  2. **Manually via `kubectl annotate`** (or any out-of-band tooling) for
+     sandboxes that were created without `autoResume`.
+- `AnnotationWakeTimeoutSeconds` ("agents.kruise.io/wake-timeout-seconds"): Stores the
+  auto-pause timeout (in seconds) to apply when the sandbox is woken by traffic. The
+  gateway reads this to set `ResumeOptions.Timeout.PauseTime`, re-arming auto-pause
+  after wake.
+
+### Wake Package (`pkg/sandbox-gateway/wake/`)
+
+The `Waker` struct wraps `sandboxcr.AsSandbox(sbx, cache).Resume()` and syncs the route
+after Resume succeeds. It does NOT reimplement spec patching or wait-for-running — it
+delegates entirely to the existing `Resume()` method.
+
+After Resume succeeds, `syncRoute` mirrors the manager's `syncRoute` flow:
+1. Get route from refreshed sandbox (`sandbox.GetRoute()`)
+2. Update local gateway registry (`registry.GetRegistry().Update`)
+3. Sync route to peer gateways (`proxy.SyncRouteWithPeers`)
+
+### Cache Provider
+
+`cache.NewCache(mgr)` creates the same informer-backed cache + `WaitReconciler` used by
+sandbox-manager. The gateway's controller-runtime manager hosts both the existing
+`SandboxReconciler` (for local registry updates) and the cache provider's wait
+reconciler (for `NewSandboxResumeTask().Wait()`).
+
+### Route Sync
+
+`proxy.SyncRouteWithPeers` was extracted as a package-level function so the gateway can
+call it without creating a full `proxy.Server` instance. The gateway's `server.Server`
+exposes its `peerManager` via `GetPeerManager()` for use by the Waker.
+
+## Authorization
+
+K8s RBAC grants the gateway ServiceAccount `update`/`patch` on `sandboxes` resources.
+No systemtoken, no HTTP call to sandbox-manager.
+
+## Comparison with PR #495
+
+| Aspect | PR #495 | This Proposal |
+|--------|---------|---------------|
+| Wake trigger | Gateway calls manager `/connect` HTTP API | Gateway calls `sandboxcr.Sandbox.Resume()` directly (Go function) |
+| Auth | System key (systemtoken) | K8s RBAC (ServiceAccount) |
+| Resume logic | Manager's `ResumeSandbox` HTTP handler | Same `sandboxcr.Sandbox.Resume()` method (imported, not HTTP) |
+| Wait mechanism | Manager's `NewSandboxResumeTask().Wait()` | Same (reused via Resume call) |
+| Spec patching | Manager's `retryUpdate` inside Resume | Same (reused via Resume call) |
+| Concurrent dedup | Manager's first-writer-wins | Same (reused via Resume call) |
+| Route sync after wake | Manager's `syncRoute` | Gateway mirrors: `registry.Update` + `proxy.SyncRouteWithPeers` |
+| Timeout update | Manager's connect API handles it | Gateway passes `ResumeOptions.Timeout.PauseTime` |
+| New components | `systemkey.go`, `wake/client.go`, `wake/wake.go` | `wake/wake.go` (thin wrapper) |
+| Cross-component deps | Gateway -> Manager HTTP | Gateway imports `pkg/sandbox-manager/infra/sandboxcr` (Go import) |
+| Cache provider | Manager has its own | Gateway creates its own via `cache.NewCache(mgr)` |
+
+## Timeout Handling
+
+The `AnnotationWakeTimeoutSeconds` annotation stores the auto-pause timeout. When the
+gateway wakes a sandbox:
+1. It reads the annotation to determine `PauseTime` (re-arming auto-pause)
+2. If the annotation is absent, it uses the filter's `WakeTimeoutSeconds` config default
+3. The timeout is passed as `ResumeOptions.Timeout.PauseTime`, which is written atomically
+   with `Spec.Paused = false` inside `retryUpdate` — closing the auto-pause race
--- a/pkg/sandbox-gateway/filter/filter.go
+++ b/pkg/sandbox-gateway/filter/filter.go
@@ -292,7 +292,8 @@ func (f *sandboxFilter) verifierUnavailable(sandboxID string) api.StatusType {
 // shouldWakeSandbox determines whether a non-Running sandbox should be woken
 // by traffic. Returns true only when wake-on-traffic is enabled, the sandbox
 // is Paused, the waker is initialized, the route carries a full ObjectKey,
-// and either the route registry already has WakeOnTraffic set or the
+// the informer still holds a sandbox with the route's UID (stale-route
+// fence), and either the route registry already has WakeOnTraffic set or the
 // annotation fallback check succeeds.
 func (f *sandboxFilter) shouldWakeSandbox(route sandboxroute.Route, waker *wake.Waker) bool {
 	if route.State != agentsv1alpha1.SandboxStatePaused {
@@ -306,6 +307,19 @@ func (f *sandboxFilter) shouldWakeSandbox(route sandboxroute.Route, waker *wake.
 	}
 	key, ok := route.ObjectKey()
 	if !ok {
+		return false
+	}
+	// UID fence: the route registry can lag behind deletions, so the route
+	// may reference a sandbox that was deleted and recreated under the same
+	// name. The request was authenticated against the route's identity, so a
+	// stale route must neither wake the new object nor inherit its wake
+	// configuration. Verify the informer still holds the exact object.
+	if !waker.SandboxUIDMatches(context.Background(), key.Namespace, key.Name, route.UID) {
+		logger.Warn("Stale wake route: sandbox UID mismatch or object gone",
+			zap.String("sandboxID", route.ID),
+			zap.String("namespace", key.Namespace),
+			zap.String("name", key.Name),
+			zap.String("routeUID", string(route.UID)))
 		return false
 	}
 	// route.WakeOnTraffic is the primary check (fast, from registry).
--- a/pkg/sandbox-gateway/filter/filter.go
+++ b/pkg/sandbox-gateway/filter/filter.go
@@ -17,8 +17,13 @@ limitations under the License.
 package filter
 
 import (
+	"context"
 	"crypto/subtle"
+	"errors"
 	"fmt"
+	"strings"
+	"sync"
+	"time"
 
 	"github.com/envoyproxy/envoy/contrib/golang/common/go/api"
 	"go.uber.org/zap"
@@ -26,6 +31,7 @@ import (
 
 	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
 	"github.com/openkruise/agents/pkg/sandbox-gateway/registry"
+	"github.com/openkruise/agents/pkg/sandbox-gateway/wake"
 	"github.com/openkruise/agents/pkg/sandboxroute"
 	"github.com/openkruise/agents/pkg/servers/e2b/adapters"
 	"github.com/openkruise/agents/pkg/utils"
@@ -48,6 +54,12 @@ const (
 	runtimeMTLSMetadataKey       = "upstream-mtls"
 )
 
+// Compile-time assertion that sandboxFilter implements the api.StreamFilter interface.
+// This catches mismatches like implementing Destroy() (Config interface) instead of
+// OnDestroy(DestroyReason) (StreamFilter interface).
+var _ api.StreamFilter = (*sandboxFilter)(nil)
+
+// FilterFactory creates a new sandbox filter instance for each stream.
 func FilterFactory(c interface{}, callbacks api.FilterCallbackHandler) api.StreamFilter {
 	cfg := c.(*FilterConfig)
 	return &sandboxFilter{
@@ -64,6 +76,13 @@ type sandboxFilter struct {
 	config         *Config
 	adapter        *adapters.E2BAdapter
 	jwtAuthManager JWTAuthManager
+
+	// Async wake completion state. Protected by mu.
+	mu         sync.Mutex
+	completing bool // wakeAndContinue is setting metadata
+	completed  bool // Continue or SendLocalReply already called
+	destroyed  bool // Envoy destroyed the filter (stream gone)
+	cancel     context.CancelFunc
 }
 
 func (f *sandboxFilter) DecodeHeaders(header api.RequestHeaderMap, endStream bool) api.StatusType {
@@ -87,8 +106,9 @@ func (f *sandboxFilter) DecodeHeaders(header api.RequestHeaderMap, endStream boo
 		return api.Continue
 	}
 
-	logger.Debug("DecodeHeaders: adapter mapped request",
-		zap.String("sandboxID", sandboxID),
+	log := logger.With(zap.String("sandboxID", sandboxID))
+
+	log.Debug("DecodeHeaders: adapter mapped request",
 		zap.Int("sandboxPort", sandboxPort),
 		zap.Any("extraHeaders", extraHeaders))
 
@@ -111,7 +131,7 @@ func (f *sandboxFilter) DecodeHeaders(header api.RequestHeaderMap, endStream boo
 	}
 	route, ok := routeRegistry.Get(sandboxID)
 	if !ok {
-		logger.Warn("Sandbox not found in registry", zap.String("sandboxID", sandboxID))
+		log.Warn("Sandbox not found in registry")
 		f.callbacks.DecoderFilterCallbacks().SendLocalReply(
 			502,
 			"sandbox not found: "+sandboxID,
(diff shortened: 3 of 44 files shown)


pkg/sandbox-gateway/wake/wake.go
```diff
@@ -0,0 +1,178 @@
...
+func (w *Waker) Wake(ctx context.Context, namespace, name string, defaultWakeTimeout time.Duration) error {
+	key := namespace + "/" + name
+
+	// Use a detached context so the shared Resume is not tied to any single
+	// caller's context.  When the caller's ctx has a shorter deadline it can
+	// still bail out via DoChan + select.
+	wakeCtx, wakeCancel := context.WithTimeout(context.Background(), defaultWakeTimeout)
+	defer wakeCancel()
+
+	type result struct{ err error }
```
AiRanthem: The `result struct{ err error }` wrapper is redundant. `singleflight.Result` already carries an `Err` field, and the values returned by the `DoChan` func map directly to `Result.Val`/`Result.Err`. Wrapping the error into `Val` (with a nil func-error) and then type-asserting `r.Val.(result).err` is semantically identical to returning the error directly and reading `r.Err` — singleflight shares a returned error across all waiters just the same.

Suggested simplification:

```go
ch := w.flight.DoChan(key, func() (interface{}, error) {
    return nil, w.wakeInternal(wakeCtx, namespace, name, defaultWakeTimeout)
})

select {
case r := <-ch:
    return r.Err
case <-ctx.Done():
    return ctx.Err()
}
```

A payload struct would only be justified if `wakeInternal` needed to return business data (e.g. route/sandboxID) alongside the error, which is not the case here (`Wake` returns `error`).

pkg/sandbox-gateway/filter/filter.go
```diff
@@ -100,15 +114,61 @@ func (f *sandboxFilter) DecodeHeaders(header api.RequestHeaderMap, endStream boo
 	}
 
 	if route.State != agentsv1alpha1.SandboxStateRunning {
```
AiRanthem, line 136: The wake entry condition here is broader than the intent. `!= Running` covers `creating`/`available`/`paused`/`dead`, but only `paused` is ever resumed (the actual `Wake` call at L136 is already guarded by `route.State == Paused`). So this isn't a functional bug, but mixing the wake path and the generic 502 fallback under one broad branch makes the logic hard to follow and leads to scattered/redundant checks:

- `route.State == Paused` is evaluated twice (L126 and L136)
- `EnableWakeOnTraffic` is evaluated twice (L125 and L136)
- `len(parts) == 2` is evaluated twice (L126 and L138)
- The `route.State != Running` re-check at L161 is always true inside this block (state doesn't change), i.e. dead condition

Suggest expressing the wake intent strictly on `Paused` and letting all other non-Running states fall through to 502:

```go
if route.State != agentsv1alpha1.SandboxStateRunning {
    if f.config.EnableWakeOnTraffic && route.State == agentsv1alpha1.SandboxStatePaused {
        // wake path (Paused / EnableWakeOnTraffic already established)
    }
    // all other non-Running states -> 502 (drop the always-true re-check)
    ...
    return api.LocalReply
}
```

AiRanthem (commented):
Review notes on the wake-on-traffic changes (local `go build ./...`, `go test` for `pkg/sandbox-gateway/...` + `pkg/utils/proxyutils/...`, and `-race` on `wake`/`filter` all pass). Four inline notes below: one under-implemented gap, one simplification (drop singleflight), one log nit, and one RBAC least-privilege nit.

docs/proposals/20260627-wake-on-traffic-via-spec-patch.md
```diff
@@ -0,0 +1,101 @@
...
+       |-> Forward request or return 503 on timeout/failure
+  |-> sandbox Paused + WakeOnTraffic disabled: return 502 (current behavior)
+```
+
+## Key Components
+
+### Annotation Lifecycle
+
+- `AnnotationWakeOnTraffic` ("agents.kruise.io/wake-on-traffic"): Set to "true" at
+  creation time when `autoResume=true` is passed in the E2B create request. Updated
```
AiRanthem: The annotation lifecycle described here is not fully implemented yet. `AnnotationWakeOnTraffic` / `AnnotationWakeTimeoutSeconds` are only *read* in production code (`proxyutils.GetRouteFromSandbox`, `wake.wakeInternal`) — nothing *writes* them. The `autoResume` create/connect parameter that this section says sets them was removed in this same branch (11e81c1f), and `basicSandboxCreateModifier` does not set them either. So today the only way to enable wake-on-traffic on a sandbox is an out-of-band `kubectl annotate` (which is exactly what the e2e test does).

Suggest either wiring up a writer on the create/connect path, or updating this proposal to state that enabling is currently manual/out-of-band — otherwise the doc and the code diverge.

pkg/sandbox-gateway/wake/wake.go
```diff
@@ -0,0 +1,178 @@
...
+	ch := w.flight.DoChan(key, func() (interface{}, error) {
```
AiRanthem: I think the `singleflight` here can be dropped — `sandboxcr.Sandbox.Resume` already provides its own concurrency control, so coalescing at the gateway is not needed for correctness:

1. **Spec patch is first-writer-wins.** `retryUpdate` runs under `retry.RetryOnConflict`, and the modifier short-circuits with `if !sbx.Spec.Paused { return false }`, so only one caller actually flips `Spec.Paused=false` / `PauseTime`; the rest re-read and no-op.
2. **The wait is already refcount-shared.** `NewSandboxResumeTask` -> `AcquireEntry` keys the wait hook by `type/namespace/name` and refcounts it, so concurrent Resume callers for the same sandbox share one `WaitEntry` / done-channel / checker (opposite actions get a `WaitTaskConflictError`). The expensive part — waiting for Ready — is already deduplicated.

Net: singleflight only saves a few informer GETs within a single gateway replica, and it does not help cross-replica bursts (those rely on Resume's first-writer-wins anyway).

More importantly, singleflight currently adds a footgun: `Wake` runs the shared `DoChan` fn under the *first* caller's detached `wakeCtx`, and `defer wakeCancel()` cancels it when that caller returns. If that caller's stream is torn down (Envoy `Destroy` -> ctx cancel), the shared Resume's `Wait` is cut short for *all* coalesced callers, so they get `context.Canceled` -> spurious 503 even though the sandbox is resuming. `TestWakeSingleflightCallerCancel` does not catch this because it only asserts `Spec.Paused==false`, which the early patch already set.

Calling `wakeInternal` directly under each caller's own ctx would be simpler, remove the shared-cancellation footgun, and still be correct thanks to Resume's first-writer-wins + refcounted wait. Trade-off: a burst of N concurrent requests to one paused sandbox on one replica would issue ~N extra informer GETs + N-1 conflict retries, which `RetryOnConflict` already handles gracefully.

pkg/sandbox-gateway/filter/filter.go
```diff
@@ -100,15 +114,61 @@ func (f *sandboxFilter) DecodeHeaders(header api.RequestHeaderMap, endStream boo
...
+		// directly, covering the window between kubectl annotate and the
+		// gateway controller reconciling the change into the route registry.
+		waker := wake.GetWaker()
+		parts := strings.SplitN(sandboxID, "--", 2)
+		shouldWake := route.WakeOnTraffic
+		if !shouldWake && f.config.EnableWakeOnTraffic && waker != nil &&
+			len(parts) == 2 && route.State == agentsv1alpha1.SandboxStatePaused {
+			shouldWake = waker.HasWakeAnnotation(context.Background(), parts[0], parts[1])
+		}
+		logger.Info("Wake eligibility check",
```
AiRanthem: This `Info` log fires on every request to a non-Running sandbox (regardless of whether wake is enabled), so a paused sandbox under traffic will flood the logs with this 6-field `eligibility` line — and it is emitted even on the plain 502 path, so it is noisy without adding much signal.

Suggestions: drop it to `Debug` to match the other per-request logs in this filter, and/or only log when a wake is actually attempted (or skipped for a non-obvious reason) rather than unconditionally.

config/sandbox-gateway/rbac.yaml
```diff
@@ -8,13 +8,16 @@ metadata:
...
-    verbs: ["get", "list", "watch"]
+    verbs: ["get", "list", "watch", "update", "patch"]
   - apiGroups: ["agents.kruise.io"]
     resources: ["sandboxes/status"]
     verbs: ["get"]
   - apiGroups: [""]
     resources: ["pods"]
     verbs: ["list"]
+  - apiGroups: ["agents.kruise.io"]
+    resources: ["sandboxsets"]
```
AiRanthem: This `sandboxsets` grant looks unnecessary. The wake path (`sandboxcr.Sandbox.Resume` -> `refreshFromAPIReader` / `retryUpdate` / `InplaceRefresh` / `GetRoute`) only touches the `Sandbox` object; `SandboxSet` is read exclusively by the claim/clone flows (`PickSandboxSet` / `newSandboxFromSandboxSet`), which the gateway never calls. The gateway also builds its cache with `sandboxOnly=true` (skips the SandboxSet informer/index), and its `SandboxReconciler` only reads `Sandbox`. Unless I am missing a lazy client read of SandboxSet, this grant can be dropped to keep the gateway least-privileged. (The added `update`/`patch` on `sandboxes` is correct and needed.)

AiRanthem (dismissed):

pkg/sandbox-gateway/filter/filter.go
```diff
@@ -116,8 +152,7 @@ func (f *sandboxFilter) DecodeHeaders(header api.RequestHeaderMap, endStream boo
 	if f.config.EnableAuth && route.AccessToken != "" {
```
AiRanthem: **Blocking: this access-token check is bypassed on the wake path.**

This validation only runs for requests that were already `Running` at lookup time. The Paused/wakeable branch above (lines 117-148) returns `api.Running` before reaching here, and the async completion calls `Continue(api.Continue)` (see `completeWithContinue`) without re-entering this code. So a paused sandbox is woken and proxied without the token ever being validated.

With `EnableAuth=true` and a token configured, a request with a missing or wrong token can wake a Paused sandbox and get proxied downstream. The route already carries `AccessToken`, so this is an ordering bug, not missing data.

Suggested fix: validate the token right after a successful route lookup, before any state/wake handling. Please verify: Paused+valid token wakes; Paused+missing token -> 401 with `Spec.Paused` untouched; Paused+wrong token -> 401 and no wake started; Running path unchanged.

pkg/sandbox-gateway/filter/filter.go
```diff
@@ -137,6 +172,207 @@ func (f *sandboxFilter) DecodeHeaders(header api.RequestHeaderMap, endStream boo
...
+	f.completing = false
+	if markDestroyed {
+		f.destroyed = true
+	}
+	return !f.destroyed && !f.completed
+}
+
+// Destroy cancels any in-flight async wake context. Called by Envoy when
+// the filter is destroyed (e.g., stream reset).
+func (f *sandboxFilter) Destroy() {
```
AiRanthem: **Blocking: `Destroy()` is never called by Envoy.**

The vendored `api.StreamFilter` lifecycle method is `OnDestroy(DestroyReason)` (`vendor/github.com/envoyproxy/envoy/contrib/golang/common/go/api/filter.go:80,92`), and the embedded `PassThroughStreamFilter` already provides a no-op `OnDestroy` (line 105); Envoy invokes it via `shim.go:371`. This `Destroy()` compiles but Envoy calls the embedded no-op instead, so the custom cancel logic never runs. The test at `filter_test.go:1475` calls `Destroy()` directly, which validates a method Envoy never calls and hides the gap.

Consequently, on client disconnect / reset / stream destroy the in-flight wake context is not canceled and `destroyed` is not set; the goroutine can run to full timeout and then attempt `Continue` / `SendLocalReply` on a dead stream, relying only on panic-string matching for safety.

Suggested fix: implement `func (f *sandboxFilter) OnDestroy(reason api.DestroyReason)` (move the current logic into it), add `var _ api.StreamFilter = (*sandboxFilter)(nil)`, and drive `OnDestroy` through the `api.StreamFilter` interface in tests.

pkg/sandbox-gateway/server/server.go
```diff
@@ -114,6 +125,7 @@ func (s *Server) Start(ctx context.Context) error {
 	labelSelector := os.Getenv(EnvLabelSelector)
 
 	s.peerManager = peers.NewMemberlistPeers(s.client, peers.NodePrefixSandboxGateway+nodeName, namespace, labelSelector)
+	globalPeerManager = s.peerManager
```
AiRanthem: **Blocking: Paused routes get deleted by peer refresh, breaking wake-on-traffic.**

Wiring the wake path to peer sync via `globalPeerManager` surfaces a problem in `handleRefresh` (lines 182-192, not shown in this diff): it Updates only `Running` routes and Deletes every other state, including `Paused`.

The manager pushes Paused routes on pause (`PauseSandbox` -> `syncRoute` -> `SyncRouteWithPeers`, `pkg/sandbox-manager/api.go:326,300-302`), and manager + gateway share one memberlist mesh (same `agents.kruise.io/peer=true` selector + namespace; `GetPeers()` returns all alive members regardless of the `sm-` / `sg-` prefix). So a Paused refresh reaches the gateway and deletes the route. The local controller wrote all states (`gateway_controller.go:63-65`), but nothing re-adds it until the Sandbox changes again. Traffic then hits `sandbox not found` at the initial registry lookup and wake is never attempted.

The delete-non-Running behavior is pre-existing (this PR only adds `globalPeerManager` here), but wake-on-traffic now depends on Paused routes surviving peer sync, so it must be addressed. Suggested fix: in `handleRefresh`, Update (with resourceVersion guard) for Running / Paused / Creating / Available; Delete only for Dead or an explicit deletion marker; decide the empty-state case explicitly instead of a blanket non-Running delete.

pkg/sandbox-gateway/wake/wake.go
```diff
@@ -0,0 +1,152 @@
...
+	// Reuse the existing sandbox-manager connect Resume implementation.
+	// AsSandbox wraps the sandbox with the cache provider + storage registry.
+	// Resume refreshes from API reader, so the sandbox object is re-fetched
+	// with the latest state before patching.
+	sandbox := sandboxcr.AsSandbox(&sbx, w.cache)
+
+	var opts infra.ResumeOptions
+	if wakeTimeout > 0 {
+		opts.Timeout = &timeout.Options{
+			PauseTime: time.Now().Add(wakeTimeout),
```
AiRanthem, line 132: **Non-blocking: `wakeTimeoutSeconds` carries two different meanings.**

The config comment defines it as the max wait for resume (`filter/config.go:55-57`), but here the same value also becomes the next auto-pause `PauseTime`. E2B create only writes the wake-on-traffic annotation on `autoResume.enabled` and does not persist the request `timeout` (`pkg/servers/e2b/create.go:399-401`); the e2e sets `wake-timeout-seconds` manually via `kubectl annotate`.

So a user creating `{timeout:120, autoPause:true, autoResume:{enabled:true}}` may re-pause after the gateway 60s default instead of their 120s lifecycle. Consider splitting `wakeWaitTimeout` vs `resumeAutoPauseTimeout`, or persisting the create timeout as the wake annotation. Whether this is a blocker depends on the intended E2B timeout-compatibility contract.

pkg/sandbox-gateway/server/server.go
```diff
@@ -44,6 +44,17 @@ const (
 	EnvMemberlistBindPort = "MEMBERLIST_BIND_PORT"
 )
 
+// globalPeerManager is set when server.Start() creates the peerManager.
+// It allows other packages (e.g. wake) to access the peer manager for
+// SyncRouteWithPeers without creating a full Server instance.
+var globalPeerManager peers.Peers
```
AiRanthem: **Non-blocking: `globalPeerManager` lifecycle / concurrency.**

Unsynchronized package-level var, written in `Start` and read from the wake path, and not cleared in `Stop`. Prefer explicit dependency injection, or at least atomic/mutex plus clearing on stop. There is a nil guard downstream, so not a blocker on its own.

test/e2b/test_wake_on_traffic.py
```diff
@@ -0,0 +1,215 @@
...
+        headers["X-Access-Token"] = access_token
+
+    print(f"sending wake traffic to {GATEWAY_URL} for {sandbox_id}")
+    resp = _request_gateway_until_forwarded(headers, timeout_sec=120)
+    assert resp is not None, "gateway did not respond within timeout"
+    print(f"wake response: status={resp.status_code} body={resp.text[:200]!r}")
+    print(f"wake response headers: {dict(resp.headers)}")
+
+    # Step 7: Assert wake succeeded (not 502/503)
+    assert resp.status_code != 502, (
```
AiRanthem, line 167: **Non-blocking: success oracle is loose.**

The loop polls until any non-502/503 response, and these final assertions only exclude 502/503, so 200/401/404/500 would pass the HTTP check. The test does additionally assert the sandbox reaches Running and the annotations persist (lines ~200, 209-213), which partially covers correctness, but the HTTP oracle should assert an explicit expected 2xx / known app response, and ideally verify that the first suspended request is actually forwarded rather than relying on later polling requests to mask a failure.

AiRanthem (changes requested):
Requesting changes. Verified against the current head by tracing the concrete code paths (not coverage inference). There are 3 blocking issues plus 3 non-blocking notes, all left as inline comments:

- Blocking 1: access-token auth is bypassed on the paused wake path.
- Blocking 2: `Destroy()` is not Envoy's `OnDestroy`, so the cancel-on-destroy never fires.
- Blocking 3: peer refresh deletes Paused routes, breaking wake-on-traffic in multi-replica topologies.

Suggested review order: `filter.go` (auth order, paused wake branch, destroy lifecycle) -> `server.go` (`handleRefresh`, `globalPeerManager`) -> `api.go` (Pause/Resume/syncRoute) -> `wake.go` (deadline, timeout annotation) -> `create.go` + `test_wake_on_traffic.py`.

PR #590, opened 2026-06-30
fix(sandboxclaim): avoid claim identity collisions

### LINKED ISSUE #311: [feature request] Avoid identity collisions in SandboxClaim
SandboxClaim claimed-sandbox counting should use claim-name label selector plus claim-uid filtering to avoid identity  collisions

> This issue should be fixed after sandbox-manager cache is migrated to controller-runtime architecture

## Background

  The current claimed-sandbox counting logic relies on two identifiers:

  - spec.annotations[AnnotationOwner] as the owner identity
  - LabelSandboxClaimName as the claim-name label

  This creates two correctness risks:

  1. AnnotationOwner is not a strong unique identity, so recreated claims or claims with the same name can collide.
  2. LabelSandboxClaimName is an internal label, but it can be overwritten by user-provided spec.labels, which makes
     already claimed sandboxes invisible during later reconciliations.

  ## Problem

  As a result, SandboxClaim status may become inaccurate:

  - Old sandboxes may be incorrectly counted as belonging to a new claim.
  - Already claimed sandboxes may not be counted anymore after reconciliation.
  - ClaimedReplicas may become too high or too low, which can lead to missing allocations or duplicate allocations.

  ## Expected Behavior

  Claimed sandboxes should be identified in a stable way that cannot be broken by user-controlled labels.

  The counting flow should be:

  1. First filter sandboxes by claim-name using a LabelSelector.
  2. Then filter the remaining results by claim-uid for exact identity matching.

  This keeps the query efficient while ensuring the final result is unique to the current claim.

  ## Proposed Solution

  Update the counting logic to:

  - keep claim-name as the first-level label filter
  - use claim-uid as the final unique identity check
  - prevent internal labels from being overwritten by user-provided spec.labels, or make the counting logic independent
    from that merge behavior

  ## Acceptance Criteria

  - Recreated SandboxClaim objects with the same name do not inherit old sandboxes incorrectly.
  - User-defined labels cannot break claimed-sandbox counting.
  - ClaimedReplicas remains stable across reconciliations.
  - Tests cover:
      - recreated claim with the same name
      - user overriding agents.kruise.io/claim-name in spec.labels
      - combined claim-name selector plus claim-uid filter

zhuangzhewei09:

This PR fixes SandboxClaim claimed-sandbox counting so a claim is identified by both internal claim-name bookkeeping and the claim UID.

The previous flow could become unstable because `agents.kruise.io/claim-name` is an internal label but user-provided `spec.labels` could override it, while claimed-sandbox recovery did not narrow the cache query by claim name. This could make recreated same-name claims or later reconciliations report incorrect `ClaimedReplicas`.

Key changes:

1. Add a claim-name sandbox cache index and extend `ListSandboxesOptions` with `ClaimName`.
2. Apply claim-name filtering together with the existing claim UID owner filter in `ListSandboxes` and `CountActiveSandboxes`.
3. Pass `ClaimName: claim.Name` from SandboxClaim claimed-sandbox recovery.
4. Preserve the internal `agents.kruise.io/claim-name` label after applying user labels so `spec.labels` cannot override it.
5. Add regression tests for recreated same-name claims, combined claim-name/UID filtering, dead sandbox exclusion, and user-provided claim-name label overrides.

Fixes #311

```bash
go test -count=1 ./pkg/cache/... ./pkg/controller/sandboxclaim/core/...
git diff --check
```

- No CRD/API changes are introduced, so `make generate` and `make manifests` are not required.
- The new claim-name index is registered through `GetIndexFuncs`, keeping production cache and test cache index definitions consistent.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -367,6 +367,23 @@ func listObjectWithUserAndNamespace[T ctrlclient.ObjectList](ctx context.Context
 	return client.List(ctx, list, listOpts...)
 }
 
+func listSandboxesWithOptions(ctx context.Context, client ctrlclient.Client, list *agentsv1alpha1.SandboxList, opts ListSandboxesOptions, listOpts ...ctrlclient.ListOption) error {
+	if opts.Namespace != "" {
+		listOpts = append(listOpts, ctrlclient.InNamespace(opts.Namespace))
+	}
+	matchingFields := ctrlclient.MatchingFields{}
+	if opts.User != "" {
+		matchingFields[IndexUser] = opts.User
+	}
+	if opts.ClaimName != "" {
+		matchingFields[IndexSandboxClaimName] = opts.ClaimName
+	}
+	if len(matchingFields) > 0 {
+		listOpts = append(listOpts, matchingFields)
+	}
+	return client.List(ctx, list, listOpts...)
+}
+
 func (c *Cache) ListSandboxSets(ctx context.Context, opts ListSandboxSetsOptions) ([]*agentsv1alpha1.SandboxSet, error) {
 	resultVal, err, _ := c.indexGetGroup.Do("sandboxsets:"+opts.Namespace, func() (any, error) {
 		list := &agentsv1alpha1.SandboxSetList{}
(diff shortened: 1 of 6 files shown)


zhuangzhewei09:
@furykerry @zmberg @PersistentJZH PTAL~

PR #591, opened 2026-06-30
fix(sandbox-metrics): fix stale condition metrics and add _time metrics for abnormal states

liangxiaoping:

This PR fixes a bug where stale condition metrics caused false alerts, and adds two new `_time` metrics for abnormal state tracking.

**Bug fix: stale condition metrics**

When a condition (e.g. `SandboxConditionResumed`) is removed from the sandbox's conditions slice during lifecycle transitions (e.g. entering Paused phase removes the Resumed condition), the corresponding "un" metric (`sandbox_status_unresumed`, `sandbox_status_unpaused`, `sandbox_status_inplace_updating`) persisted with its previous value of `1` from an earlier reconcile. This caused false L1 alerts because Prometheus continued to report `sandbox_status_unresumed == 1` even though the sandbox was actually in a healthy Running state with no `SandboxResumed` condition at all.

**Root cause:** The metrics were only set inside the `for _, condition := range sandbox.Status.Conditions` loop. When a condition was absent, the loop never executed the `case` branch for it, so the old gauge value was never cleared.

**Fix:** Reset all three condition-based "un" metrics to `0` at the start of `recordSandboxConditionMetrics`, before iterating the conditions slice. If the condition is present and False, it will be set back to 1 during the loop; if absent, it correctly stays at 0.

**New metrics: `sandbox_status_abnormal_time` and `sandbox_runtime_container_abnormal_time`**

Following the same pattern as `sandbox_status_unresumed_time`, these new gauge metrics record the Unix timestamp when a sandbox (or runtime container) entered an abnormal state. This enables duration-based alerting without relying on the "un" condition metrics:

```yaml
- alert: SandboxAbnormalCountL1
  expr: |
    sum by(namespace) (
      (sandbox_status_abnormal{} == 1)
      and
      (time() - sandbox_status_abnormal_time{} > 600)
    ) > 10
```

NONE

1. Run unit tests:
   ```bash
   go test ./pkg/controller/sandbox/ -v -run "TestStale|TestSandboxStatusAbnormalTime|TestSandboxRuntimeContainerAbnormalTime|TestDeleteSandboxMetrics_ClearsAbnormalTime" -count=1
   ```
2. Run full test suite:
   ```bash
   go test ./pkg/controller/sandbox/ -v -count=1
   ```
3. Verify the bug fix: Create a sandbox that transitions from Resuming (SandboxResumed=False) to Running (SandboxResumed condition removed). The `sandbox_status_unresumed` metric should be `0`, not `1`.

- The bug fix is in `recordSandboxConditionMetrics` — 3 lines that reset the "un" gauges to 0 before the conditions loop
- The `_time` metrics follow the exact same pattern as the existing `sandbox_status_unresumed_time` metric
- `DeleteSandboxMetrics` uses `DeletePartialMatch` for the new time metrics (same as the existing abnormal metrics), since the cleanup function only receives namespace+name
- `recordRuntimeContainerAbnormal` deletes the time metric (via `DeleteLabelValues`) when the container is healthy, rather than setting it to 0 — this matches the behavior of `sandbox_status_abnormal_time` which is also deleted (not zeroed) when the abnormal state clears

### DIFF
--- a/pkg/controller/sandbox/metrics_test.go
+++ b/pkg/controller/sandbox/metrics_test.go
@@ -2525,12 +2525,45 @@ func TestSandboxRuntimeContainerAbnormalTime(t *testing.T) {
 // cleans up sandbox_status_abnormal_time and sandbox_runtime_container_abnormal_time.
 func TestDeleteSandboxMetrics_ClearsAbnormalTimeMetrics(t *testing.T) {
 	ns, name := "default", "cleanup-abnormal-time-sandbox"
+	now := metav1.NewTime(time.Now())
 
-	// Set up abnormal metrics
-	sandboxStatusAbnormal.WithLabelValues(ns, name, "pause_incomplete").Set(1)
-	sandboxStatusAbnormalTime.WithLabelValues(ns, name, "pause_incomplete").Set(float64(time.Now().Unix()))
-	sandboxRuntimeContainerAbnormal.WithLabelValues(ns, name, "agent-runtime").Set(1)
-	sandboxRuntimeContainerAbnormalTime.WithLabelValues(ns, name, "agent-runtime").Set(float64(time.Now().Unix()))
+	// Populate abnormal state through the real recording functions so that both
+	// the Prometheus gauges AND the sync.Map start-time trackers are populated.
+	// Directly setting gauges would leave the sync.Maps empty, causing the Range
+	// cleanup in DeleteSandboxMetrics to be uncovered.
+	sandbox := &agentsv1alpha1.Sandbox{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name, Namespace: ns,
+			CreationTimestamp: metav1.NewTime(time.Now()),
+		},
+		Status: agentsv1alpha1.SandboxStatus{
+			Phase: agentsv1alpha1.SandboxPaused,
+			Conditions: []metav1.Condition{{
+				Type:               string(agentsv1alpha1.SandboxConditionPaused),
+				Status:             metav1.ConditionFalse,
+				LastTransitionTime: now,
+			}},
+		},
+	}
+	pod := &corev1.Pod{
+		Status: corev1.PodStatus{
+			InitContainerStatuses: []corev1.ContainerStatus{
+				{Name: "agent-runtime", Ready: false, RestartCount: 2},
+			},
+		},
+	}
+
+	recordSandboxMetrics(sandbox, pod)
+
+	// Verify sync.Map entries were populated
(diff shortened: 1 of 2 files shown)


pkg/controller/sandbox/metrics_test.go
```diff
@@ -2219,3 +2219,564 @@ func TestDeleteSandboxMetrics_ClearsRuntimeContainerAbnormal(t *testing.T) {
...
+
+	timeVal = testutil.ToFloat64(sandboxStatusAbnormalTime.WithLabelValues(ns, name, "pause_incomplete"))
+	if timeVal != 0 {
+		t.Errorf("sandbox_status_abnormal_time{type=pause_incomplete} = %v, want 0 after recovery", timeVal)
+	}
+}
+
+// TestSandboxRuntimeContainerAbnormalTime verifies that sandbox_runtime_container_abnormal_time
+// is set when a runtime container is abnormal and deleted when it becomes healthy.
+func TestSandboxRuntimeContainerAbnormalTime(t *testing.T) {
```
furykerry, line 2454: Location: pkg/controller/sandbox/metrics_test.go — TestSandboxRuntimeContainerAbnormalTime and TestRuntimeContainerAbnormalTimeStableAcrossReconciles

Issue: These two test functions do not call defer DeleteSandboxMetrics(ns, name), unlike every other test in the file. After the tests complete, the sandboxRuntimeContainerAbnormal gauge series (value 0) and runtimeContainerAbnormalStartTimes sync.Map entries persist, potentially affecting subsequent tests that share the Prometheus registry.

pkg/controller/sandbox/metrics_test.go
```diff
@@ -2219,3 +2219,564 @@ func TestDeleteSandboxMetrics_ClearsRuntimeContainerAbnormal(t *testing.T) {
...
+
+	unpausedVal = testutil.ToFloat64(sandboxStatusUnpaused.WithLabelValues(ns, name))
+	if unpausedVal != 0 {
+		t.Errorf("sandbox_status_unpaused = %v, want 0 (condition absent, should be reset)", unpausedVal)
+	}
+}
+
+// TestSandboxStatusAbnormalTime verifies that sandbox_status_abnormal_time is set when
+// the sandbox enters an abnormal state and deleted when it returns to normal.
+func TestSandboxStatusAbnormalTime(t *testing.T) {
```
furykerry, line 2358: TestSandboxStatusAbnormalTime and TestSandboxStatusAbnormalTime_PauseIncomplete test the same logic with different abnormal types (resume_incomplete vs pause_incomplete) and could be merged into a single table-driven test parameterized by phase/condition/type.

PR #597, opened 2026-07-01
docs(proposal): add dynamic environment variable injection design proposal

furykerry:

Add a design proposal for per-claim environment variable injection into pre-warmed sandbox pods.

The proposal introduces `run_with_envs.sh` — a deferred execution wrapper that blocks the container's main command until the sandbox is claimed and env vars are written. Key design points:

- **`EnvVarsInjectionPolicy` field** on `SandboxClaim` and `SandboxSpec`: `None` (default) or `Auto` (opt-in). No annotation-based toggle.
- **Command + exec-probe wrapping**: Under `Auto`, the controller wraps the container command and exec probes (`startupProbe`, `livenessProbe`, `readinessProbe`) with `run_with_envs.sh`. Probe mode (`--probe`) returns success immediately before env initialization, keeping pre-warmed pods healthy.
- **Annotation-only env persistence**: Env vars are written to `AnnotationInitRuntimeRequest` only — not to the container spec — to avoid triggering the in-place upgrade path which does not support env var changes.
- **Resume/upgrade restoration**: The `Initialize()` hook re-writes env vars to the filesystem after pod recreation.
- **Limitations**: `postStartHook` and non-exec probes (`httpGet`, `tcpSocket`) are not wrapped. Secrets should use `traffic-extension` SecurityProfile instead of `envVars`.

NONE

1. Review the proposal document at `docs/proposals/20260701-dynamic-env-injection.md`
2. Confirm the design aligns with existing sandbox lifecycle flows (claim, pause/resume, upgrade)

- The `EnvVarsInjectionPolicy` default is `None` to preserve backward compatibility — `Auto` is opt-in.
- The `postStartHook` and non-exec probe limitations are explicit user-facing constraints.
- In-place upgrade support for env vars is tracked as future work.

### DIFF
--- a/docs/proposals/20260701-dynamic-env-injection.md
+++ b/docs/proposals/20260701-dynamic-env-injection.md
@@ -1,0 +1,692 @@
+---
+title: Dynamic Environment Variable Injection for Pre-warmed Sandboxes
+authors:
+  - "@furykerry"
+reviewers:
+  - "@BH4AWS"
+  - "@zmberg"
+creation-date: 2026-07-01
+status: provisional
+---
+
+# Dynamic Environment Variable Injection for Pre-warmed Sandboxes
+
+This proposal introduces a mechanism to dynamically inject environment variables into pre-warmed sandbox pods
+**after** allocation, enabling users to specify per-claim environment variables and custom commands that only
+execute once the sandbox is assigned and initialized.
+
+## Motivation
+
+In the current architecture, a SandboxSet pre-creates a pool of warm sandbox pods. The pod's container command
+is defined in the SandboxTemplate and starts executing **immediately** when the pod is created — long before any
+user claims the sandbox. This means:
+
+1. **No per-claim environment customization**: Environment variables injected via the envd `/init` endpoint
+   after claim are not available to the container's main process, which has already started.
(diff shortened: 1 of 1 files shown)


zmberg (commented):
Superseded — see the inline comments.

docs/proposals/20260701-dynamic-env-injection.md
```diff
@@ -0,0 +1,692 @@
...
+- Maintain backward compatibility with existing SandboxSet/SandboxClaim/E2B API flows.
+- Support both SandboxClaim CRD and E2B sandbox create API.
+
+### Non-Goals/Future Work
+
+- Dynamic env var injection for already-running (post-claim) sandboxes.
+- Secret/ConfigMap-based env var sources (future enhancement).
+- Env var injection without agent-runtime (envd-less path).
+- In-place upgrade support for env var changes on the Sandbox CR. Currently, modifying the
+  container spec triggers in-place upgrade which does not support env var updates.
```
zmberg, line 48: Add VM-based sandboxes (e.g. OSWorld) as a non-goal too. They have no container concept, so the `run_with_envs.sh` command-wrapping approach fundamentally does not apply.

Issue #593, opened 2026-07-01
feat: Add K8s events for key lifecycle transitions across core CRD controllers


furykerry:
### Background
The core CRD controllers (Sandbox, SandboxSet, SandboxClaim, SandboxUpdateOps, Checkpoint)
currently emit K8s events only for a subset of lifecycle transitions — primarily error and
warning cases. Many key phase transitions (e.g., Sandbox Pending→Running, Paused, Resuming;
SandboxUpdateOps Pending→Updating→Completed/Failed; Checkpoint started→succeeded) lack
corresponding K8s events. This makes it difficult for operators and external tooling to
observe sandbox lifecycle via standard K8s event streams (e.g., `kubectl describe`,
event exporters, monitoring integrations).

### Issue Type
feature

### Relevant Code

Current event coverage by controller:

- **Sandbox** (`pkg/controller/sandbox/`): Events exist for reuse operations
  (Started/Succeeded/Rejected), checkpoint failures, and runtime initialization.
  No events for main phase transitions: Pending, Running, Paused, Resuming,
  Upgrading, Failed, Succeeded, Terminating.

- **SandboxSet** (`pkg/controller/sandboxset/`): Events exist for sandbox creation,
  scale-down deletion, and failed sandbox cleanup. No events for rolling update
  completion or scale operation summaries.

- **SandboxClaim** (`pkg/controller/sandboxclaim/`): Reasonable coverage for
  claiming, completion, and TTL deletion. May benefit from events for claim
  failure scenarios.

- **SandboxUpdateOps** (`pkg/controller/sandboxupdateops/`): Only `PatchFailed`
  and `SandboxUpgrading` events. No events for overall phase transitions
  (Pending → Updating → Completed/Failed) or batch operation summaries.

- **Checkpoint** (`pkg/controller/sandbox/core/checkpoint.go`): Only warning events
  for checkpoint failures. No normal events for checkpoint started or succeeded.

### Notes

- Events should be added for **key** lifecycle transitions only — not for every
  reconcile iteration or intermediate status change — to avoid excessive load on
  the K8s API server and etcd.
- Consider using event deduplication and aggregation where appropriate (e.g.,
  the built-in `EventRecorder` deduplication mechanism) to prevent event flooding
  when many sandboxes transition simultaneously (e.g., batch scale operations in
  SandboxSet or bulk updates in SandboxUpdateOps).
- For high-cardinality operations like SandboxSet scaling, consider emitting a
[...]

PR #603, opened 2026-07-02
feat(controller): add k8s lifecycle events

zhuangzhewei09:

This PR adds Kubernetes events for key lifecycle transitions in core controllers.

* Emits Sandbox phase events after successful status updates, so events are recorded only when `status.phase` actually changes.
* Records Sandbox deletion progress with a `SandboxTerminating` event, without changing the existing `status.phase` semantics.
* Adds normal events for Checkpoint start and success, while keeping existing warning events for failure paths.
* Adds SandboxUpdateOps phase summary events with total / updated / updating / failed replica counts.
* Adds focused unit tests for the new event recording behavior.

Refs #593

This PR covers the main lifecycle event gaps for Sandbox, Checkpoint, and SandboxUpdateOps. It intentionally does not change Sandbox `Terminating` status semantics, and it does not expand SandboxSet per-sandbox event volume in this first pass.

```bash
go test ./pkg/controller/...
```

The event emission is placed after successful status updates where possible, to avoid recording lifecycle events for failed status patches or repeated reconcile attempts.
For Sandbox deletion, this PR only emits a SandboxTerminating event when deletion is in progress. It does not set status.phase=Terminating, because the existing deletion path performs finalizer cleanup directly and does not currently persist that phase.

### DIFF
--- a/pkg/controller/sandbox/sandbox_controller.go
+++ b/pkg/controller/sandbox/sandbox_controller.go
@@ -72,6 +72,33 @@ var (
 	csiResetSignalFileName      = "reset"
 )
 
+const (
+	eventReasonSandboxTerminating = "SandboxTerminating"
+)
+
+type sandboxPhaseTransition struct {
+	oldPhase agentsv1alpha1.SandboxPhase
+	newPhase agentsv1alpha1.SandboxPhase
+}
+
+var sandboxPhaseTransitionMessages = map[sandboxPhaseTransition]string{
+	{oldPhase: "", newPhase: agentsv1alpha1.SandboxPending}:                              "Start sandbox",
+	{oldPhase: agentsv1alpha1.SandboxPending, newPhase: agentsv1alpha1.SandboxRunning}:   "Sandbox started",
+	{oldPhase: agentsv1alpha1.SandboxRunning, newPhase: agentsv1alpha1.SandboxPaused}:    "Pause sandbox",
+	{oldPhase: agentsv1alpha1.SandboxPaused, newPhase: agentsv1alpha1.SandboxResuming}:   "Resume sandbox",
+	{oldPhase: agentsv1alpha1.SandboxPaused, newPhase: agentsv1alpha1.SandboxRunning}:    "Resume sandbox",
+	{oldPhase: agentsv1alpha1.SandboxResuming, newPhase: agentsv1alpha1.SandboxRunning}:  "Sandbox resumed",
+	{oldPhase: agentsv1alpha1.SandboxRunning, newPhase: agentsv1alpha1.SandboxUpgrading}: "Upgrade sandbox",
+	{oldPhase: agentsv1alpha1.SandboxUpgrading, newPhase: agentsv1alpha1.SandboxRunning}: "Sandbox upgraded",
+	{oldPhase: agentsv1alpha1.SandboxRunning, newPhase: agentsv1alpha1.SandboxRecycling}: "Recycle sandbox",
+	{oldPhase: agentsv1alpha1.SandboxRecycling, newPhase: agentsv1alpha1.SandboxRunning}: "Sandbox recycled",
+}
+
+var sandboxPhaseTargetMessages = map[agentsv1alpha1.SandboxPhase]string{
+	agentsv1alpha1.SandboxSucceeded: "Sandbox succeeded",
+	agentsv1alpha1.SandboxFailed:    "Sandbox failed",
+}
+
 // Enqueuer is the contract the Sandbox controller depends on for async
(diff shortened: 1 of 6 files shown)


pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -341,11 +347,50 @@ func (r *SandboxReconciler) updateSandboxStatus(ctx context.Context, newStatus a
...
+func (r *SandboxReconciler) recordSandboxPhaseEvent(box *agentsv1alpha1.Sandbox, oldPhase agentsv1alpha1.SandboxPhase, newStatus agentsv1alpha1.SandboxStatus) {
+	if r.recorder == nil || oldPhase == newStatus.Phase || newStatus.Phase == "" {
+		return
+	}
+
+	eventType := corev1.EventTypeNormal
+	if newStatus.Phase == agentsv1alpha1.SandboxFailed {
+		eventType = corev1.EventTypeWarning
+	}
+	message := fmt.Sprintf("Sandbox phase changed from %s to %s", phaseForEvent(oldPhase), newStatus.Phase)
```
furykerry: these event is generic, but not informative enough, I think "resume sandbox" is better than "sandbox phase changed from paused to running"

zhuangzhewei09:
@furykerry hi, Thanks for the review. I updated the Sandbox phase event messages to be more informative.

The event `Reason` is still based on the target phase, e.g. `SandboxRunning`, `SandboxPaused`, `SandboxFailed`, so the existing phase-based reason semantics remain stable. The change only makes the event `Message` describe the key lifecycle operation more clearly.

For example:

* `Pending -> Running`: `Sandbox started`
* `Running -> Paused`: `Pause sandbox`
* `Paused -> Resuming` / `Paused -> Running`: `Resume sandbox`
* `Resuming -> Running`: `Sandbox resumed`
* `Running -> Upgrading`: `Upgrade sandbox`
* `Upgrading -> Running`: `Sandbox upgraded`
* `Running -> Reusing`: `Reuse sandbox`
* `Reusing -> Running`: `Sandbox reused`
* `Failed` / `Succeeded`: `Sandbox failed` / `Sandbox succeeded`

Unknown transitions still fall back to the generic message: `Sandbox phase changed from ... to ...`.

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -350,11 +370,60 @@ func (r *SandboxReconciler) updateSandboxStatus(ctx context.Context, newStatus a
...
+}
+
+func sandboxPhaseEventMessage(oldPhase, newPhase agentsv1alpha1.SandboxPhase) string {
+	if message, ok := sandboxPhaseTransitionMessages[sandboxPhaseTransition{oldPhase: oldPhase, newPhase: newPhase}]; ok {
+		return message
+	}
+	if message, ok := sandboxPhaseTargetMessages[newPhase]; ok {
+		return message
+	}
+	return fmt.Sprintf("Sandbox phase changed from %s to %s", phaseForEvent(oldPhase), newPhase)
```
furykerry, line 427: if old and new phase is the same, no need to emit event at all

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -301,7 +319,8 @@ func (r *SandboxReconciler) EnsureSandboxPaused(ctx context.Context, args core.E
 }
 
 func (r *SandboxReconciler) handleTerminating(ctx context.Context, args core.EnsureFuncArgs) (ctrl.Result, error) {
-	pod, _, _ := args.Pod, args.Box, args.NewStatus
+	pod, box, _ := args.Pod, args.Box, args.NewStatus
+	r.recordSandboxTerminatingEvent(box)
```
furykerry: handleTerminating can be called at least twice, one for deleting pod, one for finalizer removal, so two event is emitted, consider only emit event when deleting the pod

pkg/controller/sandbox/sandbox_controller.go
chacha923: hello, i think `recordSandboxPhaseEvent` already early-returns when `oldPhase == newStatus.Phase`, and `sandboxPhaseEventMessage` is only reachable from there. Do you mean judge if old and new phase is the same should move into `sandboxPhaseEventMessage`?

pkg/controller/sandbox/sandbox_controller.go
chacha923: fixed

pkg/controller/sandbox/sandbox_controller.go
chacha923: fixed

PR #605, opened 2026-07-02
Signal csi-sidecar to unmount stale CSI volumes before sandbox reuse

zmberg:
## Summary

Before a sandbox is returned to the pool for reuse, if it carries CSI mounts, the controller now writes an empty reset-signal file into a configurable directory through the agent-runtime files API. A stopping csi-sidecar can detect this file during prestop/SIGTERM and unmount stale CSI volumes before the sandbox is handed back for reuse.

- Add `ensureCSIResetSignal` in the reuse state machine: no-op when the sandbox carries no CSI annotation; hard-fail the reuse when CSI mounts are present but `--csi-reset-signal-dir` is unconfigured; otherwise write the signal file (with a few inline retries) while the runtime is still reachable, before the reuser resets the sandbox.
- Add `--csi-reset-signal-dir` / `--csi-reset-signal-file` controller flags wired into `SandboxReuseConfig`.

## Tests

- Unit tests: cover `ensureCSIResetSignal` (no dir, no CSI annotation, success, custom file name, retry, exhausted retries, canceled context) plus two `doReuse` integration cases exercising the CSI wiring and error propagation through the full state machine.
- E2E: new `Sandbox Reuse CSI Reset Signal` case runs a pool backed by the real agent-runtime (envd) sidecar and verifies a CSI-annotated sandbox reuses successfully — proving the reset-signal write link works end to end. The `E2E-1.32` workflow now builds/loads `agent-runtime` and starts the controller with `--csi-reset-signal-dir`.

## Test plan

- [ ] `go test ./pkg/controller/sandbox/core/...`
- [ ] E2E-1.32 workflow green (new reuse CSI case)

### DIFF
--- a/pkg/controller/sandbox/core/interface.go
+++ b/pkg/controller/sandbox/core/interface.go
@@ -101,8 +101,8 @@ type SandboxReuseConfig struct {
 	FailureShutdownGrace time.Duration
 	// CSIResetSignalDir is the directory inside the sandbox where a reset signal
 	// file is written before reuse when the sandbox carries CSI mounts, so that a
-	// restarting csi-sidecar can detect it and unmount stale volumes. Empty
-	// disables the behavior.
+	// stopping csi-sidecar can detect it during prestop/SIGTERM and unmount stale
+	// volumes. Empty disables the behavior.
 	CSIResetSignalDir string
 	// CSIResetSignalFileName is the name of the reset signal file written into
 	// CSIResetSignalDir.
--- a/pkg/controller/sandbox/core/interface.go
+++ b/pkg/controller/sandbox/core/interface.go
@@ -99,6 +99,14 @@ type SandboxReuseConfig struct {
 	Timeout              time.Duration
 	GracePeriod          time.Duration
 	FailureShutdownGrace time.Duration
+	// CSIResetSignalDir is the directory inside the sandbox where a reset signal
+	// file is written before reuse when the sandbox carries CSI mounts, so that a
+	// restarting csi-sidecar can detect it and unmount stale volumes. Empty
+	// disables the behavior.
+	CSIResetSignalDir string
+	// CSIResetSignalFileName is the name of the reset signal file written into
+	// CSIResetSignalDir.
+	CSIResetSignalFileName string
 }
 
 type SandboxControlArgs struct {
--- a/pkg/controller/sandbox/core/reuse.go
+++ b/pkg/controller/sandbox/core/reuse.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"path"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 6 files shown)


pkg/controller/sandbox/core/reuse.go
```diff
@@ -137,6 +163,54 @@ func (r *SandboxReuseControl) doReuse(ctx context.Context, args EnsureFuncArgs)
...
+			FilePath:    resetFile,
+			Content:     []byte{},
+			Permissions: 0644,
+			Timeout:     csiResetSignalWriteTimeout,
+		})
+		if lastErr == nil {
+			klog.InfoS("Wrote CSI reset signal for reuse", "sandbox", klog.KObj(box), "file", resetFile, "attempt", attempt)
+			return nil
+		}
+		klog.InfoS("Failed to write CSI reset signal, will retry", "sandbox", klog.KObj(box),
```
furykerry: consider remove "will try" text, last attempt will not retry.

test/e2e/sandbox_reuse_test.go
```diff
@@ -379,3 +380,125 @@ var _ = Describe("Sandbox Reuse", func() {
...
+									Command:         []string{"sh", "/workspace/entrypoint.sh"},
+									VolumeMounts:    []corev1.VolumeMount{{Name: "envd-volume", MountPath: "/mnt/envd"}},
+									Env:             []corev1.EnvVar{{Name: "ENVD_DIR", Value: "/mnt/envd"}},
+									RestartPolicy:   &alwaysRestart,
+								},
+							},
+							Containers: []corev1.Container{
+								{
+									Name:         "sandbox",
+									Image:        "centos:7",
```
furykerry: entOS 7 reached EOL in June 2024. Consider switching to a maintained base image

pkg/controller/sandbox/core/interface.go
```diff
@@ -99,6 +99,14 @@ type SandboxReuseConfig struct {
 	Timeout              time.Duration
 	GracePeriod          time.Duration
 	FailureShutdownGrace time.Duration
+	// CSIResetSignalDir is the directory inside the sandbox where a reset signal
+	// file is written before reuse when the sandbox carries CSI mounts, so that a
+	// restarting csi-sidecar can detect it and unmount stale volumes. Empty
```
furykerry: "restarting csi-sidecar" is not accurate, csi-sidecar should unmount the volumes in prestop or SIGTERM handling, so it should be ""stopping csi-sidecar"

zmberg:
Thanks for the review @furykerry, all three addressed in 634163a9:

1. **retry log** — moved the "will retry" log inside the `attempt < maxRetries` branch so the final attempt no longer claims it will retry.
2. **csi-sidecar wording** — replaced "restarting csi-sidecar" with "stopping csi-sidecar (prestop/SIGTERM)" across the flag help, config doc, and reuse comments.
3. **E2E base image** — switched the sandbox container from the EOL `centos:7` to `ubuntu:24.04` (and the postStart from `sh` to `bash`, since `envd-run.sh` uses bash arrays and Ubuntu's `/bin/sh` is dash).

PR #607, opened 2026-07-05
fix(security): mask AccessToken in Route log output and debug endpoint

chengzhycn:

Prevents `Route.AccessToken` from being leaked through logs and the `/debug` HTTP endpoint.

**Changes:**

1. **Route.String() masking** (`pkg/utils/proxyutils/route.go`): Implements `fmt.Stringer` on `Route` struct, always printing `AccessToken:***` regardless of whether the field is empty — preventing attackers from inferring token configuration status via log side-channel.

2. **Debug endpoint masking** (`pkg/sandbox-manager/debug.go`): `GetDebugInfo()` now unconditionally masks `AccessToken` to `"***"` for all routes before returning the JSON response, closing a privilege escalation vector where any valid API Key holder could extract all sandbox tokens.

3. **Unit tests**: Added `route_test.go` and `debug_test.go` covering both token-present and token-absent scenarios.

NONE

1. Run unit tests:
   ```bash
   go test ./pkg/utils/proxyutils/ -run TestRouteString -v
   go test ./pkg/sandbox-manager/ -run TestSandboxManager_DebugMaskAccessToken -v
   ```
2. Verify log output: `SetRoute` logs should display `AccessToken:***` (visible in test output).
3. Verify debug endpoint: `GET /debug` response should show `"accessToken":"***"` for all routes.

- `Route` is a value type stored in `sync.Map`; `ListRoutes()` returns copies, so masking in `GetDebugInfo()` does not affect the actual routing data or peer synchronization.
- The `json:"accessToken,omitempty"` tag remains unchanged — peer sync (`SyncRouteWithPeers`) intentionally transmits the real token for gateway authentication. This is internal cluster communication and is by design.
- The `String()` method uses a constant mask (no conditional on empty) to prevent timing/side-channel inference.

---

- Fix namespace parsing in E2B gateway test helper (`test/e2b/test_gateway.py`): correctly extract namespace from `namespace--name` format sandbox IDs and pass `-n` flag to `kubectl get sandbox`.

### DIFF
--- a/pkg/sandbox-manager/debug.go
+++ b/pkg/sandbox-manager/debug.go
@@ -27,8 +27,14 @@ type DebugInfo struct {
 }
 
 func (m *SandboxManager) GetDebugInfo() DebugInfo {
+	routes := m.proxy.ListRoutes()
+	// Mask AccessToken to prevent credential leakage via the debug endpoint.
+	// Always set to "***" to avoid revealing which sandboxes have tokens configured.
+	for i := range routes {
+		routes[i].AccessToken = "***"
+	}
 	info := DebugInfo{
-		Routes: m.proxy.ListRoutes(),
+		Routes: routes,
 		Peers:  m.proxy.ListPeers(),
 		Pools:  m.infra.LoadDebugInfo(),
 	}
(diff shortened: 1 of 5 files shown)


chengzhycn:
cc @furykerry @AiRanthem for review

PR #612, opened 2026-07-07
Sandbox Auto-Pause and Resume

zmberg:

Adds the design proposal `docs/proposals/20260626-sandbox-auto-pause.md` for Sandbox Auto-Pause and Resume.

The proposal introduces:
- `AutoPausePolicy` embedded in the `Sandbox` CRD for cron-based scheduling and decision rules.
- `Lifecycle.Probes` as generic named probes that write results to standard `SandboxStatus.Conditions`.
- Three composable modes: schedule-driven, probe-driven decisions, and probe-only reporting.
- Reconcile logic, probe contract, state machine, risks, upgrade strategy, and test plan.

NONE

- Review the proposal document for technical accuracy and consistency with existing CRD patterns.
- This is a documentation-only PR; no code changes or unit tests are required.

- The proposal is written in English as requested.
- The document intentionally covers design rationale only; API type implementation and controller logic will follow in subsequent PRs.
- Key design decisions reflected: generic named probes, decoupling of probing and decision-making, `idleCount` for Active probe smoothing, and `messageUnix` for Cron probe timestamp parsing.


docs/proposals/20260626-sandbox-auto-pause.md
```diff
@@ -0,0 +1,1025 @@
...
+                fi
+          periodSeconds: 60
+          timeoutSeconds: 10
+  autoPausePolicy:
+    # schedules not configured — fully probe-driven
+    decisions:
+      - type: pause
+        probe: Active
+        messageRegex: "^inactive$"
+        idleCount: 30              # Pause after 30 consecutive inactive probe reports (~15 minutes)
```
furykerry: if decisions is probe irelevant, consider rename idleCount to successThreshold

docs/proposals/20260626-sandbox-auto-pause.md
```diff
@@ -0,0 +1,1025 @@
...
+      - type: resume
+        probe: Cron
+        messageUnix: true
+        leadTime: 5m
```
furykerry: leadTime seems not necessary, when writing cron expression, one can consider the lead time beforehand

docs/proposals/20260626-sandbox-auto-pause.md
```diff
@@ -0,0 +1,1025 @@
...
+    IdleCount int32 `json:"idleCount,omitempty"`
+
+    // IdleDuration is valid only for Type=pause.
+    // Defines how long the probe must continuously report idle
+    // (i.e., message keeps matching messageRegex) before the controller
+    // pauses the sandbox. If both IdleCount and IdleDuration are set,
+    // IdleDuration takes precedence.
+    // Example: "15m" means the sandbox pauses after being idle for 15 minutes.
+    // +optional
+    IdleDuration *metav1.Duration `json:"idleDuration,omitempty"`
```
furykerry: consider change it to PeriodSeconds, combining PeriodSeconds and SuccessThreshold, one can calculate IdleDuration

docs/proposals/20260626-sandbox-auto-pause.md
```diff
@@ -0,0 +1,869 @@
...
+    // Probes defines a list of named probes that run periodically while the sandbox
+    // is Running. Each probe writes its result to a SandboxStatus.Condition with
+    // type "agents.kruise.io/<name>". Probes are generic — their semantics (e.g.,
+    // "activity detection" vs "cron task detection") are defined by
+    // AutoPausePolicy.PauseStrategy/ResumeStrategy, not by the probe itself.
+    //
+    // Each Probe embeds corev1.Probe inline. Currently only exec, periodSeconds,
+    // and timeoutSeconds are actively used; other fields may be supported later.
+    // +optional
+    Probes []Probe `json:"probes,omitempty"`
```
furykerry: consider move probes out of lifecycle, lifecycle should only contain hook that is pre or post some lifecycle phases

docs/proposals/20260626-sandbox-auto-pause.md
```diff
@@ -0,0 +1,869 @@
...
+              if [ -n "$NEXT_MS" ] && [ "$NEXT_MS" != "null" ]; then
+                echo $((NEXT_MS / 1000))
+              else
+                echo "none"
+              fi
+        periodSeconds: 60
+        timeoutSeconds: 10
+  autoPausePolicy:
+    pauseStrategy:
+      probe: Active
```
furykerry: probe, messageRegex and thresholdDuration should be part of sub-struct , e.g. 
```
  autoPausePolicy:
    pause:
         whenIdleProbeFire:
               probe: Active
               messageRegex: "^inactive$"
               thresholdDuration: 15m   # Pause after condition matches for 15 minutes
        whenRetentionPeriodExpire:
               duration: 600s
   resume: 
         whenProbedScheduleTime:
                probe: Cron
                timeFormat: unix           # Parse probe message as Unix timestamp
                leadTime: 5m              # Resume 5 minutes before the next scheduled task
         whenTrafficArrive:
                <traffic conifg>
```

docs/proposals/20260626-sandbox-auto-pause.md
```diff
@@ -0,0 +1,869 @@
...
+  conditions:
+    - type: agents.kruise.io/Active
+      status: "True"
+      reason: Succeeded
+      message: "inactive"
+    - type: agents.kruise.io/Cron
+      status: "True"
+      reason: Succeeded
+      message: "1751373600"      # Next scheduled task time
+  autoPauseStatus:
```
furykerry: consider make it a list of schedules , e.g. 

```
   schedules:
       - reason: probedSchedule
          nextResumeTime: "2026-07-01T17:55:00Z" 
       - reason: startDeadline
          nextPauseTime: "2026-07-01T17:55:00Z" 
```

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,381 @@
...
+	// Message does not match — agent is active, clear any stale NextPauseTime.
+	if !re.MatchString(cond.Message) {
+		klog.V(3).InfoS("auto-pause: agent active (message does not match)", "sandbox", klog.KObj(box), "probe", rule.Probe)
+		recordPauseSchedule(nil, newStatus)
+		return nil
+	}
+
+	// Agent is idle — pause is expected once ThresholdDuration elapses since the
+	// probe last transitioned to the idle state.
+	calculatedPause := metav1.NewTime(cond.LastTransitionTime.Add(rule.ThresholdDuration.Duration))
```
furykerry, line 327: **[Critical] ThresholdDuration smoothing is silently bypassed.** The deadline is computed from `cond.LastTransitionTime`, but `utils.SetSandboxCondition` updates `LastTransitionTime` only when `Status` changes. Since probes report `status=True` on exit 0 and the active/idle distinction lives only in the message, an `"active"` → `"inactive"` flip keeps the stale transition time (possibly when the probe first succeeded, hours ago), so `shouldPause` fires immediately on the first idle report.

I reproduced this: condition True/"active" with LastTransitionTime = now-2h, then message flipped to "inactive" through the same sync path → `evaluatePauseSchedule` returned a pause time already ~1h45m in the past. The existing `TestHandleAutoPause_RunningInactiveThresholdNotReached` passes only because it constructs LastTransitionTime directly.

Suggested fix: treat message changes as transitions for probe conditions (e.g., in `syncConditions`, refresh LastTransitionTime when the message changed), or track an explicit idle-since anchor; then add a regression test driving the condition through `syncConditions` before `handleAutoPause`.

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,381 @@
...
+			"sandbox", klog.KObj(box), "message", cond.Message)
+		// Probe is True but message is not a valid timestamp — clear any stale
+		// NextResumeTime so the status does not retain an obsolete schedule.
+		recordResumeSchedule(nil, newStatus)
+		return nil
+	}
+
+	resumedAt := time.Unix(timestamp, 0)
+	// LeadTime defaults to 5m via CRD defaulter (+kubebuilder:default="5m").
+	resumedAt = resumedAt.Add(-rule.LeadTime.Duration)
```
furykerry: **[Critical] Nil-pointer panic when `LeadTime` is nil.** `rule.LeadTime.Duration` is dereferenced here (and again in tryResume's `AutoResumed` event) without a nil check. The `+kubebuilder:default="5m"` is applied by the API server at write time only; Sandboxes created before this CRD version, or objects assembled in-process, can carry `LeadTime == nil` and crash the controller reconcile.

```go
leadTime := 5 * time.Minute
if rule.LeadTime != nil {
    leadTime = rule.LeadTime.Duration
}
resumedAt = resumedAt.Add(-leadTime)
```
Please also add a test case with `LeadTime: nil` (current tests always set it explicitly, hiding the gap).

api/v1alpha1/sandbox_types.go
```diff
@@ -201,6 +219,138 @@ type SandboxLifecycle struct {
...
+	// Condition message (stdout). When the message matches, the Agent is
+	// considered inactive. When it does not match, the Agent is considered
+	// active and the sandbox stays Running.
+	// +kubebuilder:validation:Required
+	MessageRegex string `json:"messageRegex"`
+
+	// ThresholdDuration is the minimum time the probe's Condition message
+	// must continuously match MessageRegex before the sandbox is paused.
+	// Measured from the Condition's lastTransitionTime.
+	// If nil, the sandbox is paused immediately when the message matches.
```
furykerry: **[Critical] API contract contradicts implementation.** This doc (mirrored into the CRD) says a nil ThresholdDuration pauses immediately, but `evaluatePauseSchedule` treats nil as an **invalid rule** — it emits an `InvalidPauseRule` warning event and never pauses. Users following the documented contract get a policy that silently does nothing. Since CRD docs are the user contract, please either implement immediate pause on nil (note the ThresholdDuration rework needed for the stale-LastTransitionTime issue) or update the doc and mark the field required.

furykerry (commented):
Code review of the auto-pause implementation (build + `go test ./pkg/controller/sandbox/... ./pkg/controller/sandbox/core/` verified green locally; Critical #1 additionally reproduced with a targeted test). Inline comments are anchored to the relevant lines; summary below.

## Critical Issues

1. **ThresholdDuration is effectively bypassed — sandbox pauses immediately on the first "inactive" report** (`auto_pause.go` evaluatePauseSchedule): the pause deadline is `cond.LastTransitionTime + ThresholdDuration`, but `utils.SetSandboxCondition` only refreshes `LastTransitionTime` on **Status** changes. Probes report `status=True` on exit 0; the active/idle distinction lives only in the message, so an `"active"` → `"inactive"` flip keeps the stale transition time (often hours old) and `shouldPause` fires immediately. Reproduced: condition True/"active" with LastTransitionTime=now-2h, message flipped to "inactive" → computed pause time already ~1h45m in the past. Existing tests only pass because they construct LastTransitionTime directly; nothing exercises the real sync path.
2. **Nil-pointer panic when LeadTime is not defaulted** (`auto_pause.go` evaluateResumeSchedule + tryResume event): `rule.LeadTime.Duration` is dereferenced without a nil check; the `+kubebuilder:default="5m"` only applies at API-server write time, so objects created before this CRD version (or built in-process) can crash the whole sandbox controller.
3. **API doc contradicts implementation for `ThresholdDuration == nil`**: the field doc/CRD says "If nil, the sandbox is paused immediately when the message matches", but the code treats nil as an invalid rule (warning event, never pauses). Pick one behavior and align doc + code.
4. **Paused sandbox can stall with no requeue** (`auto_pause.go` tryResume): when the resume probe condition is missing/not-True, `evaluateResumeSchedule` returns nil and tryResume returns requeue 0 — while the already-recorded `Status.Schedules[].NextResumeTime` is ignored. A paused pod produces no probe events, so nothing re-triggers reconcile and the sandbox stays Paused forever. Fall back to the recorded schedule and guarantee a bounded requeue.

## Warnings

1. `AutoPauseControllerGate` defaults to **true**; the proposal states default **false** for this alpha gate.
2. Probe injection/sync (`EnsureProbe`/`InjectProbe`) runs **regardless of the feature gate** — only the decision path is gated. Either gate pod-mutating behavior too, or document the scope.
3. `Probe` CRD schema is `Schemaless` + `PreserveUnknownFields`: the CRD validates only `name`/`containerName`; typos or invalid handlers are accepted at the API and surface only later as `ProbeValid=False`. The proposal claims non-exec handlers are "rejected by webhook validation", but this PR contains no webhook changes.
4. New controller code uses bare `klog.InfoS`/`klog.ErrorS`; this controller's convention (and AGENTS.md) is `klog.FromContext(ctx)` — thread ctx through evaluate*/record* helpers.
5. `hasProbeConditionChanged` iterates only **new** conditions; a removed probe condition does not enqueue a reconcile, so condition-removal bookkeeping in `syncConditions` may lag.
6. PR body states "documentation-only PR; no code changes" (~4.9k lines of code included), and the proposal still uses old API names (`WhenIdleProbeFires`/`IdleProbeFireRule`), references removed concepts (`startDeadline`, `MessageUnix`), marks unit tests as TODO though they're included, and states the gate defaults false. Please sync the description/proposal with the implementation.
7. `handleAutoPause` switches on stale `box.Status.Phase` instead of the freshly computed `newStatus.Phase`, delaying the first pause evaluation by one reconcile after a transition to Running.

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,381 @@
...
+// When not yet reached, it returns a requeue for the resume time.
+func (r *SandboxReconciler) tryResume(
+	ctx context.Context,
+	box *agentsv1alpha1.Sandbox,
+	newStatus *agentsv1alpha1.SandboxStatus,
+	now metav1.Time,
+	resumeTime *metav1.Time,
+) (time.Duration, error) {
+	if !shouldResume(resumeTime, now) {
+		return requeueAfter(now, resumeTime), nil
```
furykerry, line 600: **[Critical] Paused sandbox can stall indefinitely with no requeue.** When the resume probe condition is missing or not True, `evaluateResumeSchedule` returns nil and this returns `requeueAfter(now, nil) == 0` — no requeue at all — while the already-recorded `Status.Schedules[].NextResumeTime` is ignored in the decision. A paused pod generates no probe-condition events, so nothing re-triggers reconciliation and the sandbox stays Paused forever (e.g., pod recreated during pause, runtime not reporting).

Suggested fix: fall back to the recorded schedule entry when the live condition is unavailable, and return a bounded default polling interval while Paused with a resume policy instead of 0.

pkg/features/features.go
```diff
@@ -82,6 +89,7 @@ var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	SecurityIdentityProviderGate:           {Default: false, PreRelease: featuregate.Alpha},
 	SandboxPauseCheckpointGate:             {Default: false, PreRelease: featuregate.Alpha},
 	CommitGate:                             {Default: false, PreRelease: featuregate.Alpha},
+	AutoPauseControllerGate:                {Default: true, PreRelease: featuregate.Alpha},
```
furykerry: **[Warning] Gate default contradicts the proposal.** The proposal's Upgrade Strategy says "Feature gate `AutoPauseController` (default: `false`)", but this defaults to `true`. An alpha gate defaulting on enables a new behavior path — including patching user `Spec.Paused` — for everyone on upgrade. Please default to `false` or update the proposal with the rationale.

pkg/controller/sandbox/core/common_control.go
```diff
@@ -188,6 +190,12 @@ func (r *commonControl) EnsureSandboxUpdated(ctx context.Context, args EnsureFun
 			return nil
 		}
 	}
+	// Ensure probe configurations are valid and take effect on the pod.
+	if err := r.probeManager.EnsureProbe(ctx, box, pod, newStatus); err != nil {
```
furykerry, line 196: **[Warning] Probe injection/sync is not gated.** `EnsureProbe` (pod annotation patching, condition mirroring, `ProbeValid` condition) — like `InjectProbe` in pod generation — runs regardless of `AutoPauseControllerGate`; only the decision path in Reconcile is gated. If the gate exists for rollback safety, pod-mutating behavior should respect it too; otherwise please document that the gate covers only pause/resume decisions.

api/v1alpha1/sandbox_types.go
```diff
@@ -201,6 +219,138 @@ type SandboxLifecycle struct {
...
+	// ContainerName specifies which container to execute the probe in.
+	// If empty, defaults to the first container in the pod spec.
+	// +optional
+	ContainerName string `json:"containerName,omitempty"`
+
+	// Probe embeds corev1.Probe inline. Currently only exec, periodSeconds,
+	// timeoutSeconds, and failureThreshold are actively used.
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
```
furykerry: **[Warning] Schemaless probe schema removes API-level validation.** With `Schemaless` + `PreserveUnknownFields`, the generated CRD validates only `name`/`containerName` — typos like `exce:` or invalid `periodSeconds` are silently accepted and only surface later as `ProbeValid=False`. Also, the proposal claims non-exec handlers are "rejected by webhook validation", but this PR contains no webhook changes. Consider letting controller-gen emit the real `corev1.Probe` schema, and add webhook validation (or update the proposal to match the condition-based reporting).

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,381 @@
...
+) (time.Duration, error) {
+	if !shouldPause(pauseTime, now) {
+		return requeueAfter(now, pauseTime), nil
+	}
+
+	oldPaused := box.Spec.Paused
+	if err := r.patchSandboxPaused(ctx, box, true); err != nil {
+		return 0, err
+	}
+	klog.InfoS("auto-pause: pausing sandbox", "sandbox", klog.KObj(box))
```
furykerry: **[Warning] Logging convention.** New code uses bare `klog.InfoS`/`klog.ErrorS`/`klog.V(3).InfoS` throughout `auto_pause.go` and `core/pod_probe.go`, while the surrounding controller code uses `klog.FromContext(ctx)` (and AGENTS.md requires context-based logging for correlation/tracing). Please thread ctx through the helpers (evaluate*/record* currently take no ctx) and switch to `klog.FromContext(ctx)`.

pkg/controller/sandbox/pod_event_handler.go
```diff
@@ -130,6 +132,34 @@ func isActivePodUpdate(oldObj, newObj *corev1.Pod) bool {
...
+	// sandbox controller is reconciled when probe results change.
+	if hasProbeConditionChanged(&oldObj.Status, &newObj.Status) {
+		return true
+	}
+	return false
+}
+
+// hasProbeConditionChanged returns true if any condition with type prefixed
+// by agents.kruise.io/ has changed between old and new pod status.
+func hasProbeConditionChanged(oldStatus, newStatus *corev1.PodStatus) bool {
```
furykerry, line 147: **[Warning] Condition removals are not detected.** This only iterates conditions present in the **new** status; a probe condition that existed in old but was removed in new returns false, so no reconcile is enqueued and `syncConditions`' removal bookkeeping may lag until an unrelated event arrives. Consider also returning true when the number of prefixed conditions differs between old and new.

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,381 @@
...
+	if isUnclaimedPoolSandbox(box) {
+		return 0, nil
+	}
+
+	now := metav1.Now()
+	// Evaluate and record the next pause/resume times in Schedules.
+	pauseTime := r.evaluatePauseSchedule(box, newStatus)
+	resumeTime := r.evaluateResumeSchedule(box, newStatus)
+
+	switch box.Status.Phase {
```
furykerry: **[Warning] Stale phase.** This switches on `box.Status.Phase` while `newStatus.Phase` has already been computed in this cycle; a sandbox that transitioned to Running this cycle waits one extra reconcile before its first pause evaluation. Consider switching on `newStatus.Phase` (or document the one-cycle lag as intentional).

AiRanthem:
The current design has two pause executors: `handleAutoPause` patches `spec.paused`, and `handlePauseTimeout` is skipped whenever `AutoPausePolicy` is set. That avoids fighting over `spec.paused`, but it also skips the mature pause path — including paused-retention (`ShutdownTime` extension and `PauseTime` alignment). A probe-driven pause can therefore freeze a sandbox whose original `ShutdownTime` has already passed, and the next reconcile will delete it.

If we explicitly decide that **controller auto-pause covers E2B** on sandboxes that have `AutoPausePolicy.Pause`, pause scheduling and execution can converge on `spec.pauseTime`:

- While Running, `PauseTime` means “next freeze instant”: idle → `lastTransitionTime + thresholdDuration`; busy / fail-closed → clear it.
- `handlePauseTimeout` becomes the only writer of `spec.paused=true` (and the only place that applies retention). Drop `tryPause`’s direct patch and `status.schedules[].nextPauseTime` — they duplicate spec.
- `handlePauseTimeout` must **not** skip `PauseTime` in this mode.

That only works if ownership is restated together, not just in the controller:

1. **`AutoPausePolicy.Pause` is the durable auto-pause mode.** Today `PauseTime != nil` means `autoPause=true` (`ParseTimeout`). Busy sandboxes would clear `PauseTime`, so E2B must key mode off the policy, not field presence.
2. **E2B Create/Connect/SetTimeout/Resume must not write `PauseTime` while the policy is set.** Otherwise the dual authority just moves from `spec.paused` to `PauseTime`. Resume still flips `paused=false`; it should not race the controller on `PauseTime`.
3. **`ShutdownTime` while Running needs an explicit rule.** Taking over pause but leaving the original kill clock will delete a busy sandbox before any freeze. Either skip shutdown until a probe-driven pause (then retention rewrites `ShutdownTime`), or the controller extends it. Retention still applies only at the actual pause instant; no annotation ⇒ do not touch `ShutdownTime`.
4. **Resume stays out of this convergence.** There is no `spec.resumeTime`. Keep `NextResumeTime` in status and `tryResume` / API Connect as the unpause path.

Net: with those contracts, pause is “set `spec.pauseTime` + reuse `handlePauseTimeout`”. Without them, writing `PauseTime` from probes still conflicts with E2B.

AiRanthem:
Probe stdout is a workload self-report, and this design treats it as the pause/resume signal. Two issues follow.

**Executor and status writer sit in the agent’s trust domain.** Exec runs in the main container; envd is started from `postStart` on `/mnt/envd`. The agent can replace the binaries the control plane calls. A replaced envd can steal CSI Secrets and identity tokens sent on `/init`, mount, and recycle, and can lie that recycle finished so the warm pool hands a tainted instance to the next tenant. kubelet/kruise-daemon will also re-exec the probe every `periodSeconds`, so malware can keep hitting the CSI sidecar, traffic-proxy, and localhost runtime without the agent process. If reporting later lands in that same envd with `pods/status` credentials, the agent can skip the probe and write `Pod.Status.Conditions` directly; the controller only matches `agents.kruise.io/`.

**`Condition.message` is treated as pause/resume authority.** The agent can see the probe command in `/proc` and can always print `active` (fail-closed is also active) so the sandbox never pauses, or print an arbitrary unix timestamp so resume happens when it chooses.

Fail-closed covers probe failure, not a successful lie. If the adversary is an in-sandbox agent, probe output is a hint — and the executor plus `pods/status` credentials must not live in the agent container.

HARSHRAJ2789:
Built and tested a fix for Critical #1 against this branch. The diagnosis is settled, so what
follows is the patch and the evidence behind it.

Reproduced through the real condition sync path, since a hand-built condition is what hides it. `SetSandboxCondition` to True/"active" with a two-hour-old transition time, then
again to True/"inactive": status never changes, so the helper keeps the old timestamp.

```
LastTransitionTime age after the idle flip = 2h
computed pause time                        = 1h59m in the past
shouldPause right now                      = true
```

The deadline has to come from something that dates the idle period. `Status.Schedules` already
carries `NextPauseTime` for this rule, `tryPause` clears it once a pause fires, and the active
branch clears it through `recordPauseSchedule(nil, ...)`, so it can hold the deadline with no new
API surface:

```go
	if sched := findSchedule(newStatus, agentsv1alpha1.ScheduleReasonProbedIdle); sched != nil && sched.NextPauseTime != nil {
		klog.V(3).InfoS("auto-pause: agent still idle, keeping scheduled pause", "sandbox", klog.KObj(box), "probe", rule.Probe, "pauseTime", *sched.NextPauseTime)
		return sched.NextPauseTime
	}
	calculatedPause := metav1.NewTime(now.Add(rule.ThresholdDuration.Duration))
```

`now` is threaded down from `handleAutoPause` and no longer read again inside. With a second
`metav1.Now()` here the first requeue comes out a few nanoseconds over `ThresholdDuration`, because
the stamp is taken after the instant the requeue is measured against.

The same reproduction on the patched branch, with a 30 minute threshold:

```
LastTransitionTime age after the idle flip = 2h
computed pause time                        = 30m from now
shouldPause right now                      = false
```

Three existing tests carry the old contract and move with it. Two of them read as coverage of the
pause path, so they are worth naming:

- `TestEvaluatePauseSchedule` asserted `lastTransition + threshold`, now asserts `now + threshold`.
- `TestHandleAutoPause_RunningInactiveThresholdReached_Pause` and
  `TestHandleAutoPause_RunningAlreadyPaused_NoEvent` used a stale `LastTransitionTime` to mean
  "threshold elapsed". They now seed `Schedules[].NextPauseTime` in the past, which is what an
  earlier reconcile leaves behind.
- `TestHandleAutoPause_RunningInactiveThresholdNotReached` asserted the requeue was below the
  threshold. It is now exactly the threshold, so it asserts equality.

Two regression tests come with it: one drives the message-only flip through `SetSandboxCondition`
and fails without the change, one checks the deadline holds while the agent stays idle and restarts
in full after an active report. `./pkg/controller/sandbox/...` passes and `golangci-lint` reports
0 issues.

One consequence worth a deliberate decision: a `ThresholdDuration` edited while a sandbox
is already idle does not move a deadline that is already stamped, and takes effect on the next idle
period. Recomputing on every reconcile would restore the drift this removes, so it seemed better to
leave it and say so.

Happy to open this as a PR against the branch if that is easier than applying it.

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,381 @@
...
+	condType := agentsv1alpha1.ProbeConditionPrefix + rule.Probe
+	cond := utils.GetSandboxCondition(newStatus, condType)
+	if cond == nil || cond.Status != metav1.ConditionTrue {
+		klog.V(3).InfoS("auto-pause: resume probe condition not available or not true",
+			"sandbox", klog.KObj(box), "probe", rule.Probe)
+		return nil
+	}
+
+	// Parse message as Unix timestamp
+	timestamp, err := strconv.ParseInt(strings.TrimSpace(cond.Message), 10, 64)
```
furykerry: Issue: ProbedScheduleTimeRule.TimeFormat (enum unix) is documented as the format selector, but evaluateResumeSchedule unconditionally parses the message as a Unix timestamp. If a future format is added (or a client sets the field expecting validation), behavior silently ignores it.

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,381 @@
...
+) *metav1.Time {
+	policy := box.Spec.AutoPausePolicy
+	if policy == nil || policy.Pause == nil || policy.Pause.WhenProbedIdleState == nil {
+		// No pause rule configured
+		return nil
+	}
+
+	rule := policy.Pause.WhenProbedIdleState
+	// Validate required fields upfront
+	if rule.Probe == "" || rule.MessageRegex == "" || rule.ThresholdDuration == nil {
```
furykerry, line 281: Issue: The CRD/type comment says "If nil, the sandbox is paused immediately when the message matches," but evaluatePauseSchedule treats nil as an invalid rule, emits an InvalidPauseRule event and never pauses. The CRD also marks the field +optional.

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -917,6 +932,13 @@ func (r *SandboxReconciler) handlePauseTimeout(ctx context.Context, box *agentsv
 	if box.Spec.PauseTime == nil || box.Spec.Paused {
 		return ctrl.Result{}, false, nil
 	}
+	// When AutoPausePolicy with Pause/Resume is active, the controller takes
+	// over pause decisions based on probe results. Skip the one-shot PauseTime
+	// timer to avoid conflicts.
+	if hasActiveAutoPausePolicy(box) {
```
furykerry: Issue: handlePauseTimeout skips the user-set one-shot Spec.PauseTime whenever hasActiveAutoPausePolicy is true — but that returns true if either Pause or Resume is configured. A user who configures only resume.whenProbedScheduleTime silently loses the PauseTime deadline.

furykerry (commented):
## Critical Issues (Must Fix)

Following a full review of this PR, two correctness issues were found in the probe → auto-pause decision path. Details below.

---

### C1. Idle threshold is measured against a stale `LastTransitionTime` — sandbox can be paused immediately instead of after `ThresholdDuration`

- **Location**: `pkg/controller/sandbox/auto_pause.go:187` (+ `core/pod_probe.go` `syncConditions`, + `pkg/utils/utils.go` `SetSandboxCondition`)
- **Issue**: `evaluatePauseSchedule` computes `pauseTime = cond.LastTransitionTime + ThresholdDuration` from the **sandbox-side** condition. But `syncConditions` writes it through `utils.SetSandboxCondition`, which only refreshes `LastTransitionTime` when **Status** changes:

  ```go
  // pkg/utils/utils.go
  if currentCond.Status != condition.Status || currentCond.LastTransitionTime.IsZero() {
      currentCond.LastTransitionTime = condition.LastTransitionTime
  }
  ```

  The idle transition is a **message-only** change (probe stays `True`/`Succeeded`; stdout flips from "active" to "idle"). The pod-side `lastTransitionTime` is therefore discarded and the sandbox condition keeps the transition time from when the probe first became `True` (often pod startup).
- **Why**: Documented semantics ("ThresholdDuration is the minimum time the message must **continuously** match, measured from the Condition's lastTransitionTime") are violated. Example: probe `True` since 10:00, agent goes idle at 11:00, threshold 30m → controller computes pauseTime = 10:00 + 30m = 10:30, already past → **pauses immediately at 11:00**. This also makes auto-resume → immediate re-pause loops possible after Stop-pause/resume, where stale pre-pause conditions survive in status.
- **Fix**: Measure from the pod condition directly, or preserve the pod's transition time in the probe sync path, e.g.:

  ```go
  // option A: evaluate against the pod condition
  podCond := findPodCondition(pod, condType)
  if podCond == nil { ... }
  calculatedPause := metav1.NewTime(podCond.LastTransitionTime.Add(rule.ThresholdDuration.Duration))
  ```

  ```go
  // option B: in syncConditions, bypass SetSandboxCondition's preservation for probe
  // conditions and always mirror podCond.LastTransitionTime
  ```

  Suggested regression test: probe True+non-matching since T0, message becomes matching at T1, assert pauseTime == T1 + threshold (not T0 + threshold).

---

### C2. Stale probe conditions are retained when the pod condition disappears — fail-closed design is defeated

- **Location**: `pkg/controller/sandbox/core/pod_probe.go:176-189` (case 1) and `pkg/controller/sandbox/pod_event_handler.go:146-163`
- **Issue**: In `syncConditions`, when the pod condition is gone but a sandbox condition exists, the stale condition is kept:

  ```go
  if podCond == nil {
      existing := utils.GetSandboxCondition(newStatus, condType)
      if existing == nil { // only sets Unknown for brand-new probes
          ...
      }
      continue // stale True/"idle" condition persists
  }
  ```

  and `hasProbeConditionChanged` only iterates **new** conditions, so a condition **removal** on the pod never triggers a reconcile (the test at `auto_pause_test.go:1233` even documents this as intentional: "not detected").
- **Why**: The proposal explicitly promises fail-closed behavior when probe results are unavailable. Concrete failure path: sandbox paused via Stop strategy while idle → pod deleted; on resume the new pod has no probe conditions yet, but the sandbox still carries the pre-pause `True`/idle condition with a stale transition time; combined with C1 this can **immediately re-pause a freshly resumed sandbox**. Likewise, a runtime that drops/clears conditions leaves the controller deciding on stale data indefinitely.
- **Fix**:

  ```go
  if podCond == nil {
      utils.SetSandboxCondition(newStatus, metav1.Condition{
          Type: condType, Status: metav1.ConditionUnknown,
          Reason: agentsv1alpha1.ProbeReasonPending,
          Message: "probe result not yet available",
      }) // unconditionally — overwrite stale results
      continue
  }
  ```

  and detect deletions in `hasProbeConditionChanged` (compare counts / iterate old conditions too):

  ```go
  if len(oldConds) != <number of probe conds in newStatus> { return true }
  ```

api/v1alpha1/sandbox_types.go
```diff
@@ -201,6 +219,138 @@ type SandboxLifecycle struct {
...
+	// +kubebuilder:validation:Required
+	Probe string `json:"probe"`
+
+	// TimeFormat indicates the format of the probe's Condition message for
+	// parsing as a timestamp. When set to "unix", the controller parses the
+	// message as a Unix timestamp (seconds since epoch) and sets NextResumeTime
+	// to timestamp - LeadTime.
+	// +optional
+	// +kubebuilder:validation:Enum=unix
+	TimeFormat string `json:"timeFormat,omitempty"`
```
furykerry, line 331: openhermes return schedule like this, consider add datetime to the enum: 

```
    "created_at": "2026-08-20T09:00:00+08:00",
      "next_run_at": "2026-08-29T08:00:00+08:00",
      "last_run_at": "2026-08-28T08:00:05+08:00",
```

AiRanthem:
Review of the new tests (complexity only):

- In `auto_pause_test.go`, remove the duplicate nil-label case and keep only one of the valid resume-timestamp cases that exercise the same path and expectation.
- Remove the no-action `handleAutoPause` integration cases that repeat branches already covered by `TestEvaluatePauseSchedule` and `TestEvaluateResumeSchedule`. Keep the requeue, pause, and resume integration paths as the wiring checks.
- Remove the direct `ensureSchedule` and `findSchedule` tests; their create, match, and miss behavior is already exercised through `TestRecordPauseSchedule` and `TestRecordResumeSchedule`.
- In `pod_probe_test.go`, collapse the repeated valid/default-container/nil-annotations injection cases into one, and remove the individual invalid-probe injection cases that duplicate `TestValidateProbes`. Keep one mixed-valid/invalid case to prove all-or-nothing injection.
- Replace the `expectPatch` / `expectRemoved` assertion tree with direct expectations for the final annotation presence and value. The current test does not observe whether a patch call happened, so modeling that unobserved call adds complexity without coverage.
- Remove the "condition unchanged" and "existing Unknown" sync cases: they assert only final values already covered by other cases and do not observe the idempotence behavior named by the tests.
- In `pod_event_handler_test.go`, remove the new deletion-timestamp, `Generic`, `ResetComplete`, and resize tests because those paths are unchanged by this PR. Add a single probe-condition smoke case to the existing `TestIsActivePodUpdate` table and let `TestHasProbeConditionChanged` retain the detailed helper coverage.

Net: roughly 590 lines can be removed while retaining one focused regression check for each distinct behavior.

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,381 @@
...
+		r.recorder.Event(box, corev1.EventTypeWarning, "InvalidPauseRule",
+			fmt.Sprintf("pause rule has missing required field(s): probe=%q, messageRegex=%q, thresholdDuration=%v",
+				rule.Probe, rule.MessageRegex, rule.ThresholdDuration))
+		return nil
+	}
+
+	// Compile MessageRegex upfront to catch invalid patterns early
+	re, err := regexp.Compile(rule.MessageRegex)
```
AiRanthem: Consider caching compiled regexps in the reconciler to avoid recompiling on every pass.

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,381 @@
...
+	if err != nil {
+		klog.ErrorS(err, "auto-pause: invalid messageRegex", "sandbox", klog.KObj(box), "regex", rule.MessageRegex)
+		r.recorder.Event(box, corev1.EventTypeWarning, "InvalidMessageRegex",
+			fmt.Sprintf("Invalid messageRegex %q: %v", rule.MessageRegex, err))
+		return nil
+	}
+
+	condType := agentsv1alpha1.ProbeConditionPrefix + rule.Probe
```
AiRanthem: This `ProbeConditionPrefix + rule.Probe` concatenation appears in many places. Consider extracting it into a `rule.ToConditionType()` method or a shared `RuleToConditionType` helper.

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	// Parse message as Unix timestamp
+	timestamp, err := strconv.ParseInt(strings.TrimSpace(cond.Message), 10, 64)
+	if err != nil || timestamp <= 0 {
+		klog.InfoS("auto-pause: failed to parse resume probe message as unix timestamp",
+			"sandbox", klog.KObj(box), "message", cond.Message)
+		// Probe is True but message is not a valid timestamp — clear any stale
+		// NextResumeTime so the status does not retain an obsolete schedule.
+		recordResumeSchedule(nil, newStatus)
```
AiRanthem: If the condition message is malformed (e.g. due to a bug in the probe script), all information is silently discarded here, which makes the issue hard to troubleshoot. A warning log/event would help.

pkg/controller/sandbox/pod_event_handler.go
```diff
@@ -130,6 +132,34 @@ func isActivePodUpdate(oldObj, newObj *corev1.Pod) bool {
...
+	for _, c := range oldStatus.Conditions {
+		if strings.HasPrefix(string(c.Type), prefix) {
+			oldConds[string(c.Type)] = c
+		}
+	}
+	for _, c := range newStatus.Conditions {
+		if !strings.HasPrefix(string(c.Type), prefix) {
+			continue
+		}
+		old, exists := oldConds[string(c.Type)]
```
AiRanthem, line 161: This only detects condition additions and modifications, not deletions — a removed probe condition will not trigger a reconcile.

AiRanthem (commented):
Code review comments on the auto-pause/resume implementation.

HARSHRAJ2789 (commented):
Read `c9c2da0` and ran the pause path against it. The stale-anchor problem is solved: `pod_probe.go:204` now advances `LastTransitionTime` on a message-only transition, and a probe that flips from `active` to `inactive` two hours after going green anchors the threshold at the moment of the flip. That was the harder half.

The nil `ThresholdDuration` contract is still the way it was. `sandbox_types.go:299` says:

```go
// If nil, the sandbox is paused immediately when the message matches.
// +optional
ThresholdDuration *metav1.Duration `json:"thresholdDuration,omitempty"`
```

and `auto_pause.go:189` treats nil as a missing required field:

```go
if rule.Probe == "" || rule.MessageRegex == "" || rule.ThresholdDuration == nil {
    r.recorder.Event(box, corev1.EventTypeWarning, "InvalidPauseRule", ...)
    return nil
}
```

Running both cases through `calculatePauseTime` on this commit, with a probe reporting `inactive` against `messageRegex: ^inactive$`:

```
ThresholdDuration = 15m   -> pause scheduled
ThresholdDuration = nil   -> no pause, ever
                             event: Warning InvalidPauseRule
                             pause rule has missing required field(s): thresholdDuration=nil
```

The CRD is what makes this reachable. `agents.kruise.io_sandboxes.yaml:98` lists `required: [messageRegex, probe]`, and `thresholdDuration` carries no default, so the API server accepts a `whenProbedIdleState` written with just those two fields. That is the shape the field doc describes, and it is the shape that silently never pauses.

This is the same thing @furykerry raised on 26 August as `[Critical] API contract contradicts implementation`, with two remedies offered: implement immediate pause on nil, or update the doc and mark the field required. Both are still open at `c9c2da0`, and with 4801 lines in the rewrite it seemed worth confirming which review points the push carried and which it did not.

One smaller thing while you are in the description: it still reads "This is a documentation-only PR; no code changes or unit tests are required", which was true of the proposal but not of this commit.

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,578 @@
...
+		return nil
+	}
+	// Message does not match — agent is active.
+	if !re.MatchString(cond.Message) {
+		logger.V(3).Info("auto-pause: agent active (message does not match)", "sandbox", klog.KObj(box), "probe", rule.Probe)
+		return nil
+	}
+
+	// Agent is idle — pause is expected once ThresholdDuration elapses since the
+	// probe last transitioned to the idle state.
```
furykerry, line 326: we should ignore probe result when the lastTransitionTime is earlier than the AnnotationClaimTime

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,578 @@
...
+// calculatePauseTime returns cond.LastTransitionTime + ThresholdDuration when the
+// agent is currently idle (probe succeeded and message matches MessageRegex), or
+// nil when the sandbox should not be paused: no/invalid rule, probe unavailable
+// or not succeeded (fail-closed), or agent active.
+func (r *SandboxReconciler) calculatePauseTime(
+	ctx context.Context,
+	box *agentsv1alpha1.Sandbox,
+	newStatus *agentsv1alpha1.SandboxStatus,
+) *metav1.Time {
+	logger := klog.FromContext(ctx)
```
furykerry: plz avoid pass sandbox value every time by init a new logger 

```
	log := klog.FromContext(ctx).WithValues("sandbox", klog.KObj(box))
```

pkg/controller/sandbox/auto_pause.go
zmberg: Thanks. I handled this by clamping the mirrored probe condition effective LastTransitionTime to AnnotationClaimTime whenever the Pod timestamp is earlier, rather than discarding the result entirely. Unclaimed pool Sandboxes do not mirror probe result conditions; after claim, an already-stable Pod result can be used immediately, but its idle threshold starts at claim time. Dropping it would require a later condition change and could leave the decision pending indefinitely when the probe output remains unchanged. Implemented in 62d5c82a.

docs/proposals/20260626-sandbox-auto-pause.md
zmberg: I kept leadTime intentionally. The probe reports the workload-owned task time, while startup lead time is an infrastructure/runtime concern and may vary independently. Keeping it explicit avoids forcing every probe script or cron source to pre-adjust timestamps and preserves the original schedule for observability.

pkg/controller/sandbox/auto_pause.go
zmberg: Done in 62d5c82a. calculatePauseTime and calculateResumeTime now bind the sandbox once with klog.FromContext(ctx).WithValues, and the individual log calls no longer repeat the sandbox key.

pkg/controller/sandbox/auto_pause.go
zmberg: I kept this at V(3) rather than emitting a warning/event because an unparsable value is also the normal no-schedule signal (for example, "none") reported every probe period. Warning-level reporting would spam logs and events for an expected steady state. The raw value remains visible in the probe Condition, while NextResumeTime is cleared.

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -962,16 +987,28 @@ func (r *SandboxReconciler) handleShutdownTimeout(ctx context.Context, box *agen
...
+	// in this same reconcile once PauseTime is reached, so without
+	// pauseTimeReached it could not act and we must proceed with deletion.
+	// handleAutoPause instead waits for the idle probe: while it owns the pause
+	// decision the probe, not PauseTime, decides when the sandbox stops, so
+	// ShutdownTime must not delete a sandbox the probe still reports as busy.
 	if _, hasRetention := box.Annotations[agentsv1alpha1.AnnotationReservePausedSandboxDuration]; hasRetention &&
 		!box.Spec.Paused &&
-		pauseTimeReached(box.Spec.PauseTime, now) {
+		box.Spec.PauseTime != nil &&
+		(autoPauseOwnsPause(box) || pauseTimeReached(box.Spec.PauseTime, now)) {
```
furykerry: Issue: The reworked deferral skips deletion when hasRetention && !Paused && PauseTime != nil && (autoPauseOwnsPause(box) || pauseTimeReached(...)). The comment argues "A pending PauseTime is what makes this deferral bounded ... waiting costs at most the distance to that pause." But when autoPauseOwnsPause is true, checkTimers skips handlePauseTimeout entirely (sandbox_controller.go:926-933) — nothing pauses the sandbox at PauseTime. The only pause trigger left is the idle probe.

pkg/controller/sandbox/auto_pause.go
```diff
@@ -0,0 +1,605 @@
...
+// When not yet reached, it returns a requeue for the resume time.
+func (r *SandboxReconciler) tryResume(
+	ctx context.Context,
+	box *agentsv1alpha1.Sandbox,
+	newStatus *agentsv1alpha1.SandboxStatus,
+	now metav1.Time,
+	resumeTime *metav1.Time,
+) (time.Duration, error) {
+	if !shouldResume(resumeTime, now) {
+		return requeueAfter(now, resumeTime), nil
```
furykerry, line 600: Issue:  when the resume probe condition is missing/unparsable (calculateResumeTime → nil), tryResume returns requeueAfter(now, nil) == 0. The recorded NextResumeTime is actively cleared instead of serving as a fallback, and there is no bounded polling, so such a sandbox waits for an unrelated event (or the informer resync).
Recommendation: return a bounded default polling interval while Paused with a resume policy, or fall back to the last recorded schedule entry before clearing it.

PR #616, opened 2026-07-08
e2b add network

ZhaoQing7892:


zmberg:
Could you please share the official E2D documentation corresponding to this feature?

pkg/sandbox-manager/infra/interface.go
```diff
@@ -254,6 +259,10 @@ type Sandbox interface {
 	Request(ctx context.Context, method, path string, port int, body io.Reader) (*http.Response, error) // Make a request to the Sandbox
 	CSIMount(ctx context.Context, driver string, request string) error                                  // request is string config for csi.NodePublishVolumeRequest
 	CreateCheckpoint(ctx context.Context, opts CreateCheckpointOptions) (string, error)
+	CreateSandboxNetwork(ctx context.Context, network SandboxNetworkConfig) error // Create TrafficPolicy CR for the sandbox
```
furykerry: consider change the func name from XXXSandboxNetwork to XXXNetworkPolicy

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,334 @@
...
+	sandboxID := s.GetSandboxID()
+	namespace := s.GetNamespace()
+
+	allowCIDRs, allowDomains := splitAllowOut(network.AllowOut)
+
+	// --- Reconcile TrafficPolicy ---
+	tpList := &agentsv1alpha1.TrafficPolicyList{}
+	if err := k8sClient.List(ctx, tpList,
+		client.InNamespace(namespace),
+		client.MatchingLabels{labelSandboxID: sandboxID},
```
furykerry: plz add an index for TrafficPolicy using sandboxID, so that trafficpolicy list can be more efficient

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,334 @@
...
+	); err != nil {
+		return nil, fmt.Errorf("failed to list TrafficPolicies: %w", err)
+	}
+	if len(tpList.Items) > 0 {
+		tp := &tpList.Items[0]
+		if tp.Spec.Egress != nil {
```
furykerry: what about the ingress setting?

pkg/servers/e2b/services.go
```diff
@@ -90,6 +103,12 @@ func (sc *Controller) DeleteSandbox(r *http.Request) (web.ApiResponse[struct{}],
 	}
 
 	log.Info("sandbox deleted", "id", id)
+
+	// Cleanup network CRs (TrafficPolicy) associated with the sandbox.
+	if err := sbx.DeleteSandboxNetwork(r.Context()); err != nil {
```
furykerry: it is not necessary to delete trafficpolicy manually, k8s gc controller will delete the trafficpolicy anyway

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,334 @@
...
+	if len(config.AllowOut) == 0 && len(config.DenyOut) == 0 {
+		log.Info("no network CRs found for sandbox")
+		return nil, nil
+	}
+
+	return config, nil
+}
+
+// DeleteSandboxNetwork deletes the TrafficPolicy CR associated with the sandbox.
+func (s *Sandbox) DeleteSandboxNetwork(ctx context.Context) error {
```
furykerry: DeleteSandboxNetwork is not necessary

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,334 @@
...
+			GenerateName:    "tp-",
+			Namespace:       namespace,
+			Labels:          map[string]string{labelSandboxID: sandboxID},
+			OwnerReferences: []metav1.OwnerReference{sandboxOwnerRef(owner)},
+		},
+		Spec: agentsv1alpha1.TrafficPolicySpec{
+			Priority: 1000,
+			Selector: metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					labelSandboxID: sandboxID,
```
furykerry: consider using uid of sandbox CR as the matching label, sandbox id may be too long for label value

pkg/servers/e2b/network.go
```diff
@@ -0,0 +1,149 @@
...
+
+// labelSandboxID is the label key used to associate TrafficPolicy CRs with sandbox.
+const labelSandboxID = agentsv1alpha1.AnnotationSandboxID
+
+// allTrafficCIDR represents all IPv4 addresses, used when allowInternetAccess
+// is set to false (equivalent to denyOut: ["0.0.0.0/0"]).
+const allTrafficCIDR = "0.0.0.0/0"
+
+// isCIDROrIP returns true if the entry is a valid CIDR or bare IP address.
+func isCIDROrIP(entry string) bool {
```
furykerry: isCIDROrIP allTrafficCIDR/defaultDenyCIDR is duplicated func, consider extract them to a util package in pkg/utils/network

pkg/servers/e2b/create.go
```diff
@@ -387,6 +420,12 @@ func (sc *Controller) basicSandboxCreateModifier(ctx context.Context, sbx infra.
 	// directly for metadata-only changes (no image/resources) without setting the
 	// InplaceUpdate condition.
 	infra.MergePodLabels(sbx, request.Extensions.Labels)
+
+	// Inject sandbox-id label into pod template so that TrafficPolicy CRs can
+	// select this pod via label selector.
+	infra.MergePodLabels(sbx, map[string]string{
```
furykerry: consider move the label assignment to L416, so no extra map is required

furykerry:
> Could you please share the official E2D documentation corresponding to this feature?

https://e2b.dev/docs/network/internet-access

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,334 @@
...
+	hasAllowOut := len(allowOutCIDRs) > 0 || len(allowOutDomains) > 0
+	rules := make([]agentsv1alpha1.TrafficPolicyRule, 0, 3)
+
+	if hasAllowOut {
+		// Whitelist mode: allow CIDR/IP and FQDN entries, explicit deny, then default deny
+		allowPeers := make([]agentsv1alpha1.TrafficPolicyPeer, 0, len(allowOutCIDRs)+len(allowOutDomains))
+		for _, cidr := range allowOutCIDRs {
+			allowPeers = append(allowPeers, agentsv1alpha1.TrafficPolicyPeer{CIDR: cidr})
+		}
+		for _, fqdn := range allowOutDomains {
```
furykerry, line 65: Issue: The E2B documentation states: "When any domain is used, the default nameserver 8.8.8.8 is automatically allowed to ensure proper DNS resolution." The PR's buildTrafficPolicy function does not implement this automatic DNS allowlist behavior. When a user specifies allowOut: ["api.example.com"] with a default-deny, the sandbox will be unable to resolve DNS because 8.8.8.8 is not in the allow list.

Consider: In buildTrafficPolicy, when allowOutDomains is non-empty and the default-deny rule is being added, automatically include 8.8.8.8/32 in the allow list

pkg/servers/e2b/create.go
```diff
@@ -277,6 +289,17 @@ func (sc *Controller) createSandboxWithClone(ctx context.Context, request models
...
 	log.Info("sandbox cloned", "id", sbx.GetSandboxID(), "sbx", klog.KObj(sbx),
 		"resourceVersion", sbx.GetResourceVersion(), "totalCost", time.Since(start))
+
+	// Create network CRs (TrafficPolicy) if network config is provided.
+	if request.Network != nil {
+		if netErr := sbx.CreateSandboxNetwork(ctx, infra.SandboxNetworkConfig{
+			AllowOut: request.Network.AllowOut,
+			DenyOut:  request.Network.DenyOut,
+		}); netErr != nil {
+			log.Error(netErr, "failed to create network CRs, sandbox is still usable")
```
furykerry: network creation failure should fail the sandbox creation entirely for security-sensitive workloads.

pkg/servers/e2b/models/sandbox.go
```diff
@@ -32,34 +32,47 @@ const (
...
-	TemplateID   string            `json:"templateID"`
-	Timeout      int               `json:"timeout,omitempty"`
-	AutoPause    bool              `json:"autoPause,omitempty"`
-	Metadata     map[string]string `json:"metadata,omitempty"`
-	EnvVars      EnvVars           `json:"envVars,omitempty"`
-	VolumeMounts []VolumeMount     `json:"volumeMounts,omitempty"`
+	TemplateID          string                `json:"templateID"`
+	Timeout             int                   `json:"timeout,omitempty"`
+	AutoPause           bool                  `json:"autoPause,omitempty"`
+	AllowInternetAccess *bool                 `json:"allow_internet_access,omitempty"`
```
furykerry, line 59: Issue: The new field uses snake_case json:"allow_internet_access", while every other E2B model field in this repo is camelCase (templateID, envVars, metadata) and the sibling network fields are camelCase (allowOut, denyOut). E2B's HTTP API surface for these options is camelCase (allowInternetAccess); allow_internet_access/allow_out are the Python SDK parameter names, not the wire format.

Recommendation: Confirm the exact wire format the SDK sends; align to the repo/E2B convention to avoid the toggle silently no-op'ing when the field fails to bind:

pkg/servers/e2b/models/sandbox.go
```diff
@@ -32,34 +32,47 @@ const (
...
 }
 
+type SandboxNetworkConfig struct {
+	AllowOut []string `json:"allowOut,omitempty"`
+	DenyOut  []string `json:"denyOut,omitempty"`
+}
+
+type SandboxNetworkUpdateConfig struct {
+	AllowInternetAccess *bool `json:"allow_internet_access,omitempty"`
+	*SandboxNetworkConfig
```
furykerry: Issue: Embedding *SandboxNetworkConfig relies on encoding/json's field-promotion + auto-allocation. It works for the current shapes but is fragile and non-obvious for an API request model.
Recommendation: Use explicit fields for clarity and stability:

ZhaoQing7892: done!

pkg/servers/e2b/create.go
ZhaoQing7892: done!

pkg/sandbox-manager/infra/sandboxcr/network.go
ZhaoQing7892: done!

pkg/servers/e2b/create.go
ZhaoQing7892: using MergeFrom + patch of method ensureSandboxUIDLabel()

pkg/servers/e2b/network.go
ZhaoQing7892: done!

pkg/sandbox-manager/infra/sandboxcr/network.go
ZhaoQing7892: using MergeFrom + patch of method ensureSandboxUIDLabel()

pkg/sandbox-manager/infra/sandboxcr/network.go
ZhaoQing7892: deleted DeleteSandboxNetwork

pkg/servers/e2b/services.go
ZhaoQing7892: deleted DeleteSandboxNetwork

pkg/sandbox-manager/infra/sandboxcr/network.go
ZhaoQing7892: done!

pkg/sandbox-manager/infra/interface.go
ZhaoQing7892: done!

pkg/servers/e2b/create.go
```diff
@@ -303,6 +313,23 @@ func (sc *Controller) createSandboxWithClone(ctx context.Context, request models
...
+
+	// Create network CRs (TrafficPolicy) if network config is provided.
+	// Network policy creation failure must fail the sandbox creation
+	if request.Network != nil {
+		if netErr := sbx.CreateNetworkPolicy(ctx, infra.SandboxNetworkConfig{
+			AllowOut: request.Network.AllowOut,
+			DenyOut:  request.Network.DenyOut,
+		}); netErr != nil {
+			log.Error(netErr, "failed to create network policy, sandbox creation failed",
+				"sandboxID", sbx.GetSandboxID())
```
furykerry: Issue: When CreateNetworkPolicy fails, the function returns a 500 error to the caller, but the sandbox CR has already been created/claimed. The sandbox is left running without network policies — the exact insecure state the PR intends to prevent. Each retry creates a new orphaned sandbox.
PLZ Delete or release the sandbox on failure before returning the error:

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,296 @@
...
+
+	// Read TrafficPolicy to extract allowOut (CIDRs + FQDNs) and denyOut (CIDRs)
+	tpList := &agentsv1alpha1.TrafficPolicyList{}
+	if err := k8sClient.List(ctx, tpList,
+		client.InNamespace(namespace),
+		client.MatchingFields{cache.IndexTrafficPolicySandboxID: sandboxID},
+	); err != nil {
+		return nil, fmt.Errorf("failed to list TrafficPolicies: %w", err)
+	}
+	if len(tpList.Items) > 0 {
```
furykerry: consider return early if tpList is empty to reduce code indent, e.g. 
```
if len(tpList.Items) == 0 {
   return nil, nil 
}
tp := &tpList.Items[0]
if tp.Spec.Egress == nil {
      return nil, nil 
}

```

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,296 @@
...
+							}
+							config.AllowOut = append(config.AllowOut, peer.CIDR)
+						}
+						if peer.FQDN != "" {
+							config.AllowOut = append(config.AllowOut, peer.FQDN)
+						}
+					}
+				case agentsv1alpha1.RuleActionReject:
+					for _, peer := range rule.To {
+						if peer.CIDR == network.AllTrafficCIDR && len(rule.To) == 1 {
```
furykerry: Issue: The read-back logic skips any Reject rule where peer.CIDR == AllTrafficCIDR && len(rule.To) == 1. But buildTrafficPolicy adds both the user's explicit deny entries AND an auto-injected default-deny rule with the same CIDR. When a user explicitly sets denyOut: ["0.0.0.0/0"] (or allowInternetAccess: false), the resulting TrafficPolicy has two identical Reject rules, and SelectNetworkPolicy skips both — the user's intent is lost on round-trip.
Fix: only add the default-deny rule if 0.0.0.0/0 is not already in denyOut:

pkg/servers/e2b/network.go
```diff
@@ -0,0 +1,129 @@
...
+		if entry == network.AllTrafficCIDR {
+			return denyOut
+		}
+	}
+	return append(denyOut, network.AllTrafficCIDR)
+}
+
+// validateAndBuildNetworkConfig is the single entry point for validating raw
+// network parameters and producing a normalized SandboxNetworkConfig ready for CR creation.
+func validateAndBuildNetworkConfig(allowInternetAccess *bool, network *models.SandboxNetworkConfig) (*models.SandboxNetworkConfig, error) {
```
furykerry: Issue: The function signature func validateAndBuildNetworkConfig(allowInternetAccess *bool, network *models.SandboxNetworkConfig) shadows the imported network package (pkg/utils/network). Inside this function, any reference to network. would resolve to the parameter, not the package. Currently no such reference exists inside the function body, but it's fragile and a linter may flag it.
Recommendation: Rename the parameter to netConfig or cfg:

pkg/servers/e2b/network.go
```diff
@@ -0,0 +1,129 @@
...
+		network.DenyOut = applyAllowInternetAccess(allowInternetAccess, network.DenyOut)
+	}
+
+	// Step 2: Return nil if no network rules are needed
+	if network == nil || (len(network.AllowOut) == 0 && len(network.DenyOut) == 0) {
+		return nil, nil
+	}
+
+	// Step 3: Validate denyOut — domains are not supported in deny lists
+	if err := validateDenyOut(network.DenyOut); err != nil {
```
furykerry: validateDenyOut validates that denyOut entries are CIDR/IP, but there's no validation for allowOut entries. Malformed entries in allowOut are silently treated as domains by SplitAllowOut. Consider adding basic FQDN validation for non-CIDR entries

pkg/sandbox-manager/infra/sandboxcr/network.go
ZhaoQing7892: done!

pkg/servers/e2b/create.go
ZhaoQing7892: done!

pkg/sandbox-manager/infra/sandboxcr/network.go
ZhaoQing7892: done

pkg/servers/e2b/network.go
ZhaoQing7892: done

pkg/servers/e2b/network.go
ZhaoQing7892: done

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,300 @@
...
+		return nil
+	}
+	log := klog.FromContext(ctx).WithValues("sandbox", klog.KObj(s))
+	k8sClient := s.Cache.GetClient()
+	sandboxID := s.GetSandboxID()
+	namespace := s.GetNamespace()
+
+	allowCIDRs, allowDomains := network.SplitAllowOut(netConfig.AllowOut)
+
+	if err := s.ensureSandboxUIDLabel(ctx); err != nil {
```
furykerry: consider add the label directly in basicSandboxCreateModifier to save this extra patch

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,300 @@
...
+// UpdateNetworkPolicy updates the TrafficPolicy CR for the sandbox.
+func (s *Sandbox) UpdateNetworkPolicy(ctx context.Context, netConfig infra.SandboxNetworkConfig) error {
+	log := klog.FromContext(ctx).WithValues("sandbox", klog.KObj(s))
+	k8sClient := s.Cache.GetClient()
+	sandboxID := s.GetSandboxID()
+	namespace := s.GetNamespace()
+
+	allowCIDRs, allowDomains := network.SplitAllowOut(netConfig.AllowOut)
+
+	if err := s.ensureSandboxUIDLabel(ctx); err != nil {
```
furykerry: patch sandbox template labels does not necessary means the pod is labeled, sandbox controller can update pod metadata inplace only during claiming process, so for an already claimed sandbox, updatenetworkpolicy may not work at all if sandbox is not labeled previously， consider also update the pod label

api/v1alpha1/sandboxset_types.go
```diff
@@ -36,6 +36,8 @@ const (
 	LabelTemplateHash     = InternalPrefix + "template-hash"
 	// LabelSandboxReservedFailed marks a failed sandbox retained for debugging.
 	LabelSandboxReservedFailed = InternalPrefix + "reserved-failed-sandbox"
+	// LabelSandboxUID is the label key used by TrafficPolicy Spec.Selector to select the sandbox pod.
+	LabelSandboxUID = InternalPrefix + "sandbox-uid"
```
furykerry: sandbox uid cannot be determined before creation, consider using the sandbox name instead, so that we can add the label during creation

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,300 @@
...
+		existing.Spec = newTP.Spec
+		existing.OwnerReferences = newTP.OwnerReferences
+		existing.Annotations = newTP.Annotations
+		if err := k8sClient.Update(ctx, existing); err != nil {
+			return fmt.Errorf("failed to update TrafficPolicy %s: %w", existing.Name, err)
+		}
+		log.Info("TrafficPolicy updated", "name", existing.Name)
+		// Delete any extra TrafficPolicies (shouldn't happen, but clean up)
+		for i := 1; i < len(tpList.Items); i++ {
+			tp := &tpList.Items[i]
```
furykerry, line 190: L211-L222 has no UT coverage yet

pkg/servers/e2b/network.go
```diff
@@ -0,0 +1,149 @@
...
+func applyAllowInternetAccess(allowInternetAccess *bool, denyOut []string) []string {
+	if allowInternetAccess == nil || *allowInternetAccess {
+		return denyOut
+	}
+	for _, entry := range denyOut {
+		if entry == network.AllTrafficCIDR {
+			return denyOut
+		}
+	}
+	return append(denyOut, network.AllTrafficCIDR)
```
furykerry: Issue: Disabling internet only appends 0.0.0.0/0 (IPv4). On dual-stack clusters, IPv6 egress remains permitted, so allowInternetAccess:false does not fully deny internet.

Recommendation: Also append ::/0 (note ContainsAllTrafficCIDR already recognizes ::/0), or document the IPv4-only limitation explicitly.

pkg/sandbox-manager/infra/sandboxcr/network.go
ZhaoQing7892: add labels LabelSandboxName in basicSandboxCreateModifier

api/v1alpha1/sandboxset_types.go
ZhaoQing7892: update LabelSandboxUID to LabelSandboxName

pkg/sandbox-manager/infra/sandboxcr/network.go
ZhaoQing7892: done!

pkg/servers/e2b/network.go
ZhaoQing7892: done

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,268 @@
...
+func sandboxOwnerRef(owner *agentsv1alpha1.Sandbox) metav1.OwnerReference {
+	controller := true
+	blockOwnerDeletion := true
+	return metav1.OwnerReference{
+		APIVersion:         agentsv1alpha1.GroupVersion.String(),
+		Kind:               "Sandbox",
+		Name:               owner.Name,
+		UID:                owner.UID,
+		Controller:         &controller,
+		BlockOwnerDeletion: &blockOwnerDeletion,
```
furykerry, line 45: if blockOwnerDeletion is set to true, it will delay the sandbox creation process, and the small window where sandbox exists without trafficpolicy protection is insecure, consider just set it to false

pkg/servers/e2b/network.go
```diff
@@ -0,0 +1,157 @@
...
+	// Step 4: Validate denyOut — domains are not supported in deny lists
+	if err := validateDenyOut(netConfig.DenyOut); err != nil {
+		return nil, err
+	}
+
+	return netConfig, nil
+}
+
+// UpdateSandboxNetwork replaces the sandbox's network rules with the new configuration.
+func (sc *Controller) UpdateSandboxNetwork(r *http.Request) (web.ApiResponse[struct{}], *web.ApiError) {
```
furykerry, line 89: this func is not covered by UT

pkg/servers/e2b/network.go
```diff
@@ -0,0 +1,157 @@
...
+	for _, entry := range denyOut {
+		if !network.IsCIDROrIP(entry) {
+			return fmt.Errorf("domains are not supported in denyOut: %q is not a valid CIDR or IP address", entry)
+		}
+	}
+	return nil
+}
+
+// applyAllowInternetAccess merges the allowInternetAccess flag into denyOut.
+// When internet access is disabled, both IPv4 (0.0.0.0/0) and IPv6 (::/0).
```
furykerry: The doc comment is a sentence fragment, plz make it complete

pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -0,0 +1,268 @@
...
+			} else {
+				log.Info("TrafficPolicy deleted", "name", tp.Name)
+			}
+		}
+	} else if len(tpList.Items) > 0 {
+		// Update existing TrafficPolicy with new spec and ensure OwnerReference is set
+		existing := &tpList.Items[0]
+		existing.Spec = newTP.Spec
+		existing.OwnerReferences = newTP.OwnerReferences
+		existing.Annotations = newTP.Annotations
```
furykerry: When updating an existing TrafficPolicy, the code replaces the entire Annotations map. If any controller or webhook added annotations to the TrafficPolicy after creation (e.g., last-applied-configuration, cert-manager annotations), they would be silently lost on update. Consider deepcopy,  modifiy the clone and generate an patch

pkg/servers/e2b/network.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/openkruise/agents/pkg/sandbox-manager/infra"
+	"github.com/openkruise/agents/pkg/servers/e2b/models"
+	"github.com/openkruise/agents/pkg/servers/web"
+	"github.com/openkruise/agents/pkg/utils/network"
+)
+
+// validateAllowOut checks that allowOut entries are valid CIDR, IP, or FQDN.
+// Wildcard domains are not supported.
+func validateAllowOut(allowOut []string) error {
```
furykerry, line 39: There's no limit on the number of entries in allowOut/denyOut. A malicious or buggy client could send thousands of entries, creating an extremely large TrafficPolicy CR. Consider enforcing a reasonable maximum (e.g., 100 entries per list).

pkg/utils/network/network.go
```diff
@@ -0,0 +1,104 @@
...
+			domains = append(domains, entry)
+		}
+	}
+	return cidrs, domains
+}
+
+// NormalizeToCIDR converts a bare IP to CIDR notation.
+// IPv4 becomes /32, IPv6 becomes /128.
+// If the entry is already a CIDR, it is returned as-is.
+func NormalizeToCIDR(entry string) string {
```
furykerry, line 59: net.ParseIP("::ffff:1.2.3.4").To4() returns non-nil, so the function appends /32 producing ::ffff:1.2.3.4/32, which is semantically wrong for IPv6 (should be /128). This is an edge case unlikely to be hit in practice, but worth a comment or guard.

pkg/sandbox-manager/infra/sandboxcr/network.go
ZhaoQing7892: done!

pkg/servers/e2b/network.go
ZhaoQing7892: done!

pkg/servers/e2b/network.go
ZhaoQing7892: done!

pkg/sandbox-manager/infra/sandboxcr/network.go
ZhaoQing7892: done!

pkg/servers/e2b/network.go
ZhaoQing7892: Added maxNetworkEntriesPerList =100 constant, enforced in both validateAllowOut and validateDenyOut.

pkg/utils/network/network.go
ZhaoQing7892: Switched from ip.To4() to strings.Contains(entry, ":") to correctly handle IPv4-mapped IPv6 addresses.

PR #627, opened 2026-07-09
Avoid terminating the proxy process on startup listener failures

singhsrijan46:
## Ⅰ. Describe what this PR does

This PR fixes a robustness issue in the proxy server startup path.

If the proxy server itself cannot start at all, exiting the process with `klog.Fatalf` is reasonable because there is no safe way to continue. The issue here is that startup failures in the background listener goroutines were also treated as fatal, which made the process terminate for conditions that are better handled as operational errors.

The change replaces those goroutine-level fatal exits with regular error logging and adds a regression test that verifies startup returns an error when the ext-proc listener port is already in use.

## Ⅱ. Does this pull request fix one issue?

fixes #<issue-number>

## Ⅲ. Describe how to verify it

Run the following test:

```bash
go test ./pkg/proxy -run 'TestServer_Run_ReturnsListenerError|TestHandleRefresh|TestHealthServer' -count=1
```

## Ⅳ. Special notes for reviews

- The change is intentionally small and focused on startup error handling.
- It improves resilience without changing the public behavior for successful startup.

### DIFF
--- a/pkg/proxy/server.go
+++ b/pkg/proxy/server.go
@@ -128,14 +128,14 @@ func (s *Server) Run() error {
 	go func() {
 		klog.InfoS("Starting proxy system server", "address", s.httpSrv.Addr)
 		if err := s.httpSrv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
-			klog.Fatalf("HTTP server failed to start: %v", err)
+			klog.ErrorS(err, "HTTP server failed to start")
 		}
 	}()
 
 	go func() {
 		klog.InfoS("Starting proxy gRPC server", "address", lis.Addr())
 		if err := s.grpcSrv.Serve(lis); err != nil && !errors.Is(err, grpc.ErrServerStopped) {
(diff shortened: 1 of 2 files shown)


pkg/proxy/server.go
```diff
@@ -128,14 +128,14 @@ func (s *Server) Run() error {
 	go func() {
 		klog.InfoS("Starting proxy system server", "address", s.httpSrv.Addr)
 		if err := s.httpSrv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
-			klog.Fatalf("HTTP server failed to start: %v", err)
+			klog.ErrorS(err, "HTTP server failed to start")
```
zmberg, line 131: If the server listener fails, the process should indeed exit. The current code is reasonable and should not be changed to ErrorS.

singhsrijan46: Thanks for the clarification. I agree that exiting on listener startup failure is the correct behavior, and I’ll keep the current fatal handling as-is.

PR #630, opened 2026-07-10
refactor(identity): consume security metadata from sandbox annotations

BH4AWS:
The identity-provider opt-in gate and security-metadata extraction used to read from sbx.Labels. However E2B security metadata (the security.agents.kruise.io/* entries in request.Metadata) is written onto the sandbox annotations by basicSandboxCreateModifier, so the provider never saw it - a silent data-flow break.

Switch the consumer side to annotations to close the gap:
- IsIdentityProviderRequested now reads sbx.GetAnnotations()[AnnotationAgentName]
- ExtractSecurityMetadata now filters sbx.GetAnnotations()
- rename constant LabelAgentName -> AnnotationAgentName (value unchanged: security.agents.kruise.io/agent-name)

This unifies the security metadata source with storage-auth / token-status (already annotation-based), removes the 63-char / DNS-label constraints so richer values can be expressed, and aligns naturally with the request.Metadata -> annotations mapping (no create.go / claim.go changes needed). No label selector depends on agent-name (securitytokenrefresh is gated by the token-status annotation), so the switch is behavior-preserving.

Tests updated to place the opt-in key and security entries under annotations. go build and the identity / sandbox-manager / securitytokenrefresh package tests pass.

Note: the constant rename LabelAgentName -> AnnotationAgentName must be synced in the enterprise (inner) branch.

### DIFF
--- a/pkg/identity/types.go
+++ b/pkg/identity/types.go
@@ -30,15 +30,17 @@ const (
 )
 
 const (
-	// SecurityMetadataPrefix is the prefix for all security-related label/annotations.
+	// SecurityMetadataPrefix is the prefix for all security-related annotations.
 	SecurityMetadataPrefix = "security.agents.kruise.io/"
 	// AgentKeyTokenRefreshStatus is the Sandbox Annotation Key,
 	// used to store the JSON serialized result of TokenRefreshStatus.
 	AgentKeyTokenRefreshStatus = SecurityMetadataPrefix + "token-status"
-	// LabelAgentName is the sandbox Label Key whose presence opts the sandbox
-	// into the identity provider issuance path. Its value carries the logical
-	// agent name that the identity provider uses to mint the security token.
-	LabelAgentName = SecurityMetadataPrefix + "agent-name"
+	// AnnotationAgentName is the sandbox Annotation Key whose presence opts the
+	// sandbox into the identity provider issuance path. Its value carries the
+	// logical agent name that the identity provider uses to mint the security
+	// token. An annotation is used instead of a label so the value is free of
+	// the 63-char / DNS-label constraints and can express richer content.
+	AnnotationAgentName = SecurityMetadataPrefix + "agent-name"
 )
 
 // TokenRequest represents a request to issue an identity-aware access token.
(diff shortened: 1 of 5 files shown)


pkg/identity/types.go
```diff
@@ -35,10 +35,12 @@ const (
...
-	// LabelAgentName is the sandbox Label Key whose presence opts the sandbox
-	// into the identity provider issuance path. Its value carries the logical
-	// agent name that the identity provider uses to mint the security token.
-	LabelAgentName = SecurityMetadataPrefix + "agent-name"
+	// AnnotationAgentName is the sandbox Annotation Key whose presence opts the
+	// sandbox into the identity provider issuance path. Its value carries the
+	// logical agent name that the identity provider uses to mint the security
+	// token. An annotation is used instead of a label so the value is free of
+	// the 63-char / DNS-label constraints and can express richer content.
+	AnnotationAgentName = SecurityMetadataPrefix + "agent-name"
```
furykerry, line 43: The comment in L33 still says "prefix for all security-related label/annotations" but after this PR, all consumers read from annotations exclusively. The word "label" is misleading.

pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -357,9 +355,9 @@ func runClaimPostProcesses(ctx context.Context, sbx *Sandbox, lockType infra.Loc
 }
 
 // processSecurityToken issues and propagates a sandbox security token when the
-// identity provider feature gate and sandbox annotations request it.
+// sandbox annotations opt into the identity provider path.
```
furykerry, line 358: the comment  grammatically broken, consider change it to `// processSecurityToken issues and propagates a sandbox security token when the
// identity provider feature gate and sandbox annotations request it`

pkg/identity/types.go
BH4AWS: fixed

pkg/sandbox-manager/infra/sandboxcr/claim.go
BH4AWS: fixed

PR #635, opened 2026-07-11
docs: propose short and stable Sandbox IDs

AiRanthem:
## What changed

- Add the detailed design specification for short, stable Sandbox IDs derived from the full Kubernetes Sandbox UID.
- Add a reviewer-oriented proposal covering user-visible behavior, persisted label ownership, package boundaries, cache lookup, shared manager/gateway routing, diagnostics, rollout, rollback, observability, and tests.
- Define the mixed-version Route state machine, conditional delete fencing, collision quarantine, and authoritative sweep needed for a one-ID-per-Sandbox migration.

## Why

The current `<namespace>--<name>` Sandbox ID is embedded in E2B-compatible dynamic hostnames. Long namespace and Sandbox names can make the resulting DNS name too long to access, even when the Sandbox itself is healthy.

The proposal uses a 26-character lowercase Base32 representation of the complete 128-bit Sandbox UID and persists the selected value in `agents.kruise.io/sandbox-id`. Unlabeled Sandboxes retain the legacy ID, and the assignment flag remains disabled by default for rollout safety.

## Impact

This PR changes documentation only; it does not change runtime behavior. It establishes the technical-review contract for a later implementation plan, including compatibility and activation gates.

## Validation

- Verified that the branch differs from upstream only by the design spec and proposal.
- Cross-checked the proposal against the design acceptance criteria and migration boundaries.
- Go tests were not run because this PR contains documentation only.

### DIFF
--- a/docs/proposals/20260711-short-sandbox-id.md
+++ b/docs/proposals/20260711-short-sandbox-id.md
@@ -1,0 +1,602 @@
+---
+title: Short and Stable Sandbox IDs
+authors:
+  - "@AiRanthem"
+reviewers: []
+creation-date: 2026-07-11
+last-updated: 2026-07-14
+status: provisional
+---
+
+# Short and Stable Sandbox IDs
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Current Problem](#current-problem)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [User-Visible ID Behavior](#user-visible-id-behavior)
+  - [Short-ID Encoding](#short-id-encoding)
+  - [Persisted State and Label Protection](#persisted-state-and-label-protection)
+  - [Assignment Flow](#assignment-flow)
+  - [Architecture and Ownership](#architecture-and-ownership)
+  - [Cache Lookup](#cache-lookup)
+  - [Shared Routing Model](#shared-routing-model)
+    - [Store Model](#store-model)
+    - [Collision Handling](#collision-handling)
+    - [Delete Semantics](#delete-semantics)
+    - [Asynchronous Targeted Repair](#asynchronous-targeted-repair)
+  - [E2B Diagnostics](#e2b-diagnostics)
+  - [Checkpoint and Pagination Semantics](#checkpoint-and-pagination-semantics)
+  - [Configuration](#configuration)
+- [Compatibility and Upgrade Strategy](#compatibility-and-upgrade-strategy)
+  - [Initial Rollout](#initial-rollout)
+  - [Activation](#activation)
+  - [Rollback Boundary](#rollback-boundary)
+- [Risks and Mitigations](#risks-and-mitigations)
+- [Observability](#observability)
+- [Test Plan](#test-plan)
+- [Alternatives](#alternatives)
+- [Implementation History](#implementation-history)
+
+## Summary
+
+OpenKruise Agents currently identifies a Sandbox with a readable value derived from its Kubernetes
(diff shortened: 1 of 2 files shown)


docs/proposals/20260711-short-sandbox-id.md
```diff
@@ -0,0 +1,586 @@
...
+| infra | Generic mutation/persistence and format-neutral Sandbox capabilities |
+| E2B server | Request validation and response/error presentation |
+
+The backend-neutral `infra.Sandbox` interface no longer exposes `GetSandboxID()` or `GetRoute()`.
+Manager resolves IDs and constructs route projection inputs. Infra adds only the format-neutral
+`GetPodIP()` capability and may retain an opaque Route reader for its existing cache-staleness
+check.
+
+Manager route watch registration and periodic reconciliation move out of `sandboxcr.Infra` into
+the sandbox-manager composition root. Gateway keeps its controller adapter. Both use the same
```
furykerry, line 293: The gateway controller constructs the sandbox ID directly from the reconcile request's namespace/name without reading the Sandbox's label. When short IDs are enabled, this code would still produce the legacy <namespace>--<name> key, causing a mismatch with the short ID stored in the registry by other components， please be more specific about the change of sandbox-gateway

docs/proposals/20260711-short-sandbox-id.md
```diff
@@ -0,0 +1,586 @@
...
+| Full peer delete | Require ObjectKey, ID, UID, and equal/newer comparable RV to match current full ownership; a different UID never deletes the current incarnation, even with a newer RV |
+| ID-only peer delete | Require matching ID and UID plus an equal/newer comparable RV; may delete only the matching ID-only record, never a full record |
+
+If no full record exists, a valid full peer delete may remove a matching ID-only record subject to
+the same ID/UID/RV fence. This prevents a delayed delete for UID A from deleting a recreated UID B
+at the same namespace/name.
+
+#### Authoritative Sweep
+
+Both components run the shared Sweeper once after cache synchronization and then every five minutes
```
furykerry: is it possible to only check the specified routes in doubt, the one "Different UID with equal or unorderable RV". GET is much cheaper than LIST operation

docs/proposals/20260711-short-sandbox-id.md
AiRanthem: Expanded the gateway adapter design in a4070ab2. For a present Sandbox, the reconciler projects a full Route through the injected label-aware resolver and offers it to the shared Store without deriving the registry key from the reconcile request. For NotFound or deleting objects, it deletes authoritatively by ObjectKey; `<namespace>--<name>` is only the mixed-version fallback when no full ObjectKey record exists. The gateway sweep uses the same watch visibility and inclusion predicate.

docs/proposals/20260711-short-sandbox-id.md
AiRanthem: The original List was intended as a low-frequency background cleanup: once after cache synchronization and then every five minutes with jitter, outside the request and normal reconcile paths. However, at the expected scale of hundreds of thousands of Sandboxes, a full direct List per manager/gateway replica is still too expensive. Revised in afba1b1d to remove the periodic full List. When the Store detects a known ambiguity, it quarantines the ID and returns the affected ObjectKeys and mutation generations. The event adapter enqueues them into a deduplicated, rate-limited repair queue and completes event processing immediately. A worker performs direct Gets only for those keys and applies the authoritative result only if the affected-record generation still matches; transient errors retry with backoff. Normal population and missed-event recovery continue to rely on informer synchronization and List/Watch.

AiRanthem: Design changed

PR #642, opened 2026-07-14
feat(identity): defer security token refresh until sandbox is Ready

BH4AWS:
Gate token refresh on sandbox readiness: while a sandbox is not Ready (paused, resuming, pending, upgrading) there is no serving runtime pod to receive a refreshed token, so refreshing would fail at the propagate step and spin the workqueue while burning issuance calls. The reconciler now defers all refresh work until the sandbox is Ready, and the predicate re-enqueues on the notReady->Ready transition so the schedule re-arms itself without a timing requeue.


furykerry (commented):
/lgtm

PR #646, opened 2026-07-15
feat(controller): lazy sandbox finalizer — add on pause, remove on resume

furykerry:

The sandbox controller no longer adds the `agents.kruise.io/sandbox` finalizer to every sandbox by default during reconciliation. Instead:

- **Finalizer is added** only when a sandbox first enters the Paused (hibernation) state (`EnsureSandboxPaused` with `cond == nil`)
- **Finalizer is removed** when the sandbox successfully resumes to Running (`EnsureSandboxResumed` when pod is `Running`)
- **Hash annotation** (for in-place update detection) is decoupled from the finalizer and continues to be applied early in the reconcile loop as before

Running sandboxes do not need a controller-mediated finalizer because their pods are cleaned up via Kubernetes owner-reference garbage collection. The finalizer is only meaningful for paused sandboxes where the pod has already been deleted and the controller must manage lifecycle transitions.

Key changes:
- Renamed `addSandboxFinalizerAndHash` → `addSandboxHashAnnotation`, removed finalizer-adding logic
- Added finalizer patch in `EnsureSandboxPaused` (idempotent via `PatchFinalizer`)
- Added finalizer removal in `EnsureSandboxResumed` (idempotent, retries on failure)
- `EnsureSandboxTerminated` remains unchanged — its `PatchFinalizer(RemoveFinalizerOpType)` is already idempotent

NONE

1. Run unit tests:
   ```bash
   go test -mod=mod ./pkg/controller/sandbox/... -run 'TestSandboxReconciler_AddSandboxHashAnnotation|TestAddSandboxHashAnnotation_PatchError|TestReconcile_SandboxLifecycle_ClearSpecThenDelete|TestCommonControl_EnsureSandboxPaused|TestCommonControl_EnsureSandboxResumed_RemovesFinalizer' -count=1 -v
   ```
2. Verify that a newly created Running sandbox does **not** have the finalizer
3. Verify that a paused sandbox **does** have the finalizer
4. Verify that a resumed sandbox has the finalizer **removed**
5. Verify that a sandbox deleted while paused is properly cleaned up (finalizer ensures controller-mediated termination)

- **Multi-cycle safety**: The Paused condition is removed by `calculateStatus` during the resume transition (`utils.RemoveSandboxCondition`), so on re-pause, `cond == nil` again and the finalizer is correctly re-added. Test cases cover this scenario.
- **Idempotency**: `PatchFinalizer` checks `ContainsFinalizer` before making API calls, so add/remove are both idempotent. If the finalizer patch fails, the function returns error without advancing state, so the next reconcile retries.
- **Backward compatibility**: Sandboxes created under the old behavior that already have the finalizer will continue to work — `EnsureSandboxTerminated` removes it idempotently. Sandboxes currently paused with the finalizer will have it removed on resume.
- **Metrics cleanup**: When a Running sandbox (no finalizer) is deleted, the controller receives NotFound, cleans up expectations, and enqueues metrics cleanup via `r.metricsCleanup.Enqueue()`. Pod cleanup is handled by K8s GC via owner references.

---

The following changes are side modifications not directly related to the main purpose of this PR:

- Updated stale references to `addSandboxFinalizerAndHash` → `addSandboxHashAnnotation` in documentation (`docs/specs/2026-07-02-sandbox-otel-tracing-design-en.md`, `docs/proposals/20260702-sandbox-otel-distributed-tracing-en.md`)

### DIFF
--- a/pkg/controller/sandbox/core/common_control.go
+++ b/pkg/controller/sandbox/core/common_control.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/distribution/reference"
 
@@ -201,6 +202,14 @@ func (r *commonControl) EnsureSandboxPaused(ctx context.Context, args EnsureFunc
 	pod, box, newStatus := args.Pod, args.Box, args.NewStatus
 	cond := utils.GetSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionPaused))
 	if cond == nil {
+		// Add finalizer on first entry into paused state to ensure
+		// controller-mediated cleanup if the sandbox is deleted while paused.
+		if !controllerutil.ContainsFinalizer(box, SandboxFinalizer) {
+			if _, err := utils.PatchFinalizer(ctx, r.Client, box, utils.AddFinalizerOpType, SandboxFinalizer); err != nil {
+				return fmt.Errorf("failed to add finalizer for paused sandbox: %w", err)
+			}
+			klog.InfoS("Add finalizer for paused sandbox", "sandbox", klog.KObj(box))
+		}
 		cond = &metav1.Condition{
 			Type:               string(agentsv1alpha1.SandboxConditionPaused),
 			Status:             metav1.ConditionFalse,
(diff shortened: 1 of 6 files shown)


pkg/controller/sandbox/core/common_control.go
```diff
@@ -274,6 +280,14 @@ func (r *commonControl) EnsureSandboxResumed(ctx context.Context, args EnsureFun
 
 	// when pod is running, transition sandbox from resuming to running
 	if pod.Status.Phase == corev1.PodRunning && isContainersConsistent(pod, box) {
+		// Remove finalizer on successful resume. The finalizer was added
+		// during pause; now that the pod is running again, it is no longer
+		// needed. If removal fails, return without transitioning the phase
+		// so the next reconcile retries.
+		if _, err := utils.PatchFinalizer(ctx, r.Client, box, utils.RemoveFinalizerOpType, SandboxFinalizer); err != nil {
```
zmberg: Can check first before patching? Only patch if it exists / doesn't exist.

PR #648, opened 2026-07-15
feat(gateway): add JWT verification and optional Runtime mTLS

chengzhycn:

- Adds local TrafficAccessToken JWT verification to sandbox-gateway:
  - Loads the identity CA from Kubernetes, performs OIDC discovery, snapshots JWKS public keys, and verifies signatures and required claims.
  - Binds tokens to Sandbox ID and UID, removes the traffic token before proxying, and exposes gateway readiness while the verifier initializes.
  - Keeps UUID authentication as the default and enables JWT mode only when both `enable-auth` and `enable-jwt-auth` are true.
- Adds optional Runtime upstream mTLS to sandbox-gateway:
  - Keeps `runtimeMTLS.enabled=false` by default, preserving the existing plaintext startup and routing behavior without reading the credential Secret.
  - Routes only port `49983` to a dedicated mTLS `ORIGINAL_DST` cluster; all other Sandbox ports remain plaintext.
  - Adds a startup init binary that performs a single client-go `Get` of `ack-agent-identity/ack-agent-identity-sandbox-manager-client-cert`, validates `tls.crt`, `tls.key`, and `ca.crt`, and writes them to an in-memory volume for static Envoy loading.
  - Uses fixed SNI `agentruntime.sandbox.agents.kruise.io` with server SAN validation and fails closed when credentials cannot be loaded.
- Adds Kustomize deployment paths, image/build integration, focused unit tests, and E2E coverage for JWT auth plus Runtime mTLS enabled and disabled scenarios.
- Does not change agent-runtime behavior, token issuance, client token acquisition, Sandbox Manager APIs, or claim/clone behavior.

NONE

- Run package and race tests:

  ```bash
  go test ./pkg/...
  go test -race ./pkg/identity/oidc ./pkg/sandbox-gateway/...
  go vet ./pkg/sandbox-gateway/... ./cmd/sandbox-gateway-cert-init
  ```

- Verify builds and Kustomize deployment rendering:

  ```bash
  make build-sandbox-gateway
  kubectl kustomize config/sandbox-gateway
  kubectl kustomize config/sandbox-gateway-runtime-mtls
  ```

- Run JWT E2E against a configured identity provider:

  ```bash
  TRAFFIC_ACCESS_TOKEN_JWT_E2E=true \
  JWT_E2E_TOKEN_COMMAND=<issuer-command> \
  pytest -p conftest_base test/e2b/test_gateway_jwt_auth.py
  ```

  The issuer command receives `JWT_E2E_SANDBOX_ID` and `JWT_E2E_SANDBOX_UID` and prints the bound JWT.

- The `e2e-e2b-latest` workflow includes Runtime mTLS coverage:
  - Disabled: verifies no certificate init container or mTLS cluster is installed and existing port `49983` plaintext routing still works.
  - Enabled: verifies a successful client-certificate-authenticated request to port `49983` and confirms port `8080` remains plaintext.

Gateway package coverage increased from 85.1% to 90.8%. OIDC coverage is 100%, JWT manager coverage is 96.3%, Filter coverage is 99.4%, registry coverage is 100%, and the Runtime credential package is 97.8%.

- JWT mode and Runtime mTLS are independent, process-wide gateway options; both remain disabled by default.
- JWT verification takes a startup snapshot of CA/JWKS data. Audience policy, token issuance, client token delivery, and key refresh are intentionally out of scope.
- Runtime mTLS credentials are loaded once before Envoy starts and are not rotated in place. Secret changes require a gateway Pod rollout.
- Secret RBAC is intentionally provisioned by `ack-agent-identity`, not by the gateway manifests. The CI workflow creates equivalent least-privilege test RBAC only for the enabled E2E scenario.
- The mTLS server identity contract is fixed SNI `agentruntime.sandbox.agents.kruise.io` with certificate SAN `*.sandbox.agents.kruise.io`.
- `config/sandbox-gateway/identity-ca-rbac.yaml` remains separate because the optional identity namespace may not exist when JWT auth is disabled.

### DIFF
--- a/.github/workflows/e2e-e2b-latest.yaml
+++ b/.github/workflows/e2e-e2b-latest.yaml
@@ -40,7 +40,7 @@ jobs:
             with_gateway: true
             e2b_enable_auth: true
             quota_profile: redis
-          - name: with sandbox-gateway, auth disabled
+          - name: with sandbox-gateway, sandbox-manager auth disabled
             with_gateway: true
             e2b_enable_auth: false
             quota_profile: none
@@ -166,7 +166,7 @@ jobs:
           if [ "${{ matrix.e2b_enable_auth }}" = "false" ]; then
             {
               echo ""
-              echo "# Disable E2B auth (appended by e2e-e2b-latest auth-disabled matrix)"
+              echo "# Disable sandbox-manager E2B API auth"
               echo "# Positional path targets --e2b-enable-auth in config/sandbox-manager/deployment.yaml."
               echo "- op: replace"
               echo "  path: /spec/template/spec/containers/0/args/8"
@@ -229,9 +229,6 @@ jobs:
           args=()
           if [ "${{ matrix.with_gateway }}" = "true" ]; then
             args+=(--with-gateway)
-          fi
-          if [ "${{ matrix.with_gateway }}" = "true" ] && [ "${{ matrix.e2b_enable_auth }}" = "true" ]; then
-            args+=(-m "not gateway_routing and not gateway_uuid_auth and not jwt_auth and not runtime_mtls")
           fi
           if [ "${{ matrix.e2b_enable_auth }}" = "false" ]; then
             args+=(--auth-disabled)
--- a/.github/workflows/e2e-e2b-latest.yaml
+++ b/.github/workflows/e2e-e2b-latest.yaml
@@ -38,42 +38,49 @@ jobs:
             with_gateway: false
             runtime_mtls: false
             jwt_auth: false
+            uuid_auth: false
             e2b_enable_auth: true
             quota_profile: redis-clone
-          - name: with sandbox-gateway
+          - name: with sandbox-gateway, UUID auth
             with_gateway: true
             runtime_mtls: false
             jwt_auth: false
+            uuid_auth: true
             e2b_enable_auth: true
             quota_profile: redis
           - name: with sandbox-gateway, auth disabled
             with_gateway: true
             runtime_mtls: false
             jwt_auth: false
+            uuid_auth: false
             e2b_enable_auth: false
             quota_profile: none
           - name: quota redis fault
             with_gateway: false
             runtime_mtls: false
             jwt_auth: false
+            uuid_auth: false
             e2b_enable_auth: true
             quota_profile: redis-fault
           - name: quota no redis
             with_gateway: false
             runtime_mtls: false
             jwt_auth: false
+            uuid_auth: false
             e2b_enable_auth: true
             quota_profile: no-redis
           - name: with sandbox-gateway, Runtime mTLS
             with_gateway: true
             runtime_mtls: true
             jwt_auth: false
+            uuid_auth: false
             e2b_enable_auth: false
             quota_profile: none
           - name: with sandbox-gateway, JWT auth
             with_gateway: true
             runtime_mtls: false
             jwt_auth: true
+            uuid_auth: false
             e2b_enable_auth: false
             quota_profile: none
     name: sandbox (${{ matrix.name }})
@@ -330,6 +337,14 @@ jobs:
           else
             make deploy-sandbox-gateway
           fi
+          if [ "${{ matrix.uuid_auth }}" = "true" ]; then
+            kubectl get configmap envoy-config -n sandbox-system -o json \
+              | jq '.data["envoy.yaml"] |= sub("enable-auth: false"; "enable-auth: true")' \
+              | kubectl replace -f -
+            kubectl rollout restart deployment/sandbox-gateway -n sandbox-system
+            kubectl rollout status deployment/sandbox-gateway \
+              -n sandbox-system --timeout=5m
+          fi
           if [ "${{ matrix.jwt_auth }}" = "true" ]; then
             kubectl get configmap envoy-config -n sandbox-system -o json \
               | jq '.data["envoy.yaml"] |= sub("enable-auth: false"; "enable-auth: true") | .data["envoy.yaml"] |= sub("enable-jwt-auth: false"; "enable-jwt-auth: true")' \
@@ -359,11 +374,14 @@ jobs:
           if [ "${{ matrix.runtime_mtls }}" = "true" ]; then
             export RUNTIME_MTLS_E2E=true
             args+=(-m runtime_mtls)
-          fi
-          if [ "${{ matrix.jwt_auth }}" = "true" ]; then
+          elif [ "${{ matrix.jwt_auth }}" = "true" ]; then
             export TRAFFIC_ACCESS_TOKEN_JWT_E2E=true
             export JWT_E2E_TOKEN_COMMAND="kubectl exec -n sandbox-system deployment/jwt-e2e-oidc-provider -- /jwt-oidc-provider issue"
             args+=(-m jwt_auth)
+          elif [ "${{ matrix.uuid_auth }}" = "true" ]; then
+            args+=(-m "not gateway_routing and not jwt_auth and not runtime_mtls")
+          elif [ "${{ matrix.with_gateway }}" = "true" ]; then
+            args+=(-m "not gateway_uuid_auth and not jwt_auth and not runtime_mtls")
           fi
           if [ "${{ matrix.e2b_enable_auth }}" = "false" ]; then
             args+=(--auth-disabled)
--- a/.github/workflows/e2e-e2b-latest.yaml
+++ b/.github/workflows/e2e-e2b-latest.yaml
@@ -24,8 +24,6 @@ env:
   GATEWAY_IMG: sandbox-gateway:latest
   CODE_INTERPRETER_IMG: registry-ap-southeast-1.ack.aliyuncs.com/acs/code-interpreter:v1.6
   INPLACE_UPDATE_IMG: registry-ap-southeast-1.ack.aliyuncs.com/acs/code-interpreter:v1.6-new
-  RUNTIME_MTLS_SERVER_IMG: runtime-mtls-server:latest
-  JWT_OIDC_PROVIDER_IMG: jwt-oidc-provider:latest
 
 jobs:
   sandbox:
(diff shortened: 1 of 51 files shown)


config/sandbox-gateway/identity-ca-rbac.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: sandbox-gateway-identity-ca-reader
+  namespace: ack-agent-identity
+rules:
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    resourceNames: ["ack-agent-identity-root-ca.crt"]
```
furykerry: is it possible to just mount the configmap as volume in the pod, instead of access through k8s api?

pkg/identity/oidc/options.go
```diff
@@ -0,0 +1,156 @@
...
+import (
+	"fmt"
+	"net/url"
+	"os"
+	"time"
+)
+
+// Default verifier configuration.
+const (
+	DefaultDiscoveryURL         = "https://id-provider.ack-agent-identity.svc:8445/.well-known/openid-configuration"
```
furykerry: plz remove vendor specific configurations

pkg/sandbox-gateway/jwtauth/manager.go
```diff
@@ -0,0 +1,368 @@
...
+// SetReader provides the single Kubernetes reader used to construct the verifier.
+func (m *Manager) SetReader(reader client.Reader) error {
+	if isNilReader(reader) {
+		return errors.New("JWT authentication reader must not be nil")
+	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if m.reader != nil {
+		if sameReader(m.reader, reader) {
```
furykerry: there is no update case, consider drop the comparison and return error directly

pkg/sandbox-gateway/runtimecredentials/credentials.go
```diff
@@ -0,0 +1,233 @@
...
+// Load fetches, validates, and writes the static runtime mTLS credentials.
+func Load(ctx context.Context, client kubernetes.Interface, opts Options) error {
+	if client == nil {
+		return errors.New("Kubernetes client must not be nil")
+	}
+	if err := opts.validate(); err != nil {
+		return err
+	}
+
+	secret, err := client.CoreV1().Secrets(opts.SecretNamespace).Get(ctx, opts.SecretName, metav1.GetOptions{})
```
furykerry: The OIDC verifier (NewVerifier) uses sigs.k8s.io/controller-runtime/pkg/client.Reader, but the runtime credentials loader uses k8s.io/client-go/kubernetes.Interface (typed clientset). This introduces two different Kubernetes client abstractions in the same component.

consider using client.Reader for consistency if the init binary ever shares a test harness with the gateway.

pkg/sandbox-gateway/runtimecredentials/credentials.go
```diff
@@ -0,0 +1,233 @@
...
+	SecretNamespace string
+	SecretName      string
+	OutputDirectory string
+}
+
+// DefaultOptions returns the fixed runtime mTLS credential locations.
+func DefaultOptions() Options {
+	return Options{
+		SecretNamespace: DefaultSecretNamespace,
+		SecretName:      DefaultSecretName,
```
furykerry: plz use environment to get secret name and namespace , or put the cert in sandbox-system

DahuK (commented):
Consider adding a TODO for JWT token expiration (exp claim) validation at the gateway filter level. When JWT mode lands, the filter should reject tokens whose exp has passed, independent of the controller-side refresh schedule.

pkg/identity/oidc/options.go
chengzhycn: done

pkg/sandbox-gateway/jwtauth/manager.go
chengzhycn: done

pkg/sandbox-gateway/runtimecredentials/credentials.go
chengzhycn: done

pkg/sandbox-gateway/runtimecredentials/credentials.go
chengzhycn: changed to get secret name/namespace from environment.

config/sandbox-gateway/identity-ca-rbac.yaml
chengzhycn: Since the CA ConfigMap may be managed by an identity provider in a different namespace, is it more flexible to access through the Kubernetes API?

chengzhycn:
> Consider adding a TODO for JWT token expiration (exp claim) validation at the gateway filter level. When JWT mode lands, the filter should reject tokens whose exp has passed, independent of the controller-side refresh schedule.

This is already handled by ValidateWithLeeway, which validates the exp claim against the current time with the configured clock skew. We also require exp to be present and have a unit test covering expired tokens, so no additional TODO is needed.

pkg/sandbox-gateway/runtimecredentials/credentials.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	if !now.Before(leaf.NotAfter) {
+		return credentialFiles{}, fmt.Errorf("client certificate expired at %s", leaf.NotAfter.UTC().Format(time.RFC3339))
+	}
+	if !allowsClientAuth(leaf.ExtKeyUsage) {
+		return credentialFiles{}, errors.New("client certificate does not allow ClientAuth")
+	}
+	if err := validateCertificatesPEM(ca); err != nil {
+		return credentialFiles{}, fmt.Errorf("parse CA bundle: %w", err)
+	}
```
furykerry, line 155: Issue: credentialsFromSecret verifies that the client certificate/key parse, the leaf is time-valid, and the CA bundle parses, but it never verifies that the client certificate chains to the supplied CA bundle.

Why: The init container is intended to fail closed before Envoy starts with unusable runtime mTLS credentials. Today, a Secret containing a valid client cert/key plus an unrelated valid ca.crt is accepted, so the pod can become ready while every runtime mTLS request fails later during TLS handshake. This also contradicts the intended “validate credential chain” behavior for the init path.

Fix: Parse the CA bundle into a cert pool and call leaf.Verify with ClientAuth usage and any intermediates from tls.X509KeyPair.

```
roots, err := certPoolFromPEM(ca)
if err != nil {
    return credentialFiles{}, fmt.Errorf("parse CA bundle: %w", err)
}

intermediates := x509.NewCertPool()
for _, raw := range pair.Certificate[1:] {
    cert, err := x509.ParseCertificate(raw)
    if err != nil {
        return credentialFiles{}, fmt.Errorf("parse client intermediate certificate: %w", err)
    }
    intermediates.AddCert(cert)
}

if _, err := leaf.Verify(x509.VerifyOptions{
    Roots:         roots,
    Intermediates: intermediates,
    CurrentTime:   now,
    KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
}); err != nil {
    return credentialFiles{}, fmt.Errorf("verify client certificate against CA bundle: %w", err)
}
```

config/sandbox-gateway/configmap.yaml
```diff
@@ -9,13 +9,15 @@ metadata:
 data:
   envoy.yaml: |
     admin:
+      access_log_path: /dev/stdout
```
furykerry, line 12: Issue: JWT mode requires the gateway service account to get the configured OIDC CA ConfigMap, but the production sandbox-gateway RBAC only grants sandboxes, sandboxes/status, and pods. The E2E provider manifest grants the test ConfigMap, but there is no production JWT-auth overlay/sample that grants the required ConfigMap access.

Recommendation: Add a JWT-auth deployment overlay or sample RBAC manifest alongside the env/config changes, so operators do not enable JWT mode and end up with /readyz permanently failing due RBAC denial.

.github/workflows/e2e-e2b-latest.yaml
```diff
@@ -34,24 +36,53 @@ jobs:
...
             with_gateway: false
+            runtime_mtls: false
+            jwt_auth: false
+            uuid_auth: false
             e2b_enable_auth: true
             quota_profile: redis-clone
-          - name: with sandbox-gateway
+          - name: with sandbox-gateway, UUID auth
             with_gateway: true
+            runtime_mtls: false
```
furykerry: consider move all auth related test into a seperate workflow with only test/e2b/test_gateway_auth.py, 
test/e2b/test_gateway_jwt_auth.py, test/e2b/test_gateway_runtime_mtls.py

pkg/sandbox-gateway/runtimecredentials/credentials.go
chengzhycn: Done in 38ed06c0. The init path now builds root/intermediate pools and verifies the client certificate chain with ClientAuth usage; unrelated-CA and intermediate-chain cases are covered by unit tests.

config/sandbox-gateway/configmap.yaml
chengzhycn: Done in 38ed06c0. Added config/sandbox-gateway/jwt-auth-rbac.yaml as a production sample with resourceNames-scoped ConfigMap access, including guidance for same-namespace and cross-namespace CA locations.

.github/workflows/e2e-e2b-latest.yaml
chengzhycn: Done in 6b1d8db5. UUID auth, JWT auth, and Runtime mTLS now run in a dedicated sandbox-gateway security workflow. Each matrix profile passes only its corresponding test file via --test-file, and the generic latest workflow no longer contains those security profiles.

cmd/sandbox-gateway-cert-init/main.go
```diff
@@ -0,0 +1,56 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openkruise/agents/pkg/sandbox-gateway/runtimecredentials"
+)
+
+func main() {
+	if err := run(context.Background()); err != nil {
+		log.Fatalf("initialize runtime mTLS credentials: %v", err)
+	}
+}
```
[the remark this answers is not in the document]
chengzhycn: Addressed in 56923ff2. The cert-init command now uses klog.Fatalf, matching the logging convention used by the other cmd binaries.

docs/proposals/20260713-traffic-access-token-jwt-verification.md
```diff
@@ -0,0 +1,278 @@
...
+
+The Deployment readiness probe uses `/readyz` on system port `7789`.
+
+Request failures use these statuses:
+
+| Condition | Status |
+|---|---|
+| Enabled verifier not ready | `503 Service Unavailable` |
+| Missing, malformed, expired, or invalid JWT | `401 Unauthorized` |
+| Sandbox ID or UID mismatch | `401 Unauthorized` |
```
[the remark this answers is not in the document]
chengzhycn: Addressed in 56923ff2. The proposal now documents 403 Forbidden for invalid or mismatched traffic JWTs, 503 for a JWT-required route without an available verifier, and the current annotation-derived route-scoped enforcement behavior.

pkg/utils/proxyutils/route.go
```diff
@@ -24,37 +24,41 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+const annotationEnableJWTAuth = "security.agents.kruise.io/enable-jwt-auth"
```
furykerry: consider reuse the annotation def of  AnnotationEnableJwtAuth in identity package

chengzhycn: resolved

pkg/sandbox-gateway/server/server.go
```diff
@@ -69,21 +74,25 @@ type Server struct {
...
-	return &Server{
+func NewServer(client client.Client, port int, readinessChecks ...ReadinessCheck) *Server {
+	server := &Server{
 		port:               normalizePort(port, proxy.SystemPort),
 		client:             client,
 		memberlistBindPort: getMemberlistBindPort(),
 	}
+	if len(readinessChecks) > 0 {
+		server.readinessCheck = readinessChecks[0]
+	}
```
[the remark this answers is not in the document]
chengzhycn: Addressed in 4ee1bc06. NewServer now composes every provided readiness check, skips nil checks, copies the variadic slice, and fails fast on the first error. The existing table-driven health handler test now covers all-success, first/second failure, fail-fast call counts, and nil compatibility.

PR #649, opened 2026-07-15
feat(e2b): resolve sandbox domains dynamically

AiRanthem:
## What changed

- Resolve sandbox response domains from each request's HTTP Host when `--e2b-domain` is empty, while preserving an explicit static override.
- Keep native and customized E2B routing details inside the adapter layer, including BrowserUse address formatting, ports, and IPv6 authorities.
- Remove the hard-coded domain from the default sandbox-manager manifests and support multi-domain certificate generation with optional CA reuse.
- Retain the implemented proposal as the durable design record and remove the completed design spec.

## Why

A single sandbox-manager can front multiple user-facing domains, but a process-wide domain value could not return the hostname used by each client. Request-scoped resolution keeps returned sandbox and BrowserUse addresses aligned with the ingress host.

## Impact

- Fresh default deployments use request-derived domains.
- Deployments that explicitly set `--e2b-domain` retain the previous static behavior.
- Customized BrowserUse responses use the routable `/kruise/<sandboxID>/<port>` address shape.
- `E2BMapper` gains `GetDomain` and `GetSandboxAddress`; the data-plane `proxy.RequestAdapter` contract remains unchanged.

## Validation

- `go test ./pkg/servers/e2b ./pkg/servers/e2b/adapters ./pkg/sandbox-manager ./pkg/proxy -count=1`
- `make vet`
- `make lint` (`0 issues`)
- `go build -o /private/tmp/agents-remove-e2b-domain-sandbox-manager ./cmd/sandbox-manager`
- `git diff --check`
- `bash -n hack/generate-certificates.sh`
- Generated and verified multi-domain and CA-reused certificate chains; confirmed expected SANs and distinct random serials.
- `kubectl kustomize config/sandbox-manager` confirmed no `--e2b-domain` argument and retained the admin-key argument.

E2E tests under `test/` were not run.


pkg/servers/e2b/adapters/native_e2b.go
```diff
@@ -121,5 +121,17 @@ func (a *NativeE2BAdapter) Map(req *ParsedRequest) (
 }
 
 func (a *NativeE2BAdapter) IsSandboxRequest(authority, _ string, _ int) bool {
-	return !strings.HasPrefix(authority, "api.")
+	return !strings.HasPrefix(strings.ToLower(authority), "api.")
```
furykerry, line 124: Issue: Native IsSandboxRequest treats every non-api. host as sandbox traffic.

Recommendation: Classify native sandbox traffic using the same host grammar as Map, not only !strings.HasPrefix(authority, "api."). This avoids routing unrelated hosts such as metrics.example.com, localhost, or gateway-internal authorities through sandbox mapping.

e.g. 
```
func (a *NativeE2BAdapter) IsSandboxRequest(authority, _ string, _ int) bool {
    if strings.HasPrefix(strings.ToLower(authority), "api.") {
        return false
    }
    return hostRegex.MatchString(authority)
}
```

PR #653, opened 2026-07-16
docs: define sandbox-manager state and observation model

AiRanthem:
## Summary

- Replaces the earlier persistent `SandboxRecord` direction with a backend-neutral `State`, `Observation`, and opaque `MutationToken` contract.
- Separates release due, terminal workload, and committed cleanup; uses typed serving freshness instead of a whole-object generation gate.
- Reuses the persisted claim-lock UUID as `DeliveryEpoch` and requires per-delivery fencing for CR writes, runtime access, TrafficPolicy, Route, Checkpoint, and composite Connect.
- Adds a semantically equivalent Chinese proposal alongside the English proposal.

## Confirmed contracts

- `ShutdownTime` is an eventual Controller cleanup instruction. `Release=due` does not change Route, visibility, public state, or quota; only List, Describe, and Kill remain admitted.
- DELETE returns HTTP 204 for absent, unclaimed, committed, and wrong-owner Sandboxes. Wrong-owner DELETE is a no-op with no Route or quota side effect.
- Claim persistence is the visibility commit. The same owner may discover and operate a warm Sandbox before Create or Clone returns.
- Healthy Manager replicas continue serving independently. Replica State, Route, and cache synchronization remain out of scope; the primary Lease remains background coordination only.
- A persisted cleanup trigger is committed only under the Controller recycle-or-delete contract. Controller upgrade and verification are a hard prerequisite before Manager and Gateway enable that interpretation.

## Scope

This PR changes only the English and Chinese design proposals. The proposal remains `provisional`; it does not implement the model.

## Validation

- Audited English and Chinese structure and semantics for parity.
- Checked mapper predicates against current typed Sandbox Phase, Conditions, and `Status.UpdateRevision` contracts.
- Ran whitespace checks on both proposal files.

## Review note

This commit supersedes the PR's earlier `SandboxRecord` proposal. Previous reviews, approvals, and CI results do not validate the current design; fresh review is required.


docs/proposals/20260715-sandbox-manager-state-model.md
```diff
@@ -0,0 +1,194 @@
...
+
+The stable unavailable reasons distinguish confirmed absence, temporary unavailability,
+completion, and termination. Internal state names and diagnostic explanations do not become public
+Sandbox states.
+
+Lookup first establishes existence and ownership, then the requested operation applies state
+policy. Pause and Resume accept stable state or progress in the same direction, while opposite
+progress is a conflict. Connect can join resume but rejects an in-progress pause. Delete accepts
+every owned state and remains idempotent for confirmed absence. Backend and authorization failures
+remain distinct from absence.
```
furykerry: consider make a table for each e2b api and its corresponding behavior (accpeted states, return the states)

docs/proposals/20260715-sandbox-manager-state-model.md
```diff
@@ -0,0 +1,141 @@
...
+persisted in the CR.
+
+| State | Meaning | Route | E2B representation |
+|---|---|---|---|
+| `claimable` | A pool member is available for an attempted claim. | Deny | HTTP 404 `SandboxTemporarilyUnavailable` |
+| `running` | Ready, addressed, and able to serve. | Allow | `running` |
+| `pausing` | Pause was requested but is not complete. | Deny | `paused` |
+| `paused` | Pause is confirmed. | Deny | `paused` |
+| `resuming` | Resume or post-resume initialization is in progress. | Deny | `paused` |
+| `unready` | Exists but cannot currently serve. | Deny | HTTP 404 `SandboxTemporarilyUnavailable` |
```
furykerry: unready sandbox should not return 404 in E2B request

docs/proposals/20260715-sandbox-manager-state-model.md
```diff
@@ -0,0 +1,141 @@
...
+
+### E2B behavior
+
+Public Sandbox bodies continue to expose only `running` or `paused`. Lookup first establishes
+existence and ownership; each API then applies its own state policy.
+
+| E2B API | Accepted internal states | Successful result | Other states |
+|---|---|---|---|
+| Create / Clone | No existing state | Wait for refreshed `running`; return Sandbox state `running`. | Preserve existing operation errors. |
+| List | All observations | Include `running`; include `pausing`, `paused`, and `resuming` as `paused`. | Omit `claimable`, `unready`, `terminating`, and `completed`. |
```
furykerry: unready sandbox should be included in a list operation

docs/proposals/20260715-sandbox-manager-state-model.md
```diff
@@ -0,0 +1,141 @@
...
+| Describe | `running`, `pausing`, `paused`, `resuming` | Return `running` or `paused` using the projection above. | Return the lifecycle HTTP 404 reason from the state table. |
```
furykerry: unready sandbox should be acceptable for a describe operation

docs/proposals/20260715-sandbox-manager-state-model.md
```diff
@@ -0,0 +1,141 @@
...
+| Pause | `running`, `pausing`, `paused` | Start, join, or skip pause and return the existing empty success response. | `resuming` returns HTTP 409; other states use lifecycle unavailability. |
+| Resume | `paused`, `resuming`, `running` | Start, join, or skip resume; return empty HTTP 204 after refreshed `running`. | `pausing` returns HTTP 409; other states use lifecycle unavailability. |
+| Connect | `running`, `paused`, `resuming` | Return Sandbox state `running`; start or join resume when needed. | `pausing` returns HTTP 400; other states use lifecycle unavailability. |
+| Snapshot / Set timeout | `running` | Preserve the existing success response. | Preserve the existing running-only rejection or lifecycle error. |
```
furykerry: is it acceptable to set timeout for a sandbox in unready , paused/pausing/resuming sandbox ?

docs/proposals/20260715-sandbox-manager-state-model.md
AiRanthem: Added an E2B API behavior table covering accepted structured states and results for Create/Clone, List, Describe, Pause, Resume, Connect, Snapshot, Set timeout, Update network, Browser use, and Kill. Each row now states the relevant Visibility, Removal, Transition, and Runtime boundary.

docs/proposals/20260715-sandbox-manager-state-model.md
AiRanthem: Agreed. The proposal now separates Visibility from Runtime. A visible Sandbox remains owner-resolvable when Runtime=unready; runtime policy may return a conflict or unavailable response, but not HTTP 404. Only confirmed catalog absence or Visibility=ended is not found.

docs/proposals/20260715-sandbox-manager-state-model.md
AiRanthem: Agreed. List now includes every Visibility=visible record, regardless of Removal, Transition, or Runtime. An unready Sandbox is projected to the public paused state.

docs/proposals/20260715-sandbox-manager-state-model.md
AiRanthem: Agreed. Describe now accepts every visible record and returns HTTP 200 with running or paused, including when Runtime=unready or the Sandbox CR is temporarily unreadable.

docs/proposals/20260715-sandbox-manager-state-model.md
AiRanthem: Direct Set timeout remains running-only: Visibility=visible + Removal=none + Transition=none + Runtime=ready. It uses UpdatePolicyAlways and can replace the deadline; Pause owns paused retention, while Resume writes a placeholder and finishes with ExtendOnly, so allowing direct timeout during transitions would race with those writes. Old SDKs expect Set timeout to fail on paused before falling back to Resume, so HTTP 204 would falsely signal completion. Other visible states return HTTP 500 without changing the deadline, never a lookup-driven HTTP 404. Timeout writes inside Connect and Resume remain part of those recovery operations, not permission for the direct endpoint.

PR #658, opened 2026-07-17
feat: add lifecycle tracing for controller and manager

Liquorice-Ma:
## Ⅰ. Describe what this PR does

This PR adds OpenTelemetry distributed tracing to sandbox-controller and sandbox-manager, providing end-to-end observability of sandbox lifecycle operations — from API request through controller reconcile — including per-step success/failure status visible in Jaeger.

**New `pkg/tracing/` package — reusable tracing infrastructure:**

- `provider.go` — TracerProvider init; `--tracing-mode` flag (otel/std/file/none) selecting OTLP gRPC, stdout, or file exporter; unrecognized mode values are a startup configuration error (fail fast) instead of silently disabling tracing; custom `RequestIDGenerator` makes TraceID = request ID for unified trace-log correlation; file mode appends a stream of pretty-printed JSON objects with owner-only (0600) permissions since traces carry identifying data
- `sampler.go` — `randomRatioSampler`: root-span sampling uses a server-controlled random draw instead of `TraceIDRatioBased`. Since TraceID equals the caller-supplied request ID, a TraceID-derived decision would let callers pick IDs that are always sampled; `--tracing-sampling-ratio` accepts [0, 1] where an explicit 0 means "sample nothing" and out-of-range values fail startup with a configuration error
- `requestid.go` — request-ID helpers: `NewRequestID` (server-side generation, 32 lowercase hex chars, directly usable as an OTel TraceID), `IsValidRequestID`, and `Enabled` (whether a real exporter is installed)
- `propagator.go` — W3C Trace Context propagation via CRD annotations (manager→controller)
- `reconcile.go` — instrumentation API: four Start helpers plus one shared EndSpan (see "Instrumentation convention" below)
- `processor.go` — FilteringSpanProcessor drops no-op Reconcile spans (iterations with no write operation and no failure)
- `spans.go` — span name and attribute constants

**Instrumentation API (4 Start helpers + 1 End helper)** — callers pick the Start function matching their component and never touch the raw OTel tracer API (tracer scope names — `sandbox-manager` / `sandbox-controller` — are package-private constants):

- `StartManagerRootSpan` — manager HTTP entry point (one call per request in the web framework); SpanKindServer; folds in the request-ID plumbing so TraceID = request ID and `request.id` is recorded as an attribute
- `StartManagerSpan` — sandbox-manager operations; no no-op guard, since the manager originates traces (a parent-less call starts a new root, e.g. background tasks)
- `StartReconcileSpan` — controller Reconcile entry point; extracts trace context from CR annotations and installs the write flag for no-op filtering
- `StartControllerSpan` — controller-side operations; returns a no-op span without a valid parent (never creates orphan roots, zero-cost when tracing is off)
- `EndSpan(ctx, span, err)` — the single closing function for all of the above; sets the span status to `codes.Error` with the error message on failure and `codes.Ok` on success. `EndSpan` is policy-neutral about error classification: benign errors (e.g. an `AlreadyExists` that a call site treats as idempotent success) are normalized to nil by the call site, because the same error can be a genuine failure elsewhere. Failed steps show up as red error-marked spans in Jaeger and can be filtered with `error=true`, so users can locate the failing step instead of judging only by latency.

**Controller instrumentation** — Reconcile span with trace context from CR annotations; child spans for each lifecycle phase (EnsureSandboxRunning/Paused/Resumed/Upgraded/Terminated) and write operations (CreatePod, DeletePod, UpdateStatus, Checkpoint, CSI mounts).

**Manager instrumentation** — per-request root span created via `StartManagerRootSpan` in the web framework (`pkg/servers/web/framework.go`), recording the final ApiError (including the panic-recovery path); API/infra-layer spans via `StartManagerSpan` (ClaimSandbox, CloneSandbox, Pause/Resume/Delete, CreateCheckpoint, CSI mount, WaitForCheckpoint); `InjectTraceContext` writes trace context into CR annotations before creating CRs.

**Observability manifests** — OTel Collector, Jaeger, and Grafana deployment configs under `config/`.

**New CLI flags** — `--tracing-mode` (default `none` for both components), `--tracing-endpoint`, `--tracing-file`, `--tracing-sampling-ratio`, `--tracing-insecure`. Flag definitions, defaults, and help text are owned by `tracing.Config.BindFlags` and shared by both entrypoints, so they cannot drift.

## Ⅱ. Does this pull request fix one issue?

NONE

## Ⅲ. Describe how to verify it

1. Deploy observability stack: `kubectl apply -k config/otel-collector && kubectl apply -k config/jaeger && kubectl apply -k config/grafana`
2. Enable tracing on manager (`--tracing-mode=otel`); controller likewise if controller-side spans are desired (default is `none` for both). Note the bundled demo Collector serves plaintext gRPC, so use `--tracing-insecure=true` with it (as documented in the deployment examples).
3. Trigger sandbox operations and verify traces in Jaeger (`kubectl port-forward svc/jaeger-query -n sandbox-system 16686:16686`).
4. Inject a failure (e.g., create a snapshot on a non-running sandbox) and verify the failing span is marked with `error=true` in Jaeger.
5. Verify the `traceID` field in controller/manager logs matches Jaeger traces.
6. Run unit tests: `go test ./pkg/tracing/... ./pkg/controller/sandbox/... ./pkg/sandbox-manager/... ./pkg/servers/...` — including regression tests for the sampling contract (ratio 0/1/negative/>1, TraceID-independent decisions), no-op filtering ordering (a successful read span followed by a write or a failure never produces a missing parent or a dropped failing iteration), and X-Request-ID handling (case normalization, rejection rules).
7. CI: the `E2E-Tracing` workflow deploys both the controller and the sandbox-manager with `--tracing-mode=std` and runs two suites: `Tracing Stdout` verifies controller-side trace-context extraction from the CR annotation and span export on stdout; `Tracing Full Chain` sends a real HTTP request with a caller-provided `X-Request-ID` to the manager E2B API and verifies the full manager → CR annotation → controller chain, including the requestID == TraceID contract, by asserting on both components' stdout.

## Ⅳ. Special notes for reviews

- **No-op Reconcile filtering**: The controller always creates a Reconcile span (so child write spans have a valid parent), but FilteringSpanProcessor drops it if the iteration performed no write operation and no span in it failed. Retention is decided by per-iteration write/failed flags shared across the whole Reconcile call tree, so it does not depend on the order in which child spans end. Reconciles with CreatePod/DeletePod/PatchPod/Checkpoint/UpdateStatus are retained, and failing iterations are always retained in full (with the failing span marked `error=true`) so errors stay visible instead of being silently dropped.
- **TraceID = RequestID**: Custom `RequestIDGenerator` makes TraceID equal to `X-Request-ID`. Server-generated IDs are emitted directly as 32 lowercase hex chars (`tracing.NewRequestID`); caller-provided IDs are **never replaced** — when tracing is enabled they are only lowercased to the canonical TraceID string form (the ID bytes are unchanged), so the request ID in logs, response headers, and error bodies compares equal to the TraceID string in trace backends; when tracing is disabled the value passes through untouched. When tracing is enabled, an ID that cannot serve as an OTel TraceID (not 32 hex chars, or all-zero — this includes standard hyphenated UUIDs) is rejected with **400 Bad Request** instead of being silently replaced. This is an API-visible behavior change gated on enabling tracing — see `docs/proposals/20260709-request-id-as-traceid.md`.
- **Sampling**: `ParentBased(randomRatioSampler)`. The sampling decision is drawn from crypto/rand on the server side, so external callers cannot raise the effective sampling rate above the configured ratio by choosing request IDs. Known limitation (documented at the validation site in `pkg/servers/web/framework.go`): a caller reusing the same `X-Request-ID` across independent requests makes them share one TraceID; the sampling rate is unaffected.
- **Trace context propagation**: The manager injects W3C trace context into CR annotations before creating CRs; the controller extracts it in `StartReconcileSpan`. CRs without the annotation (e.g., kubectl-created) start a new root trace. On deletion, the pre-delete trace-context Patch is built from a `DeepCopy` (the shared sandbox object is never mutated) and skipped entirely when tracing is disabled.
- **Instrumentation convention**: Pick the Start function by component — `StartManagerSpan` under `pkg/sandbox-manager/`, `pkg/servers/`, `pkg/proxy/`; `StartControllerSpan` under `pkg/controller/`; the two entry-point helpers (`StartManagerRootSpan`, `StartReconcileSpan`) are already wired at the HTTP entry and Reconcile entry respectively. All spans are closed with the shared `EndSpan`. The asymmetry is intentional: the manager originates traces (parent-less calls start new roots), while the controller continues them (parent-less calls are no-ops). This split also fixed a latent manager-side misuse of the no-op-guarded controller helper in claim.go.
- **Default values**: `--tracing-mode` defaults to `none` for both components; tracing is strictly opt-in with zero overhead when disabled (no-op TracerProvider). Any unrecognized mode value fails startup with a configuration error rather than silently disabling tracing.
- **Design docs**: See `docs/specs/2026-07-14-tracing-instrumentation-api-design.md` (standalone instrumentation guide with copy-paste patterns and FAQ), `docs/specs/2026-07-02-sandbox-otel-tracing-design.md`, and `docs/proposals/20260709-request-id-as-traceid.md`.


cmd/agent-sandbox-controller/main.go
```diff
@@ -131,6 +133,16 @@ func main() {
...
 
+	// Tracing flags
+	var tracingMode string
+	var tracingEndpoint string
+	var tracingInsecure bool
+	var tracingSamplingRatio float64
+	flag.StringVar(&tracingMode, "tracing-mode", "otel", "Tracing mode: otel, none")
+	flag.StringVar(&tracingEndpoint, "tracing-endpoint", tracing.DefaultEndpoint, "OTLP gRPC endpoint for tracing export")
+	flag.Float64Var(&tracingSamplingRatio, "tracing-sampling-ratio", 1.0, "Trace sampling ratio (0.0 to 1.0)")
+	flag.BoolVar(&tracingInsecure, "tracing-insecure", true, "Use insecure gRPC for tracing export (dev environment)")
```
zmberg: This should default to false.

cmd/agent-sandbox-controller/main.go
```diff
@@ -131,6 +133,16 @@ func main() {
 		"Buffer size for the sandbox metric GC controller event channel. "+
 			"Sends that would block are counted under sandbox_metrics_gc_dropped_total{reason=\"channel_full\"}.")
 
+	// Tracing flags
+	var tracingMode string
+	var tracingEndpoint string
+	var tracingInsecure bool
+	var tracingSamplingRatio float64
+	flag.StringVar(&tracingMode, "tracing-mode", "otel", "Tracing mode: otel, none")
```
zmberg: "otel means reporting to otel, right? What does none mean?"

cmd/agent-sandbox-controller/main.go
Liquorice-Ma: done

pkg/tracing/reconcile.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	// A write-operation Span (e.g. CreatePod, DeletePod, updateSandboxStatus)
+	// means this Reconcile did real work; mark it so the enclosing Reconcile and
+	// EnsureSandbox* Spans are retained rather than filtered as no-op.
+	if writeSpanNames[spanName] {
+		MarkWrite(ctx)
+	}
+
+	tracer := Tracer("sandbox-controller")
```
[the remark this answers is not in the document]
Liquorice-Ma: done

pkg/servers/web/framework.go
```diff
@@ -71,13 +75,31 @@ func RegisterRoute[T any](mux *http.ServeMux, method, path string, handler Handl
 		}
 		requestID := r.Header.Get("X-Request-ID")
 		if requestID == "" {
-			requestID = uuid.NewString()
+			requestID = strings.ReplaceAll(uuid.NewString(), "-", "")
 		}
```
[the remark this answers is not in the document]
Liquorice-Ma: done

cmd/agent-sandbox-controller/main.go
```diff
@@ -131,6 +133,18 @@ func main() {
 		"Buffer size for the sandbox metric GC controller event channel. "+
 			"Sends that would block are counted under sandbox_metrics_gc_dropped_total{reason=\"channel_full\"}.")
 
+	// Tracing flags
+	var tracingMode string
+	var tracingEndpoint string
+	var tracingInsecure bool
+	var tracingSamplingRatio float64
+	var tracingFilePath string
+	flag.StringVar(&tracingMode, "tracing-mode", "none", "Tracing mode: otel, std, file, none")
```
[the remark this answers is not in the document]
Liquorice-Ma: PR description has been updated

pkg/servers/e2b/core.go
```diff
@@ -94,7 +95,7 @@ func NewController(domain, sysNs, peerSelector, sandboxNamespace, sandboxLabelSe
 
 	sc.server = &http.Server{
 		Addr:              fmt.Sprintf(":%d", port),
-		Handler:           sc.mux,
+		Handler:           tracing.HTTPMiddleware(sc.mux, "sandbox-manager"),
 		ReadHeaderTimeout: 5 * time.Second,
 	}
```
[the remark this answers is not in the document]
Liquorice-Ma: done

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -297,35 +309,55 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (cr
...
 	}
+	defer tracing.EndSpanWithWriteCheck(ctx, reconcileSpan)
 
 	phaseBefore := newStatus.Phase
+
 	switch newStatus.Phase {
 	case agentsv1alpha1.SandboxPending:
+		ctx, span := tracing.StartChildSpan(ctx, tracing.SpanControllerEnsureSandboxRunning)
 		requeueAfter, err = r.getControl(args.Pod).EnsureSandboxRunning(ctx, args)
+		tracing.EndSpanWithWriteCheck(ctx, span)
```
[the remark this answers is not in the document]
Liquorice-Ma: The write flag is set via MarkWrite inside StartChildSpan before EnsureSandboxRunning returns, so EndSpanWithWriteCheck already sees the correct state; later writes like updateSandboxStatus are captured by the outer Reconcile span's defer — no change needed.

zmberg (commented):
## Code Review

Reviewed and found the following:

- **Suggestion**: 3

Please see inline comments for details.

pkg/controller/sandbox/core/pod_control.go
```diff
@@ -125,11 +128,18 @@ func (c *PodControl) CreatePod(ctx context.Context, args CreatePodArgs) (*corev1
 	}
 
 	ScaleExpectation.ExpectScale(GetControllerKey(box), expectations.Create, box.Name)
+	// Trace the pod creation as a child span; the pod name attribute is set
+	// after Create since generateName is only resolved by the API server.
+	ctx, span := tracing.StartChildSpan(ctx, tracing.SpanControllerCreatePod)
+	defer span.End()
```
zmberg: **[Suggestion]** The span is ended directly here, but the creation could succeed or fail. The span should have parameters to set the status (success or failure).

pkg/tracing/propagator.go
```diff
@@ -0,0 +1,120 @@
...
+// annotationCarrier implements propagation.TextMapCarrier over a map[string]string.
+type annotationCarrier struct {
+	annotations map[string]string
+}
+
+// Get returns the value for the given OTel propagator key.
+// The standard W3C "traceparent" key is mapped to TraceContextAnnotationKey
+// so that the annotation key follows Kubernetes naming conventions.
+func (c *annotationCarrier) Get(key string) string {
+	if key == "traceparent" {
```
zmberg: **[Suggestion]** "traceparent" should be defined as a constant instead of using a magic string.

pkg/tracing/provider.go
```diff
@@ -0,0 +1,241 @@
...
+			propagation.TraceContext{},
+			propagation.Baggage{},
+		))
+		return func(context.Context) error { return nil }, nil
+	}
+
+	// Create resource with service attributes.
+	res, err := resource.New(ctx,
+		resource.WithAttributes(
+			semconv.ServiceName(cfg.ServiceName),
```
zmberg, line 191: **[Suggestion]** serviceName and serviceVersion should be obtained via ENV variables. These ENV variables can be configured through the deployment.

pkg/servers/web/framework.go
```diff
@@ -69,15 +73,33 @@ func RegisterRoute[T any](mux *http.ServeMux, method, path string, handler Handl
 				writeJson(ctx, w, code, defaultCode, body, headers, requestID)
 			}
 		}
-		requestID := r.Header.Get("X-Request-ID")
-		if requestID == "" {
-			requestID = uuid.NewString()
+		requestID := strings.ReplaceAll(r.Header.Get("X-Request-ID"), "-", "")
+		if len(requestID) != 32 {
+			requestID = strings.ReplaceAll(uuid.NewString(), "-", "")
 		}
```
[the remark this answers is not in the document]
Liquorice-Ma: done

pkg/controller/sandbox/core/pod_control.go
Liquorice-Ma: done

pkg/tracing/propagator.go
Liquorice-Ma: done

pkg/tracing/provider.go
Liquorice-Ma: done

go.mod
```diff
@@ -25,6 +25,11 @@ require (
 	github.com/spf13/cobra v1.10.0
 	github.com/spf13/pflag v1.0.9
 	github.com/stretchr/testify v1.11.1
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.63.0
+	go.opentelemetry.io/otel v1.40.0
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.38.0
+	go.opentelemetry.io/otel/sdk v1.40.0
+	go.opentelemetry.io/otel/trace v1.40.0
```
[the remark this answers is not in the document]
Liquorice-Ma: done

pkg/sandbox-manager/infra/sandboxcr/sandbox.go
```diff
@@ -476,6 +506,12 @@ func (s *Sandbox) CSIMount(ctx context.Context, driver string, request string) e
...
 func (s *Sandbox) CreateCheckpoint(ctx context.Context, opts infra.CreateCheckpointOptions) (string, error) {
+	ctx, span := tracing.Tracer("sandbox-manager").Start(ctx, tracing.SpanInfraCreateCheckpoint,
+		trace.WithAttributes(
+			attribute.Float64(tracing.AttrCheckpointDuration, opts.WaitSuccessTimeout.Seconds()),
+		),
+	)
+	defer span.End()
 	log := klog.FromContext(ctx)
 	opts = ValidateAndInitCheckpointOptions(opts)
 	log.Info("create checkpoint options", "options", opts)
```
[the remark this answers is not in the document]
Liquorice-Ma: done

zmberg (commented):
A few suggestions on tracing coverage gaps

pkg/controller/sandbox/core/common_control.go
```diff
@@ -132,7 +134,7 @@ func (r *commonControl) EnsureSandboxUpdated(ctx context.Context, args EnsureFun
 	if initCond != nil && initCond.Status != metav1.ConditionTrue {
 		pCond := utils.GetPodCondition(&pod.Status, corev1.PodReady)
 		if pCond == nil || pCond.Status != corev1.ConditionTrue {
-			klog.InfoS("Waiting for pod ready before initialization", "sandbox", klog.KObj(box))
+			klog.FromContext(ctx).Info("Waiting for pod ready before initialization", "sandbox", klog.KObj(box))
 			return nil
 		}
 		if err := r.initializer.Initialize(ctx, box, newStatus); err != nil {
```
zmberg: Consider adding tracing for the `Initialize` call as well, so we can observe sandbox initialization latency in Jaeger.

pkg/controller/sandbox/core/pod_control.go
```diff
@@ -125,11 +128,23 @@ func (c *PodControl) CreatePod(ctx context.Context, args CreatePodArgs) (*corev1
...
 	ScaleExpectation.ExpectScale(GetControllerKey(box), expectations.Create, box.Name)
+	// Trace the pod creation as a child span; the pod name attribute is set
+	// after Create since generateName is only resolved by the API server.
+	ctx, span := tracing.StartControllerSpan(ctx, tracing.SpanControllerCreatePod)
 	err = c.Create(ctx, pod)
+	if pod.Name != "" {
+		span.SetAttributes(attribute.String(tracing.AttrPodName, pod.Name))
+	}
+	// AlreadyExists means the pod is in the desired state, so record it as success.
+	if errors.IsAlreadyExists(err) {
```
zmberg: Consider handling `errors.IsAlreadyExists` inside `tracing.EndSpan` itself (e.g. treating it as a non-error), so each call site doesn't need to special-case it before calling `EndSpan`.

zmberg:
**File:** `pkg/controller/sandbox/core/common_control.go` — `EnsureSandboxPaused`

Consider adding tracing for the `AssumePodCheckpointed` call as well, so checkpoint validation latency is visible in Jaeger.

zmberg:
**File:** `pkg/controller/sandbox/core/common_control.go` — `EnsureSandboxResumed`

Consider adding tracing for the `checkpointControl.Cleanup` call as well, so checkpoint cleanup latency is visible in Jaeger.

pkg/servers/web/framework.go
```diff
@@ -69,15 +71,39 @@ func RegisterRoute[T any](mux *http.ServeMux, method, path string, handler Handl
 				writeJson(ctx, w, code, defaultCode, body, headers, requestID)
 			}
 		}
-		requestID := r.Header.Get("X-Request-ID")
-		if requestID == "" {
-			requestID = uuid.NewString()
+		rawRequestID := r.Header.Get("X-Request-ID")
```
AiRanthem: Please do not rewrite a caller-provided `X-Request-ID`. This currently replaces every non-UUID value and canonicalizes valid UUIDs, changing the API-visible value even when tracing is disabled. Instead, make the server-side request-ID generator emit the required representation directly (for example, using a direct encoding such as base32), and decouple it from the user-visible ID if necessary. If tracing is enabled and a supplied ID is invalid for the tracing scheme, return 400 rather than silently replacing it. Please cover absent, valid, invalid, and all-zero IDs in framework tests, and update the request-ID proposal accordingly.

pkg/servers/web/framework.go
```diff
@@ -69,15 +71,39 @@ func RegisterRoute[T any](mux *http.ServeMux, method, path string, handler Handl
...
 
+		// Create the root span wrapping the entire request lifecycle
+		// (middlewares + handler). StartManagerRootSpan stores the request ID
+		// in ctx so the custom IDGenerator produces TraceID == request ID,
+		// enabling unified trace-log correlation.
+		ctx, rootSpan := tracing.StartManagerRootSpan(ctx, fmt.Sprintf("%s %s", method, path), requestID)
+		// apiErr carries the final middleware/handler error so the deferred
+		// EndSpan can record the request outcome on the root span. The explicit
+		// nil check avoids the typed-nil *ApiError turning into a non-nil error.
+		var apiErr *ApiError
```
AiRanthem: The tracing change only needs the final error to remain in scope for the deferred `EndSpan`. Please keep the existing `err` name and control flow instead of renaming it to `apiErr` and splitting `resp, err := handler(...)` into a separate declaration and assignment. This is unrelated churn and makes the PR harder to review.

pkg/servers/e2b/snapshot.go
```diff
@@ -51,12 +53,22 @@ func (sc *Controller) CreateSnapshot(r *http.Request) (web.ApiResponse[*models.S
...
+		span.SetAttributes(attribute.String(tracing.AttrSnapshotTTL, *request.Extensions.TTL))
+	}
 	checkpointID, err := sbx.CreateCheckpoint(ctx, infra.CreateCheckpointOptions{
 		KeepRunning:        request.Extensions.KeepRunning,
 		TTL:                request.Extensions.TTL,
 		PersistentContents: request.Extensions.PersistentContents,
 		WaitSuccessTimeout: time.Duration(request.Extensions.WaitSuccessSeconds) * time.Second,
 	})
+	// End the span with the checkpoint result so failures are visible in traces.
+	tracing.EndSpan(ctx, span, err)
```
AiRanthem: Could we register `EndSpan` with `defer` immediately after `StartManagerSpan`, using a stable operation-error variable? That makes span closure structurally guaranteed on every return path and keeps this call site consistent with the other manager operations.

pkg/servers/e2b/core.go
```diff
@@ -182,9 +182,13 @@ func (sc *Controller) Run() (context.Context, error) {
 	// Channel to listen for interrupt signal
 	sc.stop = make(chan os.Signal, 1)
 	signal.Notify(sc.stop, syscall.SIGINT, syscall.SIGTERM)
-	if err := sc.manager.Run(ctx); err != nil {
-		klog.Fatalf("Sandbox manager failed to start: %v", err)
-	}
+	// Start sandbox manager in a goroutine — Run() blocks on cache.Run(),
```
AiRanthem: Please restore the synchronous `sc.manager.Run(ctx)` call. `SandboxManager.Run` is not a permanent lifecycle loop: `Cache.Run` starts the controller-runtime manager in a goroutine, waits for cache sync, marks the cache synced, and then returns. The synchronous call is therefore the startup/readiness barrier. Moving it to a goroutine lets the HTTP server accept requests before the caches are ready, while `/health` remains unconditional, and it removes startup failures from `Controller.Run`'s return contract. The comment here is consequently misleading as well.

pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -367,7 +378,20 @@ func runClaimPostProcesses(ctx context.Context, sbx *Sandbox, lockType infra.Loc
 	if opts.CSIMount != nil {
 		log.Info("starting to perform csi mount")
 		var err error
-		metrics.CSIMount, err = runtime.ProcessCSIMounts(ctx, sbx.Sandbox, *opts.CSIMount)
+		// Trace the CSI mount as a child span; volume count and driver list
+		// are attached afterwards, and End() is called explicitly so the span
+		// only covers the mount itself.
+		csiCtx, csiSpan := tracing.StartManagerSpan(ctx, tracing.SpanInfraProcessCSIMounts)
```
AiRanthem: This CSI tracing + `ProcessCSIMounts` block is duplicated almost verbatim in `clone.go` starting at line 203. Please extract a shared helper that owns the child span, mount call, attributes, and `EndSpan`, returning the duration and error so each caller can retain its existing metrics and error wrapping. A full scan of the changed manager code should use reasonable reuse rather than comments that suppress duplicate detection.

pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -367,7 +378,20 @@ func runClaimPostProcesses(ctx context.Context, sbx *Sandbox, lockType infra.Loc
...
+		metrics.CSIMount, err = runtime.ProcessCSIMounts(csiCtx, sbx.Sandbox, *opts.CSIMount)
+		var drivers []string
```
AiRanthem: This is not a defensive copy of the mount list; it is a required projection from `[]MountConfig` to the `[]string` accepted by `attribute.StringSlice` (which then copies the string slice internally). If the driver-list attribute is useful, keep the projection but preallocate it with `make([]string, len(...))` and fill by index, ideally inside the shared CSI helper. Also, the values are driver names, so `csi.drivers` would be more accurate than the current `csi.volumes` attribute name.

pkg/sandbox-manager/api.go
```diff
@@ -313,7 +329,11 @@ func (m *SandboxManager) syncRoute(ctx context.Context, sbx infra.Sandbox, refre
 }
 
 // PauseSandbox pauses a sandbox and syncs route with peers
-func (m *SandboxManager) PauseSandbox(ctx context.Context, sbx infra.Sandbox, opts infra.PauseOptions) error {
+//
+//nolint:dupl // PauseSandbox and ResumeSandbox are intentionally symmetric flows.
```
AiRanthem: The full duplicate scan also finds these two newly added `//nolint:dupl` suppressions. Please do not bypass duplicate detection here. Extract only the genuinely shared Pause/Resume orchestration—span lifecycle, timing/result metrics, and route sync—behind operation-specific callbacks/configuration, and remove the suppressions.

pkg/sandbox-manager/infra/sandboxcr/sandbox.go
```diff
@@ -173,10 +179,25 @@ func (s *Sandbox) refreshFromAPIReader(ctx context.Context) error {
...
+
+	ctx, span := tracing.StartManagerSpan(ctx, tracing.SpanInfraKill,
+		attribute.String(tracing.AttrSandboxName, s.Name),
+	)
+	defer func() { tracing.EndSpan(ctx, span, err) }()
+
+	// Inject trace context before deletion so the controller's terminating
+	// Reconcile can establish a parent-child trace relationship. Patch failure
+	// is non-fatal: trace loss is acceptable, deletion must proceed.
+	patch := client.MergeFrom(s.Sandbox.DeepCopy())
```
AiRanthem: @zmberg, could you please double-check that this pre-delete Patch follows the intended Kill protocol? There are also two concrete issues to address: with `tracing-mode=none`, `InjectTraceContext` adds no trace context but this code still sends an extra Patch for every deletion; and it mutates `s.Sandbox` directly even though that pointer may be shared by cache/singleflight consumers. Please Patch only when trace context was actually injected, build the patched object from a `DeepCopy`, and keep deletion unconditional if trace propagation fails.

cmd/sandbox-manager/main.go
```diff
@@ -138,6 +140,18 @@ func main() {
 	pflag.DurationVar(&quotaAntiDriftInterval, "quota-anti-drift-interval", consts.DefaultQuotaAntiDriftInterval, "Interval for quota anti-drift reconciliation.")
 	pflag.DurationVar(&quotaAntiDriftGrace, "quota-anti-drift-grace", consts.DefaultQuotaAntiDriftGrace, "Grace period before periodic quota anti-drift releases suspected leaked entries.")
 
+	// Tracing flags
```
AiRanthem: Could the tracing package expose a small shared flag-binding function (for example, binding a `tracing.Config` to a `pflag.FlagSet`) and have both entrypoints use it? The manager and controller currently duplicate the same five flags, defaults, and help text, which makes configuration drift likely.

pkg/tracing/reconcile.go
```diff
@@ -0,0 +1,252 @@
...
+// exported traces. A Span that itself ends with a non-nil error is never
+// marked no-op, and its failure also retains the whole iteration (including
+// the Reconcile Span, which is ended afterwards with a nil error), so failing
+// Reconciles stay visible instead of being silently dropped. Spans outside a
+// Reconcile (e.g. sandbox-manager request handling) carry no write flag and
+// are always exported.
+func EndSpan(ctx context.Context, span trace.Span, err error) {
+	// AlreadyExists means the resource is already in the desired state; treat
+	// it as success so idempotent K8s writes don't show up as failures.
+	if apierrors.IsAlreadyExists(err) {
```
AiRanthem: `AlreadyExists` is not universally a successful outcome. The sandbox-manager create classifier explicitly maps it to `ErrorConflict`, so this global normalization can make an Infra span appear OK even though the request failed with a name collision. Please keep `EndSpan` policy-neutral and normalize `AlreadyExists` only at call sites where the operation has already established that it is an idempotent success.

config/sandbox-manager/deployment.yaml
```diff
@@ -42,6 +42,10 @@ spec:
             # - --quota-redis-operation-timeout=50ms
             # - --quota-anti-drift-interval=5m
             # - --quota-anti-drift-grace=10m
+            # OpenTelemetry tracing (uncomment and configure to enable)
+            # - --tracing-mode=otel
+            # - --tracing-endpoint=otel-collector:4317
+            # - --tracing-insecure=false
```
AiRanthem: The bundled Collector receiver listens on plaintext gRPC (`0.0.0.0:4317` with no TLS config), while both deployment examples set `--tracing-insecure=false`, which makes the clients attempt TLS. Uncommenting the documented configuration therefore cannot export traces. Please either set `--tracing-insecure=true` for the bundled demo stack or configure TLS on the Collector and document the certificates.

pkg/tracing/provider.go
```diff
@@ -0,0 +1,250 @@
...
+		if cfg.Insecure {
+			opts = append(opts, otlptracegrpc.WithTLSCredentials(insecure.NewCredentials()))
+		} else {
+			opts = append(opts, otlptracegrpc.WithTLSCredentials(credentials.NewTLS(nil)))
+		}
+		exporter, err = otlptracegrpc.New(ctx, opts...)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create OTLP gRPC exporter: %w", err)
+		}
+	default:
```
AiRanthem: A typo in `--tracing-mode` currently falls through here and silently disables tracing, which is difficult to diagnose operationally. Please handle `TracingModeNone` explicitly and return a configuration error for every unrecognized value.

test/e2e/tracing_test.go
```diff
@@ -0,0 +1,174 @@
...
+	})
+
+	AfterEach(func() {
+		if sandbox != nil {
+			_ = k8sClient.Delete(ctx, sandbox)
+		}
+	})
+
+	It("should export controller spans to stdout with the propagated trace ID", func() {
+		By("Building a W3C traceparent simulating a sandbox-manager root span")
```
AiRanthem: This test simulates the manager by constructing `traceparent` itself and the workflow deploys only the controller, so it does not verify the advertised manager → CR annotation → controller lifecycle chain or `requestID == TraceID`. Please add an end-to-end case that sends a real manager API request and observes both sides, or narrow the test/suite description to controller trace-context extraction and stdout export.

AiRanthem (changes requested):
Thanks for the tracing work. I am requesting changes for the startup-readiness regression, request-ID compatibility, the unconditional pre-delete Patch, the bundled OTLP TLS mismatch, and the global AlreadyExists status policy. The inline comments also include the consolidated manual-review feedback on duplication, span cleanup, and shared flag registration. Focused package tests and both entrypoint builds pass, but the issues below should be addressed before merge.

pkg/controller/sandbox/core/checkpoint.go
```diff
@@ -157,18 +162,31 @@ func (c *CheckpointControl) Cleanup(ctx context.Context, box *agentsv1alpha1.San
...
 		if delErr := c.Delete(ctx, &cpList[i]); delErr != nil && !errors.IsNotFound(delErr) {
 			ScaleExpectation.ObserveScale(GetControllerKey(box), expectations.Delete, cpList[i].Name)
-			klog.ErrorS(delErr, "Failed to delete checkpoint after resume", "sandbox", klog.KObj(box), "checkpoint", cpList[i].Name)
+			klog.FromContext(ctx).Error(delErr, "Failed to delete checkpoint after resume", "sandbox", klog.KObj(box), "checkpoint", cpList[i].Name)
+			err = delErr
 		} else {
-			klog.InfoS("Deleted checkpoint after successful resume", "sandbox", klog.KObj(box), "checkpoint", cpList[i].Name)
+			// Deleting the checkpoint is a real write; mark the Reconcile so
+			// its spans are retained.
+			tracing.MarkWrite(ctx)
```
zmberg: Is this still necessary here?

pkg/controller/sandbox/core/checkpoint.go
```diff
@@ -157,18 +162,31 @@ func (c *CheckpointControl) Cleanup(ctx context.Context, box *agentsv1alpha1.San
...
 		return
 	}
+	// Trace the cleanup so its latency is observable in Jaeger. The span is
+	// not in writeSpanNames: the write is marked explicitly below only when a
+	// checkpoint is actually deleted, so empty cleanups don't retain no-op
+	// iterations. Cleanup stays best-effort; errors are only recorded on the
+	// span so failures are visible in traces.
+	ctx, span := tracing.StartControllerSpan(ctx, tracing.SpanControllerCheckpointCleanup)
+	var err error
+	defer func() { tracing.EndSpan(ctx, span, err) }()
```
zmberg, line 173: defer tracing.EndSpan(ctx, span, err) ?

pkg/controller/sandbox/core/checkpoint.go
```diff
@@ -200,12 +218,19 @@ func (c *CheckpointControl) createCheckpoint(ctx context.Context, box *agentsv1a
 		},
 	}
 	ScaleExpectation.ExpectScale(GetControllerKey(box), expectations.Create, cpName)
-	if err := c.Create(ctx, cp); err != nil {
+	ctx, span := tracing.StartControllerSpan(ctx, tracing.SpanControllerCheckpoint,
+		attribute.String(tracing.AttrCheckpointName, cpName),
+		attribute.String(tracing.AttrSandboxName, box.Name),
+		attribute.String(tracing.AttrSandboxNamespace, box.Namespace),
```
zmberg: attribute.String(tracing.AttrCheckpointName, cpName),
attribute.String(tracing.AttrSandboxName, box.Name),
attribute.String(tracing.AttrSandboxNamespace, box.Namespace),

Are these attribute still needed? The sandbox controller should already have had these added initially.

pkg/controller/sandbox/core/common_control.go
```diff
@@ -147,10 +155,20 @@ func (r *commonControl) EnsureSandboxUpdated(ctx context.Context, args EnsureFun
...
+		done, wrote, err := r.handleInplaceUpdateSandbox(ctx, args)
 		if err != nil {
 			return err
-		} else if !done {
+		}
+		if wrote {
+			// handleInplaceUpdateSandbox performed an actual write (e.g. patched
+			// the Pod or set an InplaceUpdate condition). Mark the Reconcile write
+			// flag so the enclosing Reconcile span is retained.
+			tracing.MarkWrite(ctx)
```
zmberg: Is this still necessary here?

pkg/controller/sandbox/core/common_control.go
```diff
@@ -241,26 +259,39 @@ func (r *commonControl) EnsureSandboxPaused(ctx context.Context, args EnsureFunc
...
+	// Validate images and create pod-info checkpoint before deletion. Traced
+	// so checkpoint validation latency is observable in Jaeger; the span is
+	// read-mostly and the actual checkpoint creation inside is traced (and
+	// write-marked) by SpanControllerCheckpoint. Use a local context so the
+	// following DeletePod span stays a sibling: it must not parent under a
+	// span that has already ended and may be dropped as no-op, which would
+	// produce a missing-parent trace.
+	cpCtx, cpSpan := tracing.StartControllerSpan(ctx, tracing.SpanControllerAssumePodCheckpointed)
+	rejected := r.checkpointControl.AssumePodCheckpointed(cpCtx, pod, box, newStatus, cond)
+	cpSpan.SetAttributes(attribute.Bool(tracing.AttrCheckpointRejected, rejected))
```
zmberg, line 286: remove it?

pkg/controller/sandbox/core/pod_control.go
```diff
@@ -125,11 +128,25 @@ func (c *PodControl) CreatePod(ctx context.Context, args CreatePodArgs) (*corev1
 	}
 
 	ScaleExpectation.ExpectScale(GetControllerKey(box), expectations.Create, box.Name)
+	// Trace the pod creation as a child span; the pod name attribute is set
+	// after Create since generateName is only resolved by the API server.
+	ctx, span := tracing.StartControllerSpan(ctx, tracing.SpanControllerCreatePod)
 	err = c.Create(ctx, pod)
+	if pod.Name != "" {
+		span.SetAttributes(attribute.String(tracing.AttrPodName, pod.Name))
```
zmberg: These kinds of Pod Name fields aren't necessary, including some of the Delete/Create ones from earlier, because the Pod Name is always equal to the Sandbox Name.

pkg/controller/sandbox/core/sandbox_initializer.go
```diff
@@ -128,9 +130,22 @@ func Initialize(ctx context.Context, box *agentsv1alpha1.Sandbox, newStatus *age
...
 
 		// Cleanup ProcessCSIMounts for concurrent mount execution
-		duration, mountErr := utilruntime.ProcessCSIMounts(ctx, sbxForInit, config.CSIMountOptions{
+		// Trace the CSI remount as a child span, recording the volume count
+		// and driver list as attributes for troubleshooting slow mounts.
+		drivers := make([]string, len(mountOptionList))
+		for i, m := range mountOptionList {
+			drivers[i] = m.Driver
+		}
+		csiCtx, csiSpan := tracing.StartControllerSpan(ctx, tracing.SpanControllerProcessCSIMounts,
```
zmberg: Let's remove these for now. I see they've already been set in the outer function call.

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -252,9 +255,24 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (cr
...
+	// span this early does not produce noise.
+	ctx, reconcileSpan := tracing.StartReconcileSpan(ctx, box)
+	reconcileSpan.SetAttributes(attribute.String(tracing.AttrSandboxPhase, string(box.Status.Phase)))
+	if traceID := tracing.TraceIDFromContext(ctx); traceID != "" {
+		ctx = klog.NewContext(ctx, klog.FromContext(ctx).WithValues("traceID", traceID))
+	}
+	// End the Reconcile span with the final Reconcile error via defer: a
+	// failing iteration is marked failed and always retained even when the
+	// error path is not wrapped in its own child span.
+	defer func() { tracing.EndSpan(ctx, reconcileSpan, err) }()
```
zmberg, line 288: defer tracing.EndSpan(ctx, reconcileSpan, err)

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -380,16 +418,22 @@ func (r *SandboxReconciler) updateSandboxStatus(ctx context.Context, newStatus a
 	}
 	oldPhase := box.Status.Phase
 
+	ctx, span := tracing.StartControllerSpan(ctx, tracing.SpanControllerUpdateStatus,
+		attribute.String(tracing.AttrPhaseBefore, string(box.Status.Phase)),
```
zmberg: remove the attribute.

PR #667, opened 2026-07-19
feat(sandboxclaim): propagate designated claim annotations to pod template

BH4AWS:
…plate

SandboxClaim annotations were only copied onto the Sandbox object, while labels were propagated to both the Sandbox and its pod template. Some annotations need to reach the pod template so the resulting Pod carries them, but propagating all of them would leak internal annotations (e.g. e2b/runtime tokens).

Add an allowlist-based propagation for pod-template annotations:

- Add GetPodAnnotations/SetPodAnnotations to the infra.Sandbox interface and sandboxcr.Sandbox, plus a MergePodAnnotations helper mirroring the existing pod-label handling.
- Introduce a pluggable allowlist in the sandboxclaim core package with RegisterPodPropagatableAnnotationKeys/Prefixes registration hooks so internal/enterprise code can extend it via init() without touching the propagation logic. Both exact-key and prefix matching are supported; the community default is empty (no annotation propagated).
- Wire filtered propagation into the claim Modifier so only allowlisted annotations are merged onto the pod template.

Covered by table-driven unit tests for the merge helper and the allowlist predicate/filter.

### DIFF
--- a/pkg/controller/sandboxclaim/core/common_control.go
+++ b/pkg/controller/sandboxclaim/core/common_control.go
@@ -41,6 +41,7 @@ import (
 	"github.com/openkruise/agents/pkg/sandbox-manager/infra"
 	"github.com/openkruise/agents/pkg/sandbox-manager/infra/sandboxcr"
 	"github.com/openkruise/agents/pkg/utils"
+	annotationutils "github.com/openkruise/agents/pkg/utils/annotations"
 	"github.com/openkruise/agents/pkg/utils/csiutils"
 	utilfeature "github.com/openkruise/agents/pkg/utils/feature"
 	"github.com/openkruise/agents/pkg/utils/timeout"
@@ -314,6 +315,12 @@ func (c *commonControl) buildClaimOptions(ctx context.Context, claim *agentsv1al
 				labels[k] = v
 			}
 			sbx.SetPodLabels(labels)
+
+			// propagate user annotations to podtemplate. Keys with internal
+			// reserved prefixes (annotationutils.BlackListPrefix) are filtered
+			// out so propagated annotations cannot interfere with the sandbox
+			// controller itself.
+			infra.MergePodAnnotations(sbx, annotationutils.FilterBlackListed(claim.Spec.Annotations))
 
 			// apply shutdownTime
 			if claim.Spec.ShutdownTime != nil {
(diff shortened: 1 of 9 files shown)


pkg/controller/sandboxclaim/core/annotation_propagation.go
```diff
@@ -0,0 +1,111 @@
...
+var (
+	propagationMu sync.RWMutex
+
+	// podPropagatableAnnotationKeys holds exact-match annotation keys allowed to
+	// propagate from a SandboxClaim onto the pod template.
+	podPropagatableAnnotationKeys = map[string]struct{}{}
+
+	// podPropagatableAnnotationPrefixes holds annotation key prefixes allowed to
+	// propagate onto the pod template.
+	podPropagatableAnnotationPrefixes []string
```
zmberg: Could podPropagatableAnnotationKeys and podPropagatableAnnotationPrefixes be merged into a single parameter? Using a regex approach.

pkg/controller/sandboxclaim/core/annotation_propagation.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// RegisterPodPropagatableAnnotationKeys registers exact annotation keys that
+// should be propagated from a SandboxClaim onto the pod template. It is intended
+// to be called from an init() in internal/enterprise packages. Calling it
+// multiple times is safe; duplicate keys are ignored.
+func RegisterPodPropagatableAnnotationKeys(keys ...string) {
```
furykerry: it is overkill to construct registering logic for propagation allow-list, instead, re-use existing BlackListPrefix struct (we can move the BlackListPrefix to a pkg/utils) to ensure propagated annotation can not interface with the function of sandbox controller itself

BH4AWS: fixed

pkg/controller/sandboxclaim/core/annotation_propagation.go
BH4AWS: According to the review comments, we have changed the implementation approach by adopting a blacklist method. Only the prefix labels listed in the blacklist, which require protection, will not be synchronized; all others will be subject to the synchronization policy.

PR #668, opened 2026-07-20
fix(peers): reconcile memberlist peers after split brain

zhuangzhewei09:
This PR enables `MemberlistPeers` to recover from split-brain scenarios caused by a failed initial join or a healed network partition.

Currently, Kubernetes peer discovery and `memberlist.Join` only run during startup. If the initial join fails, the process continues as a standalone memberlist node and never rediscovers its intended peers. Likewise, disconnected memberlist partitions have no Kubernetes-backed path to rejoin after connectivity recovers.

The change:
- Reuses the existing namespace, label selector, Pod IP, and `AnnotationMemberlistURL` rules for both initial discovery and reconciliation.
- Rediscovers desired peer addresses every 60 seconds with 10% jitter to spread Kubernetes API requests across replicas.
- Calls `memberlist.Join` only for desired peers that are not currently alive.
- Stops and waits for the reconciliation goroutine before memberlist shutdown.

Normal membership detection remains handled by memberlist gossip and push/pull. The periodic Kubernetes lookup is only a low-frequency recovery backstop.

Refs #413

This implements the split-brain scope confirmed by the maintainer in https://github.com/openkruise/agents/issues/413#issuecomment-4978172819. It intentionally does not include the route-sync transport, retry, or fan-out changes originally discussed in the issue.

The following focused validations pass with Go 1.25.0:

1. `go test ./pkg/peers -count=1`
2. `go test -race ./pkg/peers -count=1`
3. `go test ./pkg/peers -run '^TestMemberlistPeers_ReconcilesSplitBrain$' -count=10`

The new tests cover:
- Existing Pod IP and annotation-based memberlist address resolution.
- Ignoring the local node, empty Pod IPs, and selector mismatches.
- Two initially isolated memberlist nodes automatically converging after Kubernetes exposes each peer to the other partition.
- Healthy peer reconciliation remaining a no-op.

- The change is limited to `pkg/peers/memberlist.go` and `pkg/peers/memberlist_test.go`.
- No public interface, command-line option, or sandbox-manager/gateway wiring changes are introduced.
- The commit is signed off.

### DIFF
--- a/pkg/peers/memberlist.go
+++ b/pkg/peers/memberlist.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"sort"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -30,6 +32,7 @@ import (
 	"github.com/openkruise/agents/pkg/utils"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


zhuangzhewei09:
@furykerry @AiRanthem PTAL~

zhuangzhewei09:
@AiRanthem @furykerry Before updating this PR against master, I'd like to clarify its scope after #920.

#920 already covers startup discovery retries. However, it stops Kubernetes discovery after a successful join, and its proposal explicitly excludes continuous seed reconciliation and does not guarantee recovery between isolated partitions.

I tested two independently formed two-node clusters: after exposing cross-partition seeds, current master did not converge within the test window. A local adaptation of #668 passed the same scenario 10 times under `-race`.

Is automatic post-join partition recovery still desired for #413? If so, would you prefer narrowing #668 to that scenario while preserving #920’s lifecycle and seed validation, or using a different recovery mechanism?

I'd like to align on this before pushing the conflict resolution.

PR #672, opened 2026-07-20
refactor(sandbox): extract syncStatusFromPod as struct field with syncReadyCondition flag

zmberg:
## Summary

Refactor  from a package-level function into a struct field on both  and , enabling pluggable behavior for different control structs.

## Key Changes

- **Extract  as a func field** on  and , with  as the default implementation assigned in constructors
- **Add  parameter** to control whether Ready condition sync is performed. The upgrade scenario passes  (uses upgrade-specific conditions), while normal running/updated paths pass 
- **Replace manual pod info sync** in  with , moved after the Pod Ready check for correctness
- ** passes ** to  for consistent behavior injection

## Test Coverage

- Package total: **95.2%**
- : **100%**
- : 87.8% (uncovered lines are pre-existing error paths unrelated to this change)

## Files Changed

-  — struct field, constructor, function signature
-  — struct field, constructor, call site
-  — test construction updates
-  — test construction updates

### DIFF
--- a/pkg/controller/sandbox/core/upgrade_control.go
+++ b/pkg/controller/sandbox/core/upgrade_control.go
@@ -55,7 +55,7 @@ func NewUpgradeControl(
 	podControl *PodControl,
 	lifecycleHookFunc LifecycleHookFunc,
 	initializer SandboxInitializer,
-	syncStatusFromPod func(pod *corev1.Pod, newStatus *agentsv1alpha1.SandboxStatus, syncCondition bool),
+	syncStatusFromPod func(pod *corev1.Pod, newStatus *agentsv1alpha1.SandboxStatus, syncReadyCondition bool),
 ) *UpgradeControl {
 	return &UpgradeControl{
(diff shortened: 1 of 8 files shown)


pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -54,13 +55,15 @@ func NewUpgradeControl(
 	podControl *PodControl,
 	lifecycleHookFunc LifecycleHookFunc,
 	initializer SandboxInitializer,
+	syncStatusFromPod func(pod *corev1.Pod, newStatus *agentsv1alpha1.SandboxStatus, syncCondition bool),
```
furykerry: Issue: The struct field declares the func type with parameter name syncReadyCondition bool, but the NewUpgradeControl constructor parameter uses syncCondition bool:
Recommendation: Align the constructor parameter name to syncReadyCondition for consistency and documentation clarity

PR #677, opened 2026-07-22
fix: harden API key persistence and sandbox owner labels

AiRanthem:
## Summary

- Replace whole-Secret API-key writes with retrying merge patches so concurrent replicas preserve unrelated key entries.
- Keep team UUID deduplication consistent by preparing upserts from the same API-server Secret revision guarded by an optimistic lock.
- Validate sandbox owners as Kubernetes label values and mirror the owner annotation into the claimed Pod template labels.
- Add regression coverage for concurrent Secret updates, APIReader usage, owner-label synchronization, and invalid owners.

## Why

The Secret-backed API-key store used whole-object updates, which could hit resourceVersion conflicts or interfere with concurrent key mutations across sandbox-manager replicas. Sandbox claims also recorded the owner only as an annotation, leaving the Pod template without the corresponding owner label.

## Impact

API-key create and delete operations now avoid clobbering unrelated concurrent entries, while preserving the existing team identity rules. Claimed Pods carry the sandbox owner label, and invalid owner values fail before persistence.

## Validation

- `env GOFLAGS=-mod=readonly GOCACHE=/private/tmp/go-build-cache go test ./pkg/servers/e2b/keys ./pkg/sandbox-manager/infra/sandboxcr -count=1`
- `env GOFLAGS=-mod=readonly GOCACHE=/private/tmp/go-build-cache go test -race ./pkg/servers/e2b/keys -run 'TestSecretKeyStorage_(PatchSecretKey|CreateKey)$' -count=10`
- `gofmt -l pkg/servers/e2b/keys/secret.go pkg/servers/e2b/keys/secret_test.go pkg/sandbox-manager/infra/sandboxcr/claim.go pkg/sandbox-manager/infra/sandboxcr/claim_test.go` (no output)

### DIFF
--- a/pkg/sandbox-manager/infra/sandboxcr/claim.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/claim.go
@@ -31,6 +31,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/api/validate/content"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
@@ -752,6 +753,17 @@ func performLockSandbox(ctx context.Context, sbx *Sandbox, lockType infra.LockTy
 	log := klog.FromContext(ctx)
 	c := cache.GetClient()
 	utils.LockSandbox(sbx.Sandbox, opts.LockString, opts.User)
+	// sync user annotation to pod label. Annotation values are unrestricted but
+	// label values are not, so reject an owner that cannot be a valid label.
+	if errs := content.IsLabelValue(opts.User); len(errs) > 0 {
+		return fmt.Errorf("invalid owner %q for pod label %s: %s", opts.User, v1alpha1.AnnotationOwner, strings.Join(errs, "; "))
+	}
+	podLabels := sbx.GetPodLabels()
+	if podLabels == nil {
+		podLabels = make(map[string]string, 1)
+	}
+	podLabels[v1alpha1.AnnotationOwner] = opts.User
+	sbx.SetPodLabels(podLabels)
 	var updated *v1alpha1.Sandbox
 	var err error
 	if lockType == infra.LockTypeCreate {
--- a/pkg/sandbox-manager/infra/sandboxcr/claim.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/claim.go
@@ -62,6 +62,11 @@ func ValidateAndInitClaimOptions(opts infra.ClaimSandboxOptions) (infra.ClaimSan
 	if opts.User == "" {
 		return infra.ClaimSandboxOptions{}, fmt.Errorf("user is required")
 	}
+	// Owner is also synced to a pod label (AnnotationOwner key). Annotation
+	// values are unrestricted but label values are not, so reject early.
+	if errs := content.IsLabelValue(opts.User); len(errs) > 0 {
+		return infra.ClaimSandboxOptions{}, fmt.Errorf("invalid owner %q for pod label %s: %s", opts.User, v1alpha1.AnnotationOwner, strings.Join(errs, "; "))
+	}
 	if opts.Template == "" {
 		return infra.ClaimSandboxOptions{}, fmt.Errorf("template is required")
 	}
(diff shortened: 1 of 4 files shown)


pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -752,6 +753,17 @@ func performLockSandbox(ctx context.Context, sbx *Sandbox, lockType infra.LockTy
 	log := klog.FromContext(ctx)
 	c := cache.GetClient()
 	utils.LockSandbox(sbx.Sandbox, opts.LockString, opts.User)
+	// sync user annotation to pod label. Annotation values are unrestricted but
+	// label values are not, so reject an owner that cannot be a valid label.
+	if errs := content.IsLabelValue(opts.User); len(errs) > 0 {
```
furykerry: consider move the check to ValidateAndInitClaimOptions

pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -752,6 +753,17 @@ func performLockSandbox(ctx context.Context, sbx *Sandbox, lockType infra.LockTy
...
+		return fmt.Errorf("invalid owner %q for pod label %s: %s", opts.User, v1alpha1.AnnotationOwner, strings.Join(errs, "; "))
+	}
+	podLabels := sbx.GetPodLabels()
+	if podLabels == nil {
+		podLabels = make(map[string]string, 1)
+	}
+	podLabels[v1alpha1.AnnotationOwner] = opts.User
```
furykerry, line 769: plz add a comment about why use an annotation for this label

pkg/servers/e2b/keys/secret.go
```diff
@@ -340,78 +342,73 @@ func (k *secretKeyStorage) LoadByID(_ context.Context, id string) (*models.Creat
...
-	return createdKey, nil
+	return patchedKey, nil
 }
 
-func (k *secretKeyStorage) updateSecret(ctx context.Context, id string, apiKey *models.CreatedTeamAPIKey) error {
-	secret := &corev1.Secret{}
-	if err := k.APIReader.Get(ctx, client.ObjectKey{Namespace: k.Namespace, Name: KeySecretName}, secret); err != nil {
-		return err
-	}
+func (k *secretKeyStorage) patchSecretKey(ctx context.Context, secret *corev1.Secret, id string, apiKey *models.CreatedTeamAPIKey) error {
```
furykerry, line 376: plz comment that when apiKey is nil, an deletion operation is expected

pkg/servers/e2b/keys/secret.go
```diff
@@ -340,78 +342,73 @@ func (k *secretKeyStorage) LoadByID(_ context.Context, id string) (*models.Creat
...
+		secret := &corev1.Secret{}
+		if err := reader.Get(ctx, client.ObjectKey{Namespace: k.Namespace, Name: KeySecretName}, secret); err != nil {
 			return err
 		}
-		createdKey = key
+		apiKey := prepare(secret)
+		if err := k.patchSecretKey(ctx, secret, id, apiKey); err != nil {
+			return err
+		}
+		patchedKey = apiKey
```
furykerry, line 365: for delete key operation, apiKey is nil, plz comment what nil return value means

pkg/sandbox-manager/infra/sandboxcr/claim.go
AiRanthem: Done — moved the owner label-value check into `ValidateAndInitClaimOptions` so invalid owners fail before claim/lock.

pkg/sandbox-manager/infra/sandboxcr/claim.go
AiRanthem: Done — added a comment explaining we reuse `AnnotationOwner` as the pod label key so it stays aligned with the sandbox owner annotation from `LockSandbox`, and so selectors/policies can match the owner.

pkg/servers/e2b/keys/secret.go
AiRanthem: Done — documented on `patchSecretKey` that a nil `apiKey` means delete the entry at `id`.

pkg/servers/e2b/keys/secret.go
AiRanthem: Done — added a comment that on delete `prepare` returns nil, so a nil return with a nil error means the deletion patch succeeded (or was already a no-op).

pkg/servers/e2b/keys/secret.go
```diff
@@ -171,7 +171,9 @@ func (k *secretKeyStorage) Init(ctx context.Context) error {
 			Name:      "admin",
 			Team:      models.AdminTeam(),
 		}
-		if err := k.retryUpdateSecret(ctx, AdminKeyID.String(), adminKey); err != nil && !apierrors.IsConflict(err) {
+		if _, err := k.retryPatchSecretKey(ctx, k.APIReader, AdminKeyID.String(), func(*corev1.Secret) *models.CreatedTeamAPIKey {
```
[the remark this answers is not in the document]
AiRanthem: Confirmed and fixed. Init now checks `secret.Data[AdminKeyID.String()]`, matching the map key used by patch/refresh. Added coverage for id-only presence and legacy `"admin-key"` migration to the id-keyed entry.

PR #685, opened 2026-07-23
feat(runtime): add agent-runtime client with CSI storage mount API

BH4AWS:
Introduce a capability-grouped agent-runtime client (the Runtime interface) in pkg/utils/runtime, exposing Storage().Mount() backed by the runtime POST /v1/storage/mounts endpoint. Request/response models mirror the server contract: {driver, config} in, {success, mountPath, linkPath, message} out.

The shared transport primitive speaks HTTP or HTTPS/mTLS through an injectable client, retries transient failures with exponential backoff by default (5xx and transport errors are retried, a 4xx APIError is permanent), stops on context cancellation, and can re-resolve the sandbox runtime URL via an optional refresh hook. Non-2xx responses surface as a typed APIError carrying the status code.

Mount records the total call duration (including retries) with structured klog logging.

Add httptest table-driven unit tests covering retry-on-5xx, no-retry-on-4xx, context cancellation, refresh-then-success, success=false handling and error-message extraction.


pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -354,26 +363,76 @@ func pauseTimeReached(pauseTime *metav1.Time, now metav1.Time) bool {
...
+	agentsv1alpha1.AnnotationRuntimeTLSPort,
+}
+
+// syncMirroredPodAnnotations aligns the Sandbox annotations with the pod's for
+// every key in mirroredPodAnnotations and persists the change with a meta
+// patch. It must run before any status write in the same reconcile so that a
+// Ready sandbox always carries the annotations of its current pod. A nil pod
+// (paused or not yet created) is skipped: the stale mirror is harmless while
+// unreachable and is re-aligned right after the next pod is rendered.
+func (r *SandboxReconciler) syncMirroredPodAnnotations(ctx context.Context, box *agentsv1alpha1.Sandbox, pod *corev1.Pod) error {
```
furykerry: reverse sync pod annotations e.g. disable-runtime-tl, to sandbox CR is ugly, instead consider the following approach
1. add an flag annotation to control whether to use tls for runtime, and inject the annotation automatically during sandbox creation

pkg/utils/runtime/storage.go
```diff
@@ -0,0 +1,98 @@
...
+// CreateMountRequest is the request body accepted by POST /v1/storage/mounts.
+// It is the transport-neutral representation of a single CSI mount intent,
+// identical to the arguments already carried by the sandbox-storage CLI:
+//
+//   - Driver: the CSI driver name (e.g. "ossplugin.csi.alibabacloud.com").
+//   - Config: base64(proto.Marshal(csi.NodePublishVolumeRequest)), as produced by
+//     csiutils.CSIMountHandler.CSIMountOptionsConfig.
+type CreateMountRequest struct {
+	Driver string `json:"driver"`
+	Config string `json:"config"`
```
furykerry: consider replace config of string type with publishRequest with NodePublishVolumeRequest

pkg/utils/runtime/storage.go
```diff
@@ -0,0 +1,98 @@
...
+// POST /v1/storage/mounts. The runtime returns this shape on both success and
+// failure; Success is authoritative, and MountPath/LinkPath are populated only
+// on success.
+type CreateMountResponse struct {
+	// Success indicates whether the mount operation completed successfully.
+	Success bool `json:"success"`
+	// MountPath is the actual filesystem path where the volume was mounted.
+	MountPath string `json:"mountPath,omitempty"`
+	// LinkPath is the user-facing symlink path pointing to MountPath.
+	LinkPath string `json:"linkPath,omitempty"`
```
furykerry: LinkPath is implementation specific and LinkPath can be obtained from CreateMountRequest, we can exclude it from the response struct.

pkg/utils/runtime/client.go
```diff
@@ -0,0 +1,481 @@
...
+		req.Header.Set("Authorization", "Basic cm9vdDo=") // Basic root:
+
+		// In automatic TLS mode dial the sandbox Pod IP while verifying the server
+		// certificate against the authority hostname (curl --resolve). The pinned
+		// transport is built per attempt because WithRefresh may change the Pod IP
+		// between retries. Otherwise use the shared client (plain HTTP, or a
+		// caller-provided client via WithHTTPClient/WithBaseURL).
+		httpClient := r.client
+		if dialIP := r.dialIPFor(sbx); dialIP != "" {
+			httpClient = &http.Client{Transport: newPinnedTransport(dialIP, r.tlsPort, r.tlsClientConfig.Clone())}
```
furykerry, line 397: Issue: In TLS mode every attempt builds a fresh http.Client{Transport: newPinnedTransport(...)}. The zero-value http.Transport has IdleConnTimeout: 0 (never expires), so after the response body is drained "to enable connection reuse", the kept-alive TLS connection is parked in a transport that is immediately discarded — it is never reused and never closed until the remote end closes it. Every mount/init call in TLS mode leaks one TCP+TLS connection per attempt for the lifetime of the sandbox pod.

Fix: consider close idle conns after the attempt, or disable keep-alives on the one-shot transport. 

In addition, please add a TODO, after we migrate all process/filesystem api to tls transport, we have to use connection reuse properly, otherwise the tls handshake can be quite costly repeated connection

pkg/utils/runtime/client.go
```diff
@@ -0,0 +1,481 @@
...
+	}
+}
+
+// WithBaseURL overrides the runtime endpoint that would otherwise be resolved
+// from the sandbox via GetRuntimeURL. It is required to target the HTTPS server,
+// which the agent-runtime exposes on a separate port from the plain-HTTP one:
+// GetRuntimeURL only yields the HTTP endpoint, so an HTTPS/mTLS client must be
+// paired with the matching https://host:tlsPort base URL here. A trailing slash
+// is tolerated. An empty value is ignored.
+func WithBaseURL(baseURL string) Option {
```
furykerry: WithBaseURL is never used in the code base, is it necessary?

pkg/utils/runtime/filesystem.go
```diff
@@ -0,0 +1,210 @@
...
+		// Drain and close to enable connection reuse.
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}()
+
+	result := WriteFileResult{StatusCode: resp.StatusCode}
+	if resp.StatusCode >= http.StatusBadRequest {
+		// Read up to 1 KiB of the response body to surface the runtime-side error reason
+		// without unbounded memory usage.
+		bodyBytes, _ := io.ReadAll(io.LimitReader(resp.Body, 1<<10))
```
furykerry: A mid-body transport error is silently swallowed; on a 2xx it then surfaces as a misleading "failed to decode" error (or, worse, a silently empty respOut when len(bodyBytes) == 0). Project convention is "never ignore errors."

pkg/controller/sandbox/sandbox_controller.go
BH4AWS: Done, using env ENABLE_AGENTRUNTIME_TLS

pkg/utils/runtime/client.go
BH4AWS: done

pkg/utils/runtime/client.go
BH4AWS: done

BH4AWS: done

pkg/utils/runtime/filesystem.go
BH4AWS: done

pkg/utils/runtime/storage.go
BH4AWS: done

pkg/utils/runtime/storage.go
BH4AWS: done

pkg/utils/runtime/client_test.go
```diff
@@ -0,0 +1,205 @@
...
+func writeMountResponse(t *testing.T, w http.ResponseWriter, status int, resp CreateMountResponse) {
+	t.Helper()
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(status)
+	require.NoError(t, json.NewEncoder(w).Encode(resp))
+}
+
+func TestStorageMount_RetriesOn5xxThenSucceeds(t *testing.T) {
+	var calls int32
+	server, sbx := newMountTestServer(t, func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
BH4AWS: done

pkg/controller/sandbox/core/pod_control.go
```diff
@@ -153,6 +174,63 @@ func (c *PodControl) CreatePod(ctx context.Context, args CreatePodArgs) (*corev1
...
+	if err := c.Patch(ctx, updated, client.MergeFrom(box)); err != nil {
+		klog.ErrorS(err, "failed to stamp runtime TLS annotation on sandbox", "sandbox", klog.KObj(box))
+		c.recorder.Event(box, corev1.EventTypeWarning, "RuntimeTLSStampFailed",
+			fmt.Sprintf("Failed to stamp runtime TLS annotation: %v", err))
+		return fmt.Errorf("failed to stamp runtime TLS annotation: %w", err)
+	}
+	// Sync the in-memory sandbox only after the patch succeeded so a failed
+	// stamp never leaves a dirty annotation on the object shared with the rest
+	// of the reconcile.
+	updated.DeepCopyInto(box)
```
furykerry: no need to deepcopy, if the patch fails, updated will not be modified, so we can simply pass box to the patch

api/v1alpha1/annotations.go
```diff
@@ -21,6 +21,24 @@ package v1alpha1
...
+	// the annotation before it turns Ready. The resume-from-pause path never
+	// stamps: a resumed sandbox was already stamped at first creation, and the
+	// checkpoint delta may override the sidecar configuration. The stamp is
+	// write-once and add-only: turning the switch off never removes the
+	// annotation from existing sandboxes — the cluster-wide fallback is the
+	// client-side TLS material switch or a forward fix. Its absence means the
+	// runtime only speaks plain HTTP. Clients combine it with locally loaded
+	// TLS material to decide whether to reach the runtime over HTTPS with
+	// forced resolution.
+	AnnotationRuntimeTLSPort = InternalPrefix + "runtime-tls-port"
```
furykerry, line 28: in api definition, the comment should not expose implementation detail. plz simplified the comment to 
```
	// AnnotationRuntimeTLSPort advertises that the sandbox's agent-runtime
	// listens for HTTPS on the given port (e.g. "49984").  Its absence means the
	// runtime only speaks plain HTTP
```

pkg/utils/runtime/transport.go
```diff
@@ -0,0 +1,163 @@
...
+// TLSMaterial carries the client-side certificate material used to speak
+// HTTPS/mTLS to the agent-runtime.
+//
+// Only CABundle is required: it verifies the server certificate presented by
+// the runtime. ClientCertPEM/ClientKeyPEM are optional because the runtime
+// server is configured with tls.VerifyClientCertIfGiven — presenting a client
+// certificate upgrades the connection to mutual TLS, but omitting it still
+// yields a valid server-authenticated TLS connection. Provide both the client
+// certificate and key together, or neither.
+type TLSMaterial struct {
```
furykerry: plz rename to TLSBundle

pkg/utils/sidecarutils/sidecar_config_inject.go
```diff
@@ -165,6 +165,29 @@ func setAgentRuntimeContainer(ctx context.Context, podSpec *corev1.PodSpec, conf
 }
 
 func applyInjectionTemplate(pod *corev1.Pod, config SidecarInjectConfig) {
+	applyInjectionMetadata(pod, config)
```
furykerry: plz remove unnecessary changes.

pkg/utils/runtime/client.go
```diff
@@ -0,0 +1,370 @@
...
+
+const (
+	// defaultRuntimeTimeout bounds a single runtime API call when the caller does
+	// not configure an explicit timeout via WithRequestTimeout.
+	defaultRuntimeTimeout = 30 * time.Second
+
+	// accessTokenHeader is the header the agent-runtime BearerAuthMiddleware reads
+	// to authenticate a request (see pkg/agent-runtime/auth). Despite the
+	// middleware name it validates this header value against the configured
+	// access tokens rather than a standard Authorization: Bearer scheme.
```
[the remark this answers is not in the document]
furykerry: @BH4AWS plz fix this prolem

api/v1alpha1/annotations.go
BH4AWS: done

pkg/utils/runtime/transport.go
BH4AWS: done

api/v1alpha1/annotations.go
BH4AWS: done

pkg/utils/runtime/transport.go
BH4AWS: done

pkg/utils/sidecarutils/sidecar_config_inject.go
BH4AWS: done

pkg/utils/runtime/client.go
```diff
@@ -0,0 +1,466 @@
...
+func WithRequestTimeout(d time.Duration) Option {
+	return func(rc *runtimeClient) {
+		if d > 0 {
+			rc.timeout = d
+		}
+	}
+}
+
+// WithRetry overrides the retry schedule. Retry is enabled by default
+// (defaultRetryBackoff); pass a backoff with Steps <= 1 to effectively disable
```
furykerry: With Steps: 0, wait.ExponentialBackoff never invokes the function at all and returns ErrWaitTimeout — the request is silently never sent.
Recommendation: plz correct the comment to "Steps: 1".

pkg/utils/runtime/client.go
```diff
@@ -0,0 +1,466 @@
...
+		if payload != nil {
+			req.Header.Set("Content-Type", "application/json")
+		}
+		if token := utils.GetAccessToken(sbx); token != "" {
+			req.Header.Set(accessTokenHeader, token)
+		}
+		// Optional Basic credential (root, empty password). It plays no part in
+		// access-token authentication; the runtime resolves it to the root user
+		// when present and ignores it when absent.
+		req.Header.Set("Authorization", "Basic cm9vdDo=") // Basic root:
```
furykerry: plz paas the user instead of hard coding root user

pkg/utils/runtime/client_test.go
```diff
@@ -0,0 +1,314 @@
...
+// fastBackoff keeps retry-based tests quick while still allowing several attempts.
+var fastBackoff = wait.Backoff{
+	Duration: time.Millisecond,
+	Factor:   1.0,
+	Steps:    5,
+}
+
+// newMountTestServer starts an httptest server that serves POST /v1/storage/mounts
+// with the provided handler and returns the server plus a sandbox annotated with
+// its URL. The atomic counter records how many times the handler was invoked.
```
furykerry: "The atomic counter records how many times the handler was invoked" — newMountTestServer no longer owns a counter; it lives in the table harness.

BH4AWS: done

pkg/utils/runtime/client.go
BH4AWS: done

pkg/utils/runtime/client.go
BH4AWS: done

PR #686, opened 2026-07-23
feat: short and stable sandbox IDs

AiRanthem:

- Add an optional stable Sandbox ID in the form `<operator-prefix><26-character UID encoding>`. The suffix encodes the complete 128-bit Kubernetes UID and is persisted in the system-owned `agents.kruise.io/sandbox-id` label.
- Keep assignment disabled by default. When enabled, sandbox-manager assigns a missing ID only at the final successful stage of claim or clone; existing labels remain authoritative and immutable, new Sandboxes do not inherit an ID, and public inputs or extension callbacks cannot set the reserved label.
- Preserve legacy `<namespace>--<name>` IDs for unlabeled Sandboxes without background migration or permanent aliases. All readers treat Sandbox IDs as opaque exact-match values, and cache lookup fails closed when duplicate resolved IDs are observed.
- Introduce the protocol-neutral `pkg/sandboxroute` projection and Store used by sandbox-manager, sandbox-gateway, and proxy. Routes carry explicit ObjectKey, UID, and resourceVersion; identity replacement is atomic, stale upserts/deletes are rejected, and deletion watermarks prevent delayed observations from resurrecting removed routes.
- Use Sandbox informer Add/Update/Delete as the authoritative route feed. Manager and gateway share projection and peer-refresh semantics, including access-token derivation, traffic-auth policy, token redaction, and pre-activation handling of legacy ID-only peer messages; no APIReader repair loop is added.
- Preserve client-facing identity across E2B lifecycle operations, Checkpoints, list pagination, gateway/proxy routing, and errors. Namespace/name diagnostics are restored only after successful lookup and ownership authorization, and user metadata cannot spoof the response-only resource context.
- Finalize the implemented design in `docs/proposals/20260711-short-sandbox-id.md`, consolidate affected unit tests, and document durable identity/routing ownership boundaries in the repository AGENTS guides.

NONE

- [ ] Run focused unit tests:
  ```shell
  GOFLAGS=-mod=readonly GOCACHE=/private/tmp/go-build-cache go test \
    ./pkg/sandboxid ./pkg/sandboxroute/... ./pkg/cache ./pkg/proxy \
    ./pkg/sandbox-manager/... ./pkg/sandbox-gateway/... ./pkg/servers/e2b/... \
    ./pkg/controller/sandbox/... ./pkg/controller/sandboxclaim/core \
    ./pkg/controller/sandboxset
  ```
- [ ] With assignment disabled, verify unlabeled Sandboxes keep legacy IDs while existing labeled Sandboxes still use their persisted IDs.
- [ ] With assignment enabled, verify successful claim and clone operations persist a short ID before returning it, preserve an existing ID, and never copy the source ID to a clone.
- [ ] Verify cache, manager, proxy, and gateway resolve only the active ID and fail closed on duplicate labels or malformed route observations.
- [ ] Verify route replacement and deletion under stale informer/peer events: older resource versions are ignored, the prior ID is retired atomically, and delayed updates cannot cross a deletion watermark.
- [ ] Verify E2B metadata/errors disclose `namespace/name` only after lookup and authorization, and Checkpoint/list pagination preserve the resolved source ID.
- [ ] Run `git diff --check`.

- Rollout is staged: deploy label-aware manager and gateway binaries with assignment disabled, drain old replicas and legacy peer traffic, verify informer synchronization, then enable assignment on sandbox-manager.
- Disabling assignment later stops only future assignments; it does not remove existing labels. After any short ID has been persisted, rolling back to label-unaware binaries is unsafe.
- Prefixes are validated at startup and may use at most 37 characters so the complete ID fits a Kubernetes label value. For native E2B dynamic domains with a five-digit port, keep the prefix at 31 characters or fewer to stay within one DNS label.
- Deletion watermarks are retained for the process lifetime, intentionally trading bounded-per-object memory for protection against arbitrarily delayed route events.
- The final test-refactoring and AGENTS commits do not change runtime behavior.


pkg/sandboxid/sandboxid.go
```diff
@@ -0,0 +1,112 @@
...
+	parsed, err := uuid.Parse(string(uid))
+	if err != nil {
+		return "", fmt.Errorf("invalid sandbox UID %q: %w", uid, err)
+	}
+	return strings.ToLower(shortEncoding.EncodeToString(parsed[:])), nil
+}
+
+// ValidatePrefix checks that prefix uses only [a-z0-9-] and, when non-empty,
+// starts with [a-z0-9]. Callers own broader prefix and ID correctness policy.
+func ValidatePrefix(prefix string) error {
```
furykerry, line 110: consider use regular expression to validate the format

pkg/sandboxroute/store.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (s *Store) deleteRecordLocked(
+	key types.NamespacedName,
+	current Route,
+	fenceResourceVersion string,
+) {
+	s.deactivateRouteLocked(key, current.ID)
+	delete(s.recordByObject, key)
+	s.deletionByObject[key] = fenceResourceVersion
```
furykerry: plz add TODO about deletion object cleanup

pkg/sandbox-gateway/server/server.go
```diff
@@ -151,7 +157,10 @@ func (s *Server) Start(ctx context.Context) error {
 
 func (s *Server) newServeMux() *http.ServeMux {
 	mux := http.NewServeMux()
-	mux.HandleFunc(proxy.RefreshAPI, s.handleRefresh)
+	mux.Handle(
+		http.MethodPost+" "+refresh.Path,
+		refresh.NewHandler(s.registry, nil),
```
furykerry, line 162: plz add middleware to check whether refresh operations comes from known peers

pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -382,6 +382,19 @@ func runClaimPostProcesses(ctx context.Context, sbx *Sandbox, lockType infra.Loc
 		log.Info("csi mount completed", "cost", metrics.CSIMount)
 	}
 
+	if opts.PostModifier != nil {
+		start := time.Now()
+		postCtx := logs.Extend(ctx, "stage", "post modifier")
+		_, err := sbx.retryUpdate(postCtx, func(cr *v1alpha1.Sandbox) (bool, error) {
+			return opts.PostModifier(cr)
```
furykerry: why assign and update sandbox id in the end of claim process, in that time, the sandbox is already up and running, the cost of any error is too high. Is it possible to assign id in modifyPickedSandbox?

pkg/sandboxroute/store.go
```diff
@@ -0,0 +1,246 @@
...
+	defer s.mu.RUnlock()
+	return len(s.activeKeyByID)
+}
+
+func (s *Store) installRouteLocked(key types.NamespacedName, route Route) (idTakenOver bool) {
+	if current, exists := s.recordByObject[key]; exists && current.ID != route.ID {
+		s.deactivateRouteLocked(key, current.ID)
+	}
+	if previousKey, exists := s.activeKeyByID[route.ID]; exists && previousKey != key {
+		idTakenOver = true
```
furykerry, line 210: the sequence Upsert(A,ID=X) → Upsert(B,ID=X) (takeover) → Delete(B) leaves recordByObject[A] present with ID=X but no activeKeyByID[X] entry. is it necessary to delete(s.recordByObject, key) here ?

pkg/sandbox-gateway/server/server.go
AiRanthem: Thanks for calling this out. I agree that `/refresh` needs peer authentication before accepting route state. This affects the shared refresh channel in both sandbox-manager and sandbox-gateway, so it is outside the short-Sandbox-ID boundary and should not be addressed with a Gateway-only IP/memberlist middleware here. The bidirectional authentication work is already tracked in https://github.com/openkruise/agents/issues/427, and I have added myself as a co-assignee. I’ll keep #686 focused on the ID and routing contract.

pkg/sandboxroute/logging.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	if result.Result == EventResultApplied && result.Reason == ReasonIDTakeover {
+		logger.Error(
+			errors.New(string(result.Reason)),
+			"route mutation ID takeover",
+			"reason", result.Reason,
+		)
+		return
+	}
+	if result.Result == EventResultApplied && operation == "upsert" {
```
furykerry: Issue: The result.Result == EventResultApplied && operation == "upsert" branch logs exactly the same message/KVs as the fall-through below it; the conditional is pure noise and invites future divergence.

Recommendation: Delete the upsert-specific branch, or make the messages actually differ (e.g. include "operation" only where meaningful).

pkg/sandbox-manager/worker_id_allocator.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/google/uuid"
+	coordinationv1 "k8s.io/api/coordination/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	netutils "k8s.io/apimachinery/pkg/util/net"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	workerIDBits        = 20
```
furykerry: Issue: The ID format (Sonyflake bit layout) and the Lease-counter domain size must always agree. Changing one constant silently corrupts the worker-ID space.

Recommendation: Export WorkerIDBits (and workerIDLimit) from pkg/sandboxid and reference them from the allocator.

pkg/sandboxroute/route.go
```diff
@@ -0,0 +1,73 @@
...
+	if r.ID == "" {
+		return fmt.Errorf("route ID must not be empty")
+	}
+	if r.UID == "" {
+		return fmt.Errorf("route UID must not be empty")
+	}
+	if r.ResourceVersion == "" {
+		return fmt.Errorf("route resource version must not be empty")
+	}
+	if _, err := resourceversion.CompareResourceVersion(r.ResourceVersion, r.ResourceVersion); err != nil {
```
furykerry: [route.go:69](file:///Users/shouchen/.qoder/worktree/agents/BcPr7z/pkg/sandboxroute/route.go#L69) and [store.go:129](file:///Users/shouchen/.qoder/worktree/agents/BcPr7z/pkg/sandboxroute/store.go#L129) validate an RV by comparing it with itself (CompareResourceVersion(rv, rv)). A named helper makes intent obvious:

```
func isValidResourceVersion(rv string) bool {
    _, err := resourceversion.CompareResourceVersion(rv, rv)
    return err == nil
}
```

pkg/servers/e2b/pause_resume.go
```diff
@@ -124,7 +124,7 @@ func reservePausedForAnnotations(value *string) map[string]string {
 
 // buildPauseOptions builds pause options.
 // Retention is resolved with priority: request header > sandbox annotation > default.
-func buildPauseOptions(ctx context.Context, sbx infra.Sandbox, now time.Time, headerRetention *time.Duration, headerReservePausedFor *string) infra.PauseOptions {
+func (sc *Controller) buildPauseOptions(ctx context.Context, sbx infra.Sandbox, now time.Time, headerRetention *time.Duration, headerReservePausedFor *string) infra.PauseOptions {
```
furykerry: why add receiver to the function, sc is never used in the func

pkg/sandbox-manager/sandbox_identity.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"errors"
+
+	"github.com/openkruise/agents/pkg/sandbox-manager/infra"
+	"github.com/openkruise/agents/pkg/sandboxid"
+)
+
+// applySandboxIDToModifier replaces any prior delivery ID after the caller modifier.
+func applySandboxIDToModifier(
```
furykerry: the action is not to add sandbox id to  modifier but return a new modifier WITH sandbox id generation, consider rename the func as withSandboxIDAssignment

pkg/sandboxroute/projection.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/openkruise/agents/pkg/sandboxid"
+	"github.com/openkruise/agents/pkg/utils"
+)
+
+// ProjectSandbox constructs a Route from a Sandbox CR. ID, access-token, and
+// traffic-auth derivation are centralized here so every component projects
+// routes with identical compatibility policy. A sandbox without a Pod IP is
+// always treated as creating, matching the existing manager and gateway
+// behavior.
+func ProjectSandbox(sandbox *agentsv1alpha1.Sandbox) (Route, error) {
```
furykerry: it is hard to see that the func return a route, consider rename as RouteFromSandbox, and move the func to route.go

pkg/sandbox-manager/infra/interface.go
```diff
@@ -48,6 +48,25 @@ type QuotaSandboxSourceProvider interface {
 	GetQuotaSandboxSource() QuotaSandboxSource
 }
 
+// RouteSandboxEvent carries one Sandbox observation or deletion. Normal
+// deletions include their Kubernetes resource version; an empty deletion
+// resource version is reserved for DeletedFinalStateUnknown.
+type RouteSandboxEvent struct {
```
furykerry: infra.RouteSandboxEvent、infra.RouteSandboxSource、GetRouteSandboxSource()、handleRouteSandboxEvent、routeSandboxSource, all these func/strct use RouteSandbox terminology which seems a special sandbox literally. Consider rename as SandboxRoute which is more clear and consistent with the package /pkg/sandboxroute

pkg/servers/e2b/resource_context.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"slices"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openkruise/agents/pkg/servers/web"
+)
+
+func withSandboxResource(apiErr *web.ApiError, sandbox metav1.Object) *web.ApiError {
```
furykerry: consider rename as withSandboxResourceContext

pkg/sandboxroute/store_test.go
```diff
@@ -0,0 +1,433 @@
...
+	for worker := 0; worker < workers; worker++ {
+		worker := worker
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for rv := 1; rv <= 100; rv++ {
+				id := fmt.Sprintf("id-%d", worker)
+				store.Upsert(fullRoute(id, "ns", id, types.UID(id), fmt.Sprint(rv)))
+				store.Get(id)
+				store.Len()
```
furykerry, line 444: consider add Delete operation to the concurrent test

pkg/sandboxroute/store.go
```diff
@@ -0,0 +1,248 @@
...
+	}
```
furykerry, line 211: for the above review comments, consider add code comment to explain why it is not necessary to delete  recordByObject, e.g. 

```
    // Intentionally keep recordByObject[previousKey]: it remains the RV
    // watermark for that key so delayed stale observations of it are still
    // rejected. The taken-over ID is simply no longer resolvable via
    // activeKeyByID until a newer event for either key converges.
```

in addition, consider add an unit test for such case: 
1. Upsert(A,ID=X,RV=1) 
2. Upsert(B,ID=X,RV=1) 
3. Delete(B,RV=2)

Get(X) should return false，and a stale Upsert(A, old rv) return  Ignored/stale_resource_version（

api/v1alpha1/annotations.go
```diff
@@ -16,9 +16,30 @@ limitations under the License.
...
-
 const (
+	InternalPrefix = "agents.kruise.io/"
+
+	AnnotationLock               = InternalPrefix + "lock"
+	AnnotationOwner              = InternalPrefix + "owner"
+	AnnotationClaimTime          = InternalPrefix + "claim-timestamp"
+	AnnotationRestoreFrom        = InternalPrefix + "restore-from"
+	AnnotationInitRuntimeRequest = InternalPrefix + "init-runtime-request"
+	AnnotationSandboxID          = InternalPrefix + "sandbox-id"
```
furykerry: consider add a comment explain the difference with LabelSandboxID

pkg/servers/e2b/pause_resume.go
AiRanthem: Addressed in the next branch update by removing the unused receiver and calling `buildPauseOptions` as a package-level private function.

pkg/sandbox-manager/sandbox_identity.go
AiRanthem: Addressed in the next branch update by renaming the decorator to `withSandboxIDAssignment` and updating its callers and tests. Its ordering and enabled/disabled behavior are unchanged.

pkg/sandboxroute/store_test.go
AiRanthem: Addressed in the next branch update. The concurrent Store test now exercises `Upsert`, `Get`, `Len`, and `Delete` for independent per-worker keys with increasing resource versions, and asserts the final active route count is zero. The focused race test passes.

pkg/sandboxroute/store.go
AiRanthem: I added the nearby comment explaining the Sandbox-ID uniqueness precondition and why the previous record remains as its RV watermark. I intentionally did not add the proposed collision-sequence test: duplicate Sandbox IDs are outside supported producer inputs, and we should not turn this defensive path into a supported collision-recovery contract. Existing collision coverage continues to verify fail-closed behavior.

pkg/sandboxroute/route.go
AiRanthem: Addressed in the next branch update with a private `validateResourceVersion` helper reused by both format-validation sites. It returns the original error so route validation keeps its detailed error context; actual new/old RV comparisons are unchanged.

pkg/sandbox-manager/worker_id_allocator.go
AiRanthem: Addressed in the next branch update. `pkg/sandboxid` now exports `WorkerIDLimit`, derived from its private Sonyflake worker-bit layout, and the Lease allocator uses that shared limit for generation mapping. This matches the current wraparound semantics without exposing layout details. I also added a straightforward comment explaining why replicas use a per-prefix Lease, how API-server CAS and rereads make allocation unambiguous, and why this Lease is a persistent allocation ledger rather than leader election.

pkg/servers/e2b/resource_context.go
AiRanthem: Addressed in the next branch update by renaming the helper and its callers and tests to `withSandboxResourceContext`; error contents and disclosure boundaries are unchanged.

pkg/sandboxroute/store.go
AiRanthem: Sandbox IDs are guaranteed to be unique across ObjectKeys by the supported producer contract, so this takeover branch only defensively handles an invalid duplicate observation. We must keep `recordByObject[A]` because it is the resource-version watermark for A. Deleting it would allow a delayed A event to be accepted after B is deleted and resurrect the old route. I added a nearby comment documenting this; the Store logic remains unchanged.

api/v1alpha1/annotations.go
AiRanthem: Addressed in the next branch update with a comment clarifying that `AnnotationSandboxID` records the referenced ID on associated resources such as Checkpoints and TrafficPolicies, while `LabelSandboxID` is the authoritative identity on the Sandbox itself.

pkg/sandboxroute/projection.go
AiRanthem: Addressed in the next branch update: the function is now `RouteFromSandbox`, its implementation lives in `route.go`, and the single-function `projection.go` file is removed.

pkg/sandbox-manager/infra/interface.go
AiRanthem: Addressed in the next branch update by consistently renaming the event, handler, source, getter, Manager handler, concrete source, tests, and related internal messages from `RouteSandbox*` to `SandboxRoute*`. No compatibility alias or behavior change is introduced.

pkg/sandboxroute/logging.go
AiRanthem: Addressed in the next branch update by deleting the redundant upsert-specific branch. The log level, message, and key-values remain unchanged.

PR #692, opened 2026-07-25
fix: e2b describe dead sandboxes and map not-ready running state

### LINKED ISSUE #693: bug: get_info returns 404 during sandbox wake-up (Running but not Ready)
### Background

During the sandbox wake-up (resume) process, calling `get_info()` (E2B SDK `Sandbox.get_info()`, which maps to `GET /sandboxes/{sandboxID}`) returns HTTP 404 instead of the sandbox details. This occurs because the sandbox is in the Running phase but the Ready condition has not yet been satisfied, causing `GetSandboxState` to report the state as `("dead", "RunningResourceClaimedButNotReady")`. The `DescribeSandbox` handler uses `liveSandboxStates` (running, paused), which does not include `"dead"`, so the sandbox is treated as not found.

### Issue Type

bug

### Relevant Code

- `pkg/servers/e2b/services.go` — `DescribeSandbox` calls `getSandboxOfUser(ctx, id, liveSandboxStates)`, which rejects sandboxes whose state is not `running` or `paused`.
- `pkg/servers/e2b/sandbox.go` — `liveSandboxStates = []string{SandboxStateRunning, SandboxStatePaused}` (excludes `dead`).
- `pkg/utils/utils.go` — `GetSandboxState` returns `("dead", "RunningResourceClaimedButNotReady")` when a claimed sandbox is in Running phase but the Ready condition is False:

```go
if sbx.Status.Phase == agentsv1alpha1.SandboxRunning {
    if sandboxReady {
        return agentsv1alpha1.SandboxStateRunning, "RunningResourceClaimedAndReady"
    } else {
        return agentsv1alpha1.SandboxStateDead, "RunningResourceClaimedButNotReady"
    }
}
```

- `pkg/sandbox-manager/api.go` — `GetSandbox` rejects sandboxes whose state is not in the expected states list, returning `BadRequest: sandbox ... is not healthy (state dead, reason RunningResourceClaimedButNotReady)`.

### Error Log

```
{"level":"error","ts":"2026-07-25T12:55:31Z","caller":"e2b/services.go:47","msg":"failed to get sandbox","contextID":"88cd9ec2-0709-42f9-a600-1a7b860d9205","requestID":"1b37c65e-436a-4df3-a563-72692cf8496f","api":"GET /sandboxes/{sandboxID}","user":"admin","id":"default--autoclaw-xxxx","error":"{\"code\":404,\"headers\":null,\"message\":\"Cannot get sandbox default--autoclaw-xxxx: BadRequest: sandbox default--autoclaw-xxxx is not healthy (state dead, reason RunningResourceClaimedButNotReady)\",\"request_id\":\"\"}"}
```

### Notes

- The issue occurs during the resume/wake-up window: after a paused sandbox is resumed, the sandbox transitions to Running phase, but the Ready condition may take time to be satisfied. During this window, `get_info()` returns 404.
- The standard E2B Python SDK's `SandboxState` enum does not include `"dead"`. Returning `"dead"` would cause `ValueError: 'dead' is not a valid SandboxState` on the client side, so simply including `"dead"` in the expected states may not be sufficient without additional handling.
- Other E2B API endpoints that use `liveSandboxStates` may also be affected during the same window: `ConnectSandbox` (`pause_resume.go`), `SetTimeout` (`timeout.go`), `SandboxHandshake` (`services.go`), `CreateSnapshot` (`snapshot.go`), and the network endpoint (`network.go`).

furykerry:

This PR changes the E2B `DescribeSandbox` API to inspect claimed sandboxes in `dead` state, return transient Running/NotReady sandboxes as `running`, and treat terminal-dead or creating sandboxes as not found.

1. **`DescribeSandbox` uses `claimedSandboxStates`** (running, paused, dead) instead of `liveSandboxStates` (running, paused), then filters the result through `isSandboxViewable` before responding.
2. **`convertToE2BSandbox` maps `RunningResourceClaimedButNotReady` to `running`** — when a sandbox is in the Running phase but the Ready condition is not yet satisfied, `GetSandboxState` reports it as `("dead", "RunningResourceClaimedButNotReady")`. The API now surfaces this as `"running"` because the underlying phase is Running and the sandbox is still live, avoiding the E2B SDK `ValueError` on the unparseable `"dead"` state.

fixes #693

1. Run unit tests: `go test ./pkg/servers/e2b/ -run 'TestConvertToE2BSandbox|TestDescribeSandbox' -count=1 -v`
2. Verify that a sandbox in Running phase with Ready=False returns state `"running"` (not `"dead"` or 404)
3. Verify that a terminal-dead sandbox (Terminating phase) returns 404 as not found
4. Verify that reserved-failed sandboxes still return 404

- The standard E2B Python SDK's `SandboxState` enum only has `running` and `paused` — it does not include `dead`. Returning `"dead"` would cause `ValueError: 'dead' is not a valid SandboxState` in `SandboxDetail.from_dict()`.
- The `RunningResourceClaimedButNotReady` case is the most common "dead" state for claimed sandboxes — it occurs whenever a sandbox is in the Running phase but hasn't reached Ready yet. Mapping it to `"running"` is semantically correct because the underlying phase IS Running.
- Terminal-dead sandboxes (Terminating, Succeeded, Failed, Deleted) and creating sandboxes are treated as not found from the E2B API perspective.
- `convertToE2BSandbox` is shared by all E2B API endpoints (describe, connect, create, list), so the state mapping applies uniformly.
- this solve part of the issues picture in proposal https://github.com/openkruise/agents/pull/653, which potentially can solve other parts of sandbox states handling such as routing

### DIFF
--- a/pkg/servers/e2b/sandbox.go
+++ b/pkg/servers/e2b/sandbox.go
@@ -76,6 +76,22 @@ func (sc *Controller) getSandboxOfUser(ctx context.Context, sandboxID string, ex
 	return sbx, nil
 }
 
+func isSandboxViewable(sbx infra.Sandbox) bool {
+	state, reason := sbx.GetState()
+	if state == agentsv1alpha1.SandboxStateCreating {
+		return false
+	}
+	if state != agentsv1alpha1.SandboxStateDead {
+		return true
+	}
+	switch reason {
+	case "ResourceSucceeded", "ResourceFailed", "ResourceTerminating", "ResourceDeleted":
+		return false
+	default:
+		return true
+	}
+}
+
 func (sc *Controller) getNamespaceOfUser(user *models.CreatedTeamAPIKey) string {
 	team := keys.TeamForKey(user)
 	// Keys in the admin team can access resources in cluster scope
@@ -98,7 +114,17 @@ func (sc *Controller) convertToE2BSandbox(sbx infra.Sandbox, accessToken, domain
 		TrafficAccessToken:           sbx.GetTrafficAccessToken(),
 		TrafficAccessTokenExpiration: sbx.GetTrafficAccessTokenExpiration(),
 	}
-	sandbox.State, _ = sbx.GetState()
+	// A sandbox in the Running phase that is claimed but not yet ready is
+	// reported as "dead" by GetSandboxState because the Ready condition is
+	// unsatisfied. The underlying phase is Running, so surface it as
+	// "running" to E2B API clients — this avoids returning the unparsable
+	// "dead" state to E2B SDK clients while reflecting that the sandbox is
+	// still live.
+	state, reason := sbx.GetState()
+	if state == agentsv1alpha1.SandboxStateDead && reason == "RunningResourceClaimedButNotReady" {
+		state = agentsv1alpha1.SandboxStateRunning
+	}
+	sandbox.State = state
 	annotations := sbx.GetAnnotations()
 	labels := sbx.GetLabels()
 
--- a/pkg/servers/e2b/sandbox_test.go
+++ b/pkg/servers/e2b/sandbox_test.go
@@ -20,6 +20,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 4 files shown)


AiRanthem:
Suggestion: Could we simplify these tests by keeping only the new not-ready-to-running mapping case, testing `isSandboxViewable` with a small `GetState` stub, reusing `AnonymousUser`, and using the CR’s existence as the deletion oracle? This would remove redundant fixtures, spies, and implementation-detail comments while preserving the relevant coverage.

PR #697, opened 2026-07-27
docs(proposal): add agent identity proposals for sandbox ingress authn and outb…

DahuK:
## What this PR does / why we need it

This PR adds two design proposals under `docs/proposals/` that together cover
**bidirectional identity** for sandbox agents — inbound authentication into a
sandbox, and outbound delegated credential acquisition from a sandbox. Both are
documentation-only, gated behind feature gates that are **off by default**, and
fully backward compatible.

### 1. `20260725-sandbox-ingress-authn.md` — Sandbox Ingress (envd) Token Authentication and aud Authorization

Introduces JWT-based authentication and **audience (aud) authorization** for the
**ingress direction** (client → sandbox-gateway → envd/agent-runtime). Only a
token whose `aud` explicitly contains the target sandbox's SPIFFE ID — and that
passes signature, expiry, and admin-policy checks — may reach that sandbox's
envd interfaces.

- **Issuance plane** (`sandbox-manager` + `pkg/identity`): exchange the existing
  Team API Key for a short-lived JWT carrying `aud`.
- **Authorization control plane**: a new namespaced CRD `SandboxIngressPolicy`
  (symmetric to the egress `SecurityProfile`).
- **Verification data plane** (`sandbox-gateway`): upgrade the current static
  `x-access-token` comparison into a tiered verification chain.
- Guarded by the `SandboxIngressJWTAuth` feature gate (gateway + manager, Alpha,
  disabled by default); `TokenClass=Opaque` preserves the legacy path.

### 2. `20260727-agent-dynamic-identity-spiffe-token-exchange.md` — Agent Dynamic Identity: SPIFFE Workload Identity and OAuth 2.0 Token Exchange

Introduces SPIFFE workload identity + OAuth 2.0 Token Exchange (RFC 8693) for the
**outbound direction**, so an agent can call third-party APIs on behalf of a user
without ever holding long-lived secrets.

- **Trusted Principal Derivation**: the user identity is derived from a verified
  inbound OIDC credential, never from a caller-supplied field.
- **Two-Stage Token Exchange**: keeps the user credential inside the trusted
  Identity Service (never in the untrusted sandbox) while producing a final token
  carrying both the user subject and the sandbox as authorized party; the
  intermediate audience is derived from the verified sandbox SVID.
- **Delegation Store (encrypted vault)**: refresh tokens are stored as
  envelope-encrypted ciphertext; the decryption key is held only by the Identity
  Service; the agent obtains only short-lived access tokens via an authenticated API.
- Supports 2LO / 3LO / OBO outbound flows with sequence diagrams.
- Guarded by the `SecurityIdentityProvider` feature gate (Alpha, disabled by default).

## Which issue(s) this PR fixes

Relates to roadmap → Runtime → *more secure access token and auditing*.

## Special notes for your reviewer

- **Docs only** — no code, no CRD wiring, no behavior change in this PR.
- Both features are **off by default** and backward compatible.
- The two proposals are complementary (ingress vs egress) and cross-reference
  each other via `see-also` front matter.
- Reviewers for the `reviewers:` front matter field are still `@TBD` — please
  suggest owners.

## Does this PR introduce a user-facing change?

NONE

### DIFF
--- a/docs/proposals/20260725-sandbox-ingress-authn.md
+++ b/docs/proposals/20260725-sandbox-ingress-authn.md
@@ -1,0 +1,1042 @@
+---
+title: Sandbox Ingress (envd) Token Authentication and aud Authorization
+authors:
+  - "@DahuK"
+reviewers:
+  - "@TBD"
+creation-date: 2026-07-25
+last-updated: 2026-07-25
+status: provisional
+see-also:
+  - "/docs/proposals/20260727-agent-dynamic-identity-spiffe-token-exchange.md"
+---
+
+# Sandbox Ingress (envd) Token Authentication and aud Authorization
+
+> Feature gate: `SandboxIngressJWTAuth` (gateway + manager, Alpha, disabled by default).
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Glossary](#glossary)
+- [Baseline (Current State)](#baseline-current-state)
+- [Proposal](#proposal)
+    - [Architecture: three planes](#architecture-three-planes)
+    - [Identity and aud semantics (SPIFFE)](#identity-and-aud-semantics-spiffe)
+    - [Token structure](#token-structure)
+    - [Issuance flow (sandbox-manager + identity)](#issuance-flow-sandbox-manager--identity)
+        - [Issuance credential: reuse the existing Team API Key](#issuance-credential-reuse-the-existing-team-api-key)
+        - [Issuer-side authorization: CanRequestToken (Decision A)](#issuer-side-authorization-canrequesttoken-decision-a)
+        - [Pluggable Signer detailed design](#pluggable-signer-detailed-design)
+        - [Issuer-side SDK design (extending the E2B SDK)](#issuer-side-sdk-design-extending-the-e2b-sdk)
+    - [Verification flow (sandbox-gateway filter)](#verification-flow-sandbox-gateway-filter)
+        - [JWKS retrieval and caching](#jwks-retrieval-and-caching)
+        - [Verification chain: authentication + authorization](#verification-chain-authentication--authorization)
+    - [Admin authorization: SandboxIngressPolicy CRD](#admin-authorization-sandboxingresspolicy-crd)
+    - [Two-sided enforcement and consistency](#two-sided-enforcement-and-consistency)
+    - [Revocation and key rotation](#revocation-and-key-rotation)
+    - [End-to-end sequence](#end-to-end-sequence)
+- [User Stories](#user-stories)
+- [Requirements](#requirements)
+- [Risks and Mitigations](#risks-and-mitigations)
+- [Alternatives](#alternatives)
+- [Upgrade Strategy](#upgrade-strategy)
+- [Test Plan](#test-plan)
+- [Implementation History](#implementation-history)
+
+## Summary
+
+This proposal introduces JWT-based authentication and **audience (aud)
+authorization** for the **ingress direction** of a sandbox (client →
+sandbox-gateway → envd/agent-runtime). Only a token whose `aud` explicitly
+contains the target sandbox's SPIFFE ID — and that passes signature, expiry,
+and admin policy checks — may reach that sandbox's envd ingress interfaces.
+
+Three parts:
+
+1. **Issuance plane** (sandbox-manager + `pkg/identity`): exchange an existing
+   Team API Key for a short-lived JWT carrying `aud`.
+2. **Authorization control plane**: a new namespaced CRD `SandboxIngressPolicy`
+   (symmetric to the egress `SecurityProfile`).
+3. **Verification data plane** (sandbox-gateway): upgrade the current static
+   `x-access-token` comparison into a tiered verification chain.
+
+envd itself is unchanged. The whole feature is guarded by the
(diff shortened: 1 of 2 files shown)


docs/proposals/20260725-sandbox-ingress-authn.md
```diff
@@ -0,0 +1,1076 @@
...
+  (Decision A)**; all fine-grained scope/interface authorization is enforced by
+  the gateway.
+
+## Glossary
+
+| Term | Meaning |
+|------|---------|
+| ingress | client → sandbox-gateway → envd/agent-runtime direction |
+| envd | agent-runtime sidecar inside a sandbox pod exposing E2B-compatible APIs |
+| **sandbox SPIFFE ID** | `spiffe://<trust-domain>/ns/{namespace}/sandbox/{name}` — the aud value |
```
zmberg: Have you considered a multi-cluster scenario? Would there be any issues in a multi-cluster setup?

HARSHRAJ2789: Not answering for @DahuK, but one data point on the multi-cluster question, since it turns on
something the document says about itself.

The rationale for `aud` = SPIFFE ID gives cross-cluster uniqueness as reason 3:

> 3. SPIFFE naming is readable, cross-cluster unique, and aligns with a standard zero-trust
>    identity model.

That holds when each cluster has its own trust domain. The examples here point the other way:
the block directly under that list is

```
spiffe://agents.kruise.io/ns/{namespace}/sandbox/{sandboxName}
```

and the SDK resolver carries the same value as its illustrative default
(`self._trust_domain = trust_domain  # e.g. "agents.kruise.io"`). `agents.kruise.io` names the
product, not an installation. If two clusters are configured that way and both run
`default/my-sandbox`, the two sandboxes have the same SPIFFE ID, so a token minted for one is
accepted by the other. Uniqueness then rests on an operator choosing a distinct trust domain
per cluster, which is a deployment convention and not a property of the naming scheme.

Worth contrasting with what the merged path already does, because it is the same question
answered differently. `sandbox-gateway` binds on the `sandbox` claim, never on `aud`:

```go
if claims.Sandbox.SandboxID != route.ID || claims.Sandbox.SandboxUID != string(route.UID) {
```

`filter.go:211`. `sandboxUid` is the Kubernetes object UID, so it is distinct across clusters
whatever the trust domain is set to, and a token replayed into another cluster fails the UID
comparison even when namespace and name collide.

That does not decide anything on its own, and I raised the broader `aud` question in my review
above. But if the ingress design keeps `aud` as the scoping mechanism, then either the trust
domain has to be per-installation and documented as such, or the `aud` check needs the UID
alongside it to get the property reason 3 claims.

DahuK:
Thanks @HARSHRAJ2789 for the exceptionally thorough review — you clearly track the community's direction closely, and every point checked out against the merged code and #772. And thanks @zmberg for the multi-cluster question. I've revised both documents:

1 (aud vs sandbox claim). RFC 7519 §4.1.3 is the right lens here, but the answer differs by direction, so I've made the two documents scope aud intentionally rather than uniformly:
- Outbound / egress (20260727-...token-exchange.md) is about the token the agent presents to external services. Since those tokens leave the cluster and are validated by external authorization systems we don't control, this direction is standards-first: aud follows RFC 7519 + RFC 8707 (Resource Indicators) and is the authoritative, higher-priority claim — it names the external target API so the token interoperates with any standard IdP/resource server unchanged; the sandbox is carried as azp per RFC 8693, not by overloading aud.
- Inbound / ingress (20260725-...ingress-authn.md) keeps the sandbox claim (sandboxId+sandboxUid) as the primary binding, aligned with the merged verifier and #772; aud remains available as optional defense-in-depth.

  The invariant that keeps them consistent: aud always names who is expected to accept the token — the external service on egress, the sandbox on ingress — never overloaded to mean something else. That removes the earlier "opposite directions" inconsistency you flagged.

2 (IssueToken signature). Agreed — I've dropped the removed TokenRequest shape and put RequestedAudience/Scope on #742's TokenOptions, so neither PR has to rebase around the other.

3 (#772). Confirmed #772 mints sandbox{sandboxId,sandboxUid} with no aud, and its SandboxBinding comment already states the anti-replay intent. The ingress design now uses exactly this claim as the primary anchor, so a #772 token is no longer rejected.

4 (propagation). Fair — "works without modification" was misleading. The token is of course issued, but the principle is that it must not land in the sandbox as a long-lived secret; delivery is deferred to the envd architecture split. Work on that path is already in progress — see #787 — and I've reworded the doc to reference it as related work rather than to claim the path already exists.

5 (multi-cluster, @zmberg). Right that agents.kruise.io as a shared trust domain breaks uniqueness. I've adopted two layers: a per-cluster trust domain (hard requirement) and the sandboxUid binding. I deliberately don't rely on the UID alone — Kubernetes doesn't guarantee cross-cluster UID uniqueness — so the per-cluster trust domain provides the cluster boundary while the UID provides the in-cluster binding, each backstopping the other.

HARSHRAJ2789:
Thanks for taking all of it seriously. Splitting `aud` by direction resolves the part I found
hardest to write up: with `azp` carrying the sandbox on egress and the `sandbox` claim carrying
it on ingress, `aud` means one thing in both documents, so they can be read together without a
translation table. Moving `RequestedAudience` and `Scope` onto #742's `TokenOptions` also
removes the ordering problem between the two PRs.

PR #699, opened 2026-07-28
feat(controller): route CSI mounts of TLS-capable sandboxes over HTTPS

BH4AWS:


pkg/utils/runtime/transport.go
```diff
@@ -94,21 +103,74 @@ func buildClientTLSConfig(m TLSMaterial, serverName string) (*tls.Config, error)
...
+// Semantics are strict by design: an empty dir means TLS is not configured and
+// yields (nil, nil), while a non-empty dir declares the intent to speak TLS, so
+// any problem (missing directory, missing ca.crt, unparsable material, an
+// unpaired client certificate) is an error the caller should surface at
+// startup instead of silently degrading to plain HTTP.
+//
+// Callers should re-invoke it per client construction rather than caching the
+// result: Secret rotation updates the mounted files in place, so a fresh read
+// naturally picks up rotated certificates.
+func LoadTLSMaterialFromDir(dir string) (*TLSMaterial, error) {
```
furykerry, line 128: it should be package private func

cmd/agent-sandbox-controller/main.go
```diff
@@ -307,10 +316,31 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Resolve the runtime client TLS material provider. A configured directory
+	// must be loadable now (fail fast on a broken mount); the provider re-reads
+	// the files on every use so Secret rotation is picked up without restarts.
+	var runtimeTLSMaterial agentsruntime.TLSMaterialProvider
+	if runtimeClientCertDir != "" {
+		if _, err := agentsruntime.LoadTLSMaterialFromDir(runtimeClientCertDir); err != nil {
```
furykerry, line 324: we have already load tls here , can we pass TLSMaterial directly to manager. It seems not necessary to pass runtimeTLSMaterial function if no tls refresh is needed

pkg/controller/sandbox/core/sandbox_initializer.go
```diff
@@ -41,10 +41,14 @@ type defaultSandboxInitializer struct {
 	apiReader       client.Reader
 	storageRegistry storages.VolumeMountProviderRegistry
 	recorder        record.EventRecorder
+	// tlsMaterial supplies the client TLS material for TLS-capable sandboxes.
+	// Nil means the controller is not configured for runtime TLS.
+	tlsMaterial utilruntime.TLSMaterialProvider
```
furykerry, line 46: consider rename to tlsBundle which is more commonly known

cmd/agent-sandbox-controller/main.go
```diff
@@ -307,10 +316,31 @@ func main() {
...
+			setupLog.Error(err, "unable to load runtime client TLS material", "dir", runtimeClientCertDir)
+			os.Exit(1)
+		}
+		certDir := runtimeClientCertDir
```
furykerry, line 328: why initiate a new variable, can we just use runtimeClientCertDir

cmd/agent-sandbox-controller/main.go
```diff
@@ -123,6 +124,14 @@ func main() {
 	flag.StringVar(&metricLabelsAllowlist, "metric-labels-allowlist", "",
 		"Comma-separated list of Sandbox label keys to expose as sandbox_labels metric labels (e.g., app,env,version)")
 
+	var runtimeClientCertDir string
+	flag.StringVar(&runtimeClientCertDir, "runtime-client-cert-dir", "",
+		"Directory holding the agent-runtime client TLS material (ca.crt, client.crt, client.key), "+
+			"typically the mount point of the runtime client certificate Secret. "+
```
furykerry, line 130: consider make the description more simpler e.g. in one sentence

pkg/controller/sandbox/core/sandbox_initializer.go
BH4AWS: done

cmd/agent-sandbox-controller/main.go
BH4AWS: done

PR #710, opened 2026-07-29
feat: support upgrading paused sandboxes via SandboxUpdateOps

zmberg:

Add `IncludePaused` field to `SandboxUpdateOpsSpec` to allow batch upgrading sandboxes in the Paused phase. When enabled, Paused sandboxes are resumed before upgrade and re-paused afterwards if `spec.paused` is still true.

NONE

1. UT: `go test ./pkg/controller/sandboxupdateops/... -count=1`
2. UT: `go test ./pkg/controller/sandbox/... -run TestCalculateStatus -count=1 -v`
3. E2E: create a paused sandbox, create SandboxUpdateOps with `includePaused: true`, verify the sandbox is upgraded and re-paused

Key changes:
- **API**: add `IncludePaused` field to `SandboxUpdateOpsSpec`
- **Controller**: `classifySandbox` includes Paused sandboxes when `IncludePaused` is true
- **Sandbox controller**: prioritize upgrade over resume when both are triggered in the Paused phase; re-pause after upgrade if `spec.paused` remains true
- **Upgrade control**: add Resuming stage with `resumeFunc` support for waking paused sandboxes before proceeding with the upgrade flow
- **CRD**: add `includePaused` field to the SandboxUpdateOps CRD
- **UT**: cover `IncludePaused` classification, post-upgrade re-pause/stay-running, and Resuming stage transitions
- **E2E**: cover paused sandbox upgrade with re-pause and mid-upgrade `spec.paused` change to false


furykerry (commented):
## Code Review

### Change Summary

**What changed** — Adds `IncludePaused` to `SandboxUpdateOpsSpec` so batch update ops can target `Paused` sandboxes. `calculateStatus` re-prioritizes upgrade detection over pause/resume in the `Running` and `Paused` branches, and `UpgradeControl` gains a new `Resuming` stage (with injected `resumeFunc`) that wakes a paused sandbox before `PreUpgrade`. After upgrade success, the sandbox re-pauses if `spec.paused` is still true. Resume core logic is extracted from `EnsureSandboxResumed` into `handlerResume` for reuse.

**Affected behavior** — agent-sandbox-controller (`sandbox` + `sandboxupdateops` reconcilers), API (`IncludePaused`, new condition reason `Resuming`), regenerated CRD. Schema is backward-compatible, but the `calculateStatus` reordering changes existing behavior even when `IncludePaused` is unused (see Warning 4).

---

### Critical Issues (Must Fix)

The three issues below chain together such that **the feature cannot work end-to-end as written**. The new E2E tests should fail if actually run against this branch — recommend verifying before merge.

#### 1. Paused sandboxes are filtered out *before* `classifySandbox` — `IncludePaused` is dead code in the reconcile path

**Location**: `pkg/controller/sandboxupdateops/sandboxupdateops_controller.go` — `classifySandboxes` (unchanged by this PR)

The outer loop pre-filter still skips every sandbox not in `Running`/`Upgrading`:

```go
if !sbx.DeletionTimestamp.IsZero() || (sbx.Status.Phase != agentsv1alpha1.SandboxRunning &&
    sbx.Status.Phase != agentsv1alpha1.SandboxUpgrading) {
    continue
}
```

A `Paused` sandbox never reaches the modified `classifySandbox`, so it is never selected as a candidate and never patched. It also corrupts progress accounting: an ops-labeled sandbox that is momentarily `Paused` is excluded from `total`, allowing `updated+failed == total && len(candidates) == 0` to flip the ops to `Completed` prematurely. `TestClassifySandbox_IncludePaused` calls `classifySandbox` directly, bypassing the filter, which is why it passes.

**Fix**:

```go
isPhaseEligible := sbx.Status.Phase == agentsv1alpha1.SandboxRunning ||
    sbx.Status.Phase == agentsv1alpha1.SandboxUpgrading ||
    (ops.Spec.IncludePaused && sbx.Status.Phase == agentsv1alpha1.SandboxPaused)
if !sbx.DeletionTimestamp.IsZero() || !isPhaseEligible {
    continue
}
```

Also add a `Reconcile`-level (not `classifySandbox`-level) test proving a Paused sandbox actually gets patched.

#### 2. The Paused→Upgrading transition can never fire: it requires `pod != nil`, but a completed pause implies `pod == nil`

**Location**: `pkg/controller/sandbox/sandbox_controller.go` — new `Paused` branch of `calculateStatus`

The new upgrade trigger is:

```go
if cond != nil && cond.Status == metav1.ConditionTrue {
    if pod != nil && pod.Labels[agentsv1alpha1.PodLabelTemplateHash] != newStatus.UpdateRevision && ...
```

But `EnsureSandboxPaused` deletes the pod and only sets the Paused condition to `True` **when `pod == nil`** (reason `DeletePod`). So `cond.Status == True` and `pod != nil` are mutually exclusive in steady state — the branch is unreachable, and a template patch on a fully paused sandbox never transitions it to `Upgrading`. Combined with issue 1, the whole paused-upgrade flow is inert; the E2E test would hang or fail at pod verification (NotFound).

**Fix**: detect the template change without a pod — the revision hash already captures it:

```go
if cond != nil && cond.Status == metav1.ConditionTrue {
    if newStatus.UpdateRevision != box.Status.UpdateRevision &&
        core.RequiresPodReplacementUpgrade(box) {
        newStatus.Phase = agentsv1alpha1.SandboxUpgrading
        utils.RemoveSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionUpgrading))
        // Keep SandboxConditionPaused: the Resuming stage relies on it.
    } else if !box.Spec.Paused { ...
```

Add a `TestCalculateStatus` case with `pod == nil`, phase `Paused`, cond `True`, and a changed revision — the current UT suite has no such case.

#### 3. Resuming stage deadlocks: the real `handlerResume` never creates the `Resumed` condition, so `resumedCond` stays nil forever

**Location**: `pkg/controller/sandbox/core/upgrade_control.go` (Resuming case) + `pkg/controller/sandbox/core/common_control.go` (`handlerResume`)

The Resuming stage waits for `Resumed == True`, but `handlerResume` only *flips an existing* `Resumed=False` condition to `True`:

```go
if resumedCond := utils.GetSandboxCondition(...); resumedCond != nil &&
    resumedCond.Status == metav1.ConditionFalse { ... set True ... }
```

In the normal resume path, `calculateStatus` seeds `Resumed=False` on Paused→Resuming. In the new upgrade path there is no such seeding (`Resumed` was removed when the sandbox paused, as the conditions are exclusive), so `handlerResume` recreates the pod but never sets `Resumed`, and the state machine waits in `Resuming` indefinitely.

The unit test masks this: the mock `resumeFn` in `TestEnsureSandboxUpgraded_Resuming` sets the condition **unconditionally**, diverging from `handlerResume`'s real semantics. The PR's own fixture change is the tell — `TestCommonControl_EnsureSandboxResumed_LegacyBackfill` had to manually append `Resumed=False` "so handlerResume can flip it to True".

**Fix**: seed the condition when entering the Resuming stage:

```go
case agentsv1alpha1.SandboxUpgradingReasonResuming:
    ...
    if utils.GetSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionResumed)) == nil {
        utils.SetSandboxCondition(newStatus, metav1.Condition{
            Type:               string(agentsv1alpha1.SandboxConditionResumed),
            Status:             metav1.ConditionFalse,
            Reason:             agentsv1alpha1.SandboxResumeReasonCreatePod,
        })
    }
    if err := r.resumeFunc(ctx, args); err != nil { ...
```

And make at least one test wire the **real** `control.handlerResume` (via `NewCommonControl`) through the Resuming path.

#### 4. Chinese code comments

**Location**: `pkg/controller/sandbox/sandbox_controller.go` — two new comments in `calculateStatus` (`// 如果是从 Running 进入到 Upgrading 状态…` and `// 这里不能删除  Condition SandboxConditionPaused…`). Repo rule: keep code comments in English. Suggested:

```go
// Entering Upgrading from Running must not carry a Paused condition.
...
// Do NOT remove SandboxConditionPaused here: the upgrade Resuming stage
// relies on it to decide whether the sandbox must be woken up first.
```

---

### Warnings (Should Fix)

1. **Nil-pointer risks in the Resuming case** (`upgrade_control.go`): (a) `pausedCond.Status` is dereferenced without a nil check; (b) `utils.GetPodCondition(&pod.Status, corev1.PodReady)` can run with `pod == nil` right after `resumeFunc` created the pod (if a stale `Resumed=True` was persisted and the pod was deleted externally) → panic. Guard both (`pausedCond != nil && ...`; `if pod == nil { return nil }` before the PodReady check).

2. **Mid-upgrade template change while in `Resuming` jumps to `PreUpgrade`, leaving inconsistent state**: `determineUpgradeResumeReason` has no `Resuming` case — both the gate-off path and `default:` return `PreUpgrade`. The upgrade then proceeds with a nil pod (implicit resume via `performRecreateUpgrade` creating the pod), and the `Paused` condition is never removed (only the Resuming→PreUpgrade transition removes it). After `Succeeded` → `Running` → re-pause: `EnsureSandboxPaused` sees the stale `Paused=True` and returns immediately — the sandbox reports `Paused` while its pod keeps running. Add `case SandboxUpgradingReasonResuming: return SandboxUpgradingReasonResuming`.

3. **Naming and stale docs** (`common_control.go`): `handlerResume` → `handleResume` (its doc comment even starts with "resume handles…"). The `ResumeFunc` doc says it "transitions the sandbox to running when the pod is ready" — no longer true; that responsibility deliberately stayed in `EnsureSandboxResumed`, which is exactly why the upgrade path can reuse it. Rewrite both comments to state it does **not** change the phase.

4. **Silent behavior change for plain pause requests**: in the `Running` branch, upgrade detection now takes priority over `spec.paused` for **all** sandboxes, not just those under an `IncludePaused` ops. A user pausing a running sandbox with a pending template change now sees it upgrade first and pause afterwards; previously pause won. Probably the right semantics, but it is user-visible and independent of opting into `IncludePaused` — call it out in the PR description/CHANGELOG and on `SandboxSpec.Paused` docs.

5. **`EnsureSandboxResumed` phase transition lacks a pod guard**: the new tail block sets `Phase=Running` and calls `syncStatusFromPod(pod, ...)` based purely on the `Resumed` condition; `defaultSyncStatusFromPod` dereferences `pod` unconditionally. Use `if pod != nil && resumedCond != nil && resumedCond.Status == metav1.ConditionTrue`.

---

### Suggestions

- Make the mock `resumeFn` replicate `handlerResume`'s actual contract (only flip an existing `Resumed=False`) — this would have caught Critical 3.
- The rewritten `Paused` branch now logs "sandbox pause not completed yet" on every reconcile while pausing; consider restoring the narrower condition or a debug log level.
- The nested double-negative in `classifySandbox` (`if !ops.Spec.IncludePaused || sbx.Status.Phase != SandboxPaused`) is hard to parse; a positive-form helper shared with the `classifySandboxes` pre-filter would fix Critical 1 and readability together.
- E2E: `Eventually(...Upgrading, 30s)` races a fast upgrade; consider asserting on the ops label/condition reason instead of a transient phase.

### Positive Findings

- CRD YAML matches `make manifests` output; no deepcopy change needed for the bool field.
- The `Paused` branch now nil-guards `cond` before `cond.Status` — the pre-existing code could panic; genuine robustness improvement.
- Good state-machine doc updates, and the Resuming stage follows the established one-state-per-reconcile pattern.
- Table-driven tests with descriptive names throughout; gofmt alignment cleanups are welcome.

### Bottom line

The API shape and state-machine design are sound, but Critical 1–3 form a chain where the feature is non-functional (Paused sandboxes never selected; Paused→Upgrading unreachable; Resuming stage would deadlock even if reached). The unit tests pass only because they bypass the outer filter and mock away the real resume semantics — running the two new E2E cases before merge would confirm this quickly.

pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -20,16 +20,15 @@ import (
 	"context"
 	"fmt"
 
+	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
+	"github.com/openkruise/agents/pkg/utils"
+	"github.com/openkruise/agents/pkg/utils/expectations"
```
furykerry: plz keep the repo specific dependency in the bottom of import list

pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -100,18 +102,65 @@ func (r *UpgradeControl) EnsureSandboxUpgraded(ctx context.Context, args EnsureF
...
+			return err
+		}
+		r.syncStatusFromPod(pod, newStatus, false)
+		// Resume succeeded, transition to PreUpgrade.
+		klog.InfoS("Sandbox resumed successfully, transitioning to PreUpgrade", "sandbox", klog.KObj(box))
+		upgradeCond.Reason = agentsv1alpha1.SandboxUpgradingReasonPreUpgrade
+		upgradeCond.Message = ""
+		utils.SetSandboxCondition(newStatus, *upgradeCond)
+		utils.RemoveSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionPaused))
+		return nil
```
furykerry: fallthrough here ?

pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -100,18 +102,65 @@ func (r *UpgradeControl) EnsureSandboxUpgraded(ctx context.Context, args EnsureF
...
 			Status:             metav1.ConditionFalse,
-			Reason:             agentsv1alpha1.SandboxUpgradingReasonPreUpgrade,
+			Reason:             initialReason,
 			LastTransitionTime: metav1.Now(),
 		}
 		utils.SetSandboxCondition(newStatus, *upgradeCond)
 	}
 
 	switch upgradeCond.Reason {
+	case agentsv1alpha1.SandboxUpgradingReasonResuming:
```
furykerry, line 168: one should pause the sandbox again once the upgrade is finished, plz comment in the code that how repausing is implemeneted.

pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -100,18 +102,65 @@ func (r *UpgradeControl) EnsureSandboxUpgraded(ctx context.Context, args EnsureF
...
+		// The sandbox was paused — resume it before proceeding with the upgrade.
+		pausedCond := utils.GetSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionPaused))
```
furykerry, line 170: consider change pauseTime if it is set, otherwise the sandbox may get paused again due to pauseTime trigger

api/v1alpha1/sandboxupdateops_types.go
```diff
@@ -46,6 +46,16 @@ type SandboxUpdateOpsSpec struct {
...
+
+	// IncludeStates specifies which sandbox states are eligible as upgrade
+	// candidates. When empty, defaults to [Running].
+	// Upgrading sandboxes are always tracked regardless of this field, as they
+	// represent in-progress upgrades owned by this ops.
+	// Supported values: Running, Paused.
+	// +optional
+	// +kubebuilder:validation:items:Enum=Running;Paused
+	// +listType=set
+	IncludeStates []SandboxPhase `json:"includeStates,omitempty"`
```
furykerry: Consider use a struct for for future expansion 
```

type SandboxUpdateOpsSpec struct {
     stateFilter UpgradeStateFilter `json:"stateFilter, omitempty"`
} 
struct UpgradeStateFilter {
    states []SandboxPhase
    // reserve for future use ,e.g. 
    // probes []string 
}

```

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -339,8 +364,19 @@ func (r *Reconciler) classifySandbox(ctx context.Context, sbx *agentsv1alpha1.Sa
 		return sandboxFailed
 	}
 
-	// Patched by ops but never entered upgrade flow — template already matched
+	// Patched by ops but never entered upgrade flow — template already matched.
```
furykerry, line 377: 1. L360 may need considering CheckpointFailed as sandboxFailed
2. consider label failed sandbox so that one can more easily filter failed sandbox in a large namespace

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -490,20 +490,26 @@ func (r *SandboxReconciler) calculateStatus(ctx context.Context, args core.Ensur
 			return newStatus, true
 		}
 
-		// If it is paused, first set the sandbox to the Paused state.
```
furykerry, line 493: there is no specific event for sandbox phase Paused → Upgrading , consider add mapping in sandboxPhaseTransitionMessages

pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -100,18 +103,85 @@ func (r *UpgradeControl) EnsureSandboxUpgraded(ctx context.Context, args EnsureF
...
+		if pausedCond == nil || pausedCond.Status == metav1.ConditionFalse {
+			// Sandbox is still pausing (or pause condition missing), wait for it to complete.
+			klog.InfoS("Sandbox is still pausing, waiting before upgrade", "sandbox", klog.KObj(box))
+			return nil
+		}
+		// Sandbox is paused, trigger resume.
+		if err := r.resumeFunc(ctx, args); err != nil {
```
furykerry, line 178: there is no resume event here, consider add record in  UpgradeControl , and record events for critical lifecycle transitions

PR #718, opened 2026-07-31
fix: delete claim-scoped TrafficPolicies owned by the Sandbox UID before returning to the pool (#709)

### LINKED ISSUE #709: [BUG] Recycled Sandbox retains claim-scoped TrafficPolicy
**What happened**:

A Sandbox with an E2B `TrafficPolicy` can be recycled and returned to its `SandboxSet` pool while the claim-scoped `TrafficPolicy` is left behind.

`TrafficPolicy` objects are controller-owned by the Sandbox UID, so Kubernetes garbage collection only removes them when the Sandbox itself is deleted. Recycling keeps the same Sandbox object alive. The recycle completion path resets Sandbox claim metadata and marks it unclaimed, but does not delete the owned `TrafficPolicy` objects.

The stale policy continues to select the Sandbox pod through the stable `agents.kruise.io/sandbox-name` label. This leaks network configuration across logical claims:

- A subsequent claim without `network` silently inherits the previous claim policy.
- A subsequent claim with `network` calls `CreateNetworkPolicy` and can create another policy with the same selector and priority.
- Multiple matching policies at the same priority have implementation-defined evaluation order.

The optional short Sandbox ID rollout can make the symptom more visible: an existing policy may retain a legacy ID annotation after the same Sandbox receives a short ID on a later claim, so ID-indexed Select/Update operations no longer find it. However, the root bug is the missing recycle cleanup and exists independently of short IDs.

Before a recycled Sandbox is marked successful, unclaimed, and available to the pool, the recycle flow should remove all claim-scoped `TrafficPolicy` objects owned by that Sandbox UID.

Cleanup failures should keep the Sandbox in recycling and be retried. The Sandbox should not become claimable until the cache observes that the policies are gone, so the next claim cannot overlap with stale data-plane policy.

1. Create a recycle-enabled Sandbox pool.
2. Claim a pooled Sandbox through the E2B API with a non-empty network configuration.
3. Confirm that a `TrafficPolicy` exists and selects the Sandbox pod by `agents.kruise.io/sandbox-name`.
4. Delete the logical Sandbox so sandbox-manager triggers recycle instead of deleting the Sandbox CR.
5. Wait until recycle succeeds and the Sandbox returns to the pool.
6. Observe that the `TrafficPolicy` still exists because its owner Sandbox UID is still alive.
7. Claim the same pooled Sandbox again:
   - without a network configuration, observe that the old policy still selects it; or
   - with a different network configuration, observe duplicate matching policies.

This behavior is present in the base of PR #686; that PR does not introduce the TrafficPolicy create/recycle lifecycle. PR #686 only exposes an additional legacy-ID-to-short-ID lookup symptom. The lifecycle fix should therefore be handled independently rather than migrating TrafficPolicy annotations as part of short-ID assignment.

Suggested implementation boundary:

- Add an informer-backed TrafficPolicy index keyed by the Sandbox controller owner UID.
- During recycle completion, list and delete every TrafficPolicy owned by the Sandbox UID, including historical duplicates.
- Treat List/Delete failures as retriable recycle failures and do not return the Sandbox to the pool early.
- Add recycle/reclaim coverage for both an absent network configuration and a replacement network configuration.

Relevant code:

- `pkg/controller/sandbox/core/recycle.go`: recycle completion and pool metadata reset.
- `pkg/sandbox-manager/infra/sandboxcr/network.go`: TrafficPolicy ownerReference, selector, create/update/select behavior.
- `pkg/cache/index.go`: current TrafficPolicy lookup index.

**Environment**:
- OpenKruiseAgent version: `master` / PR #686 base `68a7ac6`
- Kubernetes version (use `kubectl version`): N/A (code-path analysis)
- Install details (e.g. helm install args): recycle-enabled SandboxSet with E2B network configuration
- Others: Found during review of PR #686

lucifer4330k:
## What

Fixes [#709](https://github.com/openkruise/agents/issues/709): a recycled Sandbox keeps its UID, so the Kubernetes garbage collector does not clean up the claim-scoped `TrafficPolicy` objects created during the previous claim. Those orphan `TrafficPolicy` objects stay bound to the same UID and are silently inherited by the next claim, routing the new claim's traffic to the previous tenant.

## Why a recycled Sandbox is the leak

`handleRecycleGracePeriod` returns the Sandbox to the pool via `resetMetadataForPool` after `ShutdownTime`. The Sandbox object is never deleted, only its `status` and finalizer state are reset. `TrafficPolicy` carries an `OwnerReference` to the Sandbox with `Controller: true`, but Kubernetes GC only fires when the owner is *deleted*, not when its status changes. Without an explicit cleanup, every `TrafficPolicy` created during the previous claim outlives the claim and binds the next one.

## Fix

In `pkg/controller/sandbox/core/recycle.go`, `handleRecycleGracePeriod` now runs a new `deleteOwnedTrafficPolicies` helper **before** `resetMetadataForPool`:

1. List `TrafficPolicy` objects keyed by `fieldindex.IndexNameForOwnerRefUID = "ownerRefUID"` whose owner UID equals the Sandbox UID.
2. `Delete` each one; `IsNotFound` is success.
3. Re-list after the deletes; if anything still matches, return `RetriableError`.
4. Any other delete/list error is wrapped in `RetriableError` so the reconcile requeues.

## Index choice

The new `OwnerReference` lookup reuses the existing `pkg/utils/fieldindex.IndexNameForOwnerRefUID` index and `OwnerIndexFunc` (already used for `Sandbox`, `SandboxTemplate`, `Checkpoint`). `sandboxOwnerRef` in `pkg/sandbox-manager/infra/sandboxcr/network.go` already sets `Controller: &trueVar` on the `TrafficPolicy` `OwnerReference`, so `metav1.GetControllerOfNoCopy` extracts the Sandbox UID with no new field-index function or index name.

A new `IndexField` call for `&agentsv1alpha1.TrafficPolicy{}` is added in `pkg/utils/fieldindex/register.go` inside the existing `sync.Once` block, between the `Checkpoint` and commit-job registrations.

## Files

| File | Change |
| --- | --- |
| `pkg/utils/fieldindex/register.go` | Register `TrafficPolicy` on the existing `IndexNameForOwnerRefUID` index. |
| `pkg/controller/sandbox/core/recycle.go` | New `deleteOwnedTrafficPolicies` helper + call site before `resetMetadataForPool`. Imports `pkg/utils/fieldindex`. |
| `pkg/controller/sandbox/core/recycle_test.go` | 5 new table cases + `newTestRecycleControlWithDeleteInterceptor` helper + `WithIndex` registration on the fake client. |

## Test coverage

`TestEnsureSandboxRecycled` extended with five new cases on the existing table:

- zero claim-scoped `TrafficPolicy` objects — no-op, no error
- one owned policy — deleted, list afterwards is empty
- two owned policies — both deleted
- one policy owned by a different Sandbox UID — preserved (never deleted)
- delete interceptor returns an error — surfaced as `RetriableError`

`go test ./pkg/controller/sandbox/core/... -count=1` and `go test ./pkg/utils/fieldindex/... -count=1` pass. `go vet` is clean. `go build ./cmd/agent-sandbox-controller/...` succeeds. `gofmt -l` is clean.

- Issue: [#709](https://github.com/openkruise/agents/issues/709)

### DIFF
--- a/pkg/controller/sandbox/core/recycle.go
+++ b/pkg/controller/sandbox/core/recycle.go
@@ -328,11 +328,9 @@ func (r *SandboxRecycleControl) recyclePollingInterval(remaining time.Duration) 
 // controller-owner UID — which is stable across recycle — so we can enumerate
 // every policy that still belongs to this Sandbox and delete it explicitly.
 //
-// Any failure (list, individual delete, or a follow-up list that still
-// observes a remaining policy) is returned to the caller as a non-nil error
-// so handleRecycleGracePeriod wraps it in a RetriableError and forces a
-// requeue. The Sandbox is only returned to the pool once the cache observes
-// zero remaining policies.
+// Any failure (list or individual delete) is returned to the caller as a
+// non-nil error so handleRecycleGracePeriod wraps it in a RetriableError and
+// forces a requeue.
 func (r *SandboxRecycleControl) deleteOwnedTrafficPolicies(ctx context.Context, box *agentsv1alpha1.Sandbox) error {
 	log := klog.FromContext(ctx).WithName("recycle-traffic-policy-cleanup").WithValues("sandbox", klog.KObj(box), "uid", string(box.UID))
 
@@ -357,23 +355,6 @@ func (r *SandboxRecycleControl) deleteOwnedTrafficPolicies(ctx context.Context, 
 			log.Error(err, "Failed to delete TrafficPolicy", "trafficPolicy", klog.KObj(tp), "name", tp.Name)
 			return fmt.Errorf("delete TrafficPolicy %s/%s: %w", tp.Namespace, tp.Name, err)
 		}
-	}
-
-	if err := r.client.List(ctx, &tpList,
-		client.InNamespace(box.Namespace),
-		client.MatchingFields{fieldindex.IndexNameForOwnerRefUID: string(box.UID)},
-	); err != nil {
-		log.Error(err, "Failed to re-list TrafficPolicies after delete")
-		return fmt.Errorf("re-list TrafficPolicies by owner UID %q: %w", string(box.UID), err)
-	}
-	if len(tpList.Items) > 0 {
-		remaining := make([]string, 0, len(tpList.Items))
-		for i := range tpList.Items {
-			remaining = append(remaining, tpList.Items[i].Namespace+"/"+tpList.Items[i].Name)
-		}
-		log.Info("TrafficPolicies still observed after delete; will retry", "remaining", remaining)
-		return fmt.Errorf("%d TrafficPolicy(ies) still owned by sandbox %s/%s after delete: %v",
-			len(remaining), box.Namespace, box.Name, remaining)
 	}
 	return nil
 }
--- a/pkg/controller/sandbox/core/recycle.go
+++ b/pkg/controller/sandbox/core/recycle.go
@@ -34,6 +34,7 @@ import (
 	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
 	"github.com/openkruise/agents/pkg/identity"
 	"github.com/openkruise/agents/pkg/utils"
+	"github.com/openkruise/agents/pkg/utils/fieldindex"
 	agentsruntime "github.com/openkruise/agents/pkg/utils/runtime"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/controller/sandbox/core/recycle.go
```diff
@@ -318,6 +319,65 @@ func (r *SandboxRecycleControl) recyclePollingInterval(remaining time.Duration)
...
+
+	for i := range tpList.Items {
+		tp := &tpList.Items[i]
+		if err := r.client.Delete(ctx, tp); err != nil && !apierrors.IsNotFound(err) {
+			log.Error(err, "Failed to delete TrafficPolicy", "trafficPolicy", klog.KObj(tp), "name", tp.Name)
+			return fmt.Errorf("delete TrafficPolicy %s/%s: %w", tp.Namespace, tp.Name, err)
+		}
+	}
+
+	if err := r.client.List(ctx, &tpList,
```
furykerry: it is not necessary to list again, in fact , due to the informer sync latency, it is very likely that the next list still return previously deleted traffic policies.

lucifer4330k: Ok I removed the post-delete re-list and now rely on the initial UID-scoped list plus delete results.

lucifer4330k:
@furykerry i have made the changes , can u review this pr again .

PR #720, opened 2026-07-31
feat(controller): route CSI mounts of TLS-capable sandboxes over HTTPS

BH4AWS:
Add the per-sandbox dual-path CSI mount dispatch and wire the sandbox controller to it, keeping the legacy CLI path untouched for pre-TLS sandboxes:

- ProcessCSIMounts/doCSIMount accept variadic transport Options. Non-empty options route a mount through the runtime storage API (POST /v1/storage/mounts over HTTPS with forced resolution and mutual TLS). Empty options keep the legacy sandbox-storage CLI over the envd process protocol, so sandboxes without the TLS capability annotation are untouched. The storage API carries the typed CSI message while MountConfig still carries the pre-encoded blob the CLI consumes, so the base64 protobuf payload is decoded back into a NodePublishVolumeRequest before dispatch; an undecodable payload fails locally instead of reaching the runtime.
- NewTLSBundle is the single entry point for the runtime client certificate Secret layout (ca.crt required, client.crt/client.key as an optional pair). Semantics are strict: an empty directory disables TLS, a configured directory must yield a valid bundle, validated eagerly so a broken mount fails at startup instead of on the first runtime call. The bundle is a snapshot held for the process lifetime, so replacing the certificates requires a restart; that matches how the sandbox-gateway consumes the same Secret through its cert-init container and keeps the long-lived runtime certificates free of reload machinery.
- TransportOptionsFor turns fail-loud: a sandbox advertising AnnotationRuntimeTLSPort while the caller lacks a TLS bundle is now an error instead of a silent plaintext fallback, surfacing the misconfiguration rather than downgrading security.
- Controller wiring: the new --runtime-client-cert-dir flag loads the bundle at startup (fail fast on broken material) and threads it through Deps -> sandbox.Add -> SandboxControlArgs -> defaultSandboxInitializer, whose doInitialize resolves the per-sandbox transport and passes it to ProcessCSIMounts. The /init handshake stays on its existing transport in this phase.

The sandbox-manager claim path is intentionally not wired yet; it will reuse the same loader and dispatch in a follow-up.


pkg/utils/runtime/csi.go
```diff
@@ -127,13 +138,56 @@ func ProcessCSIMounts(ctx context.Context, sbx *agentsv1alpha1.Sandbox, opts con
...
+// "rtOpts non-empty": TransportOptionsFor is the single oracle deciding whether a
+// sandbox is served over TLS, so this function stays free of annotation parsing.
+func doCSIMount(ctx context.Context, sbx *agentsv1alpha1.Sandbox, opts config.MountConfig, rtOpts ...Option) (time.Duration, error) {
 	ctx = logs.Extend(ctx, "action", "csiMount")
 	start := time.Now()
+	if len(rtOpts) > 0 {
+		// New path: runtime storage API, which carries the typed CSI message
+		// instead of the encoded blob the CLI consumes, so the pre-encoded
+		// MountConfig.RequestRaw has to be decoded back first.
+		publishRequest, err := decodePublishRequest(opts.RequestRaw)
```
furykerry: plz save NodePublishVolumeRequest directly in the MountConfig, so as to avoid the wasted computing of encoding NodePublishVolumeRequest -> decode raw back to NodePublishVolumeRequest. We can encode NodePublishVolumeRequest before calling the legacy CSIMount func

pkg/utils/runtime/transport.go
```diff
@@ -43,6 +45,13 @@ const (
 
 	// pinnedDialTimeout bounds a single TCP dial to the sandbox Pod IP.
 	pinnedDialTimeout = 5 * time.Second
+
+	// Well-known file names inside a runtime client certificate directory, as
+	// laid out by the client certificate Secret mounted into control-plane
+	// components (see NewTLSBundle).
+	clientCADataKey   = "ca.crt"
```
furykerry: pkg/sandbox-gateway/runtimecredentials/credentials.go:38-44 had already define similar but different file name, consider unify the file name definition.

pkg/controller/sandbox/core/sandbox_initializer.go
```diff
@@ -69,15 +73,36 @@ func (d *defaultSandboxInitializer) Initialize(ctx context.Context, box *agentsv
...
 }
 
+// doInitialize resolves the per-sandbox mount transport and delegates to the
+// package-level Initialize. A sandbox advertising the runtime TLS capability
+// while this controller lacks a TLS bundle (or carries a broken annotation) is
+// a hard failure by design: silently downgrading a TLS-capable sandbox to the
+// plaintext CLI path would be a security regression, so the misconfiguration
+// must surface instead.
+func (d *defaultSandboxInitializer) doInitialize(ctx context.Context, box *agentsv1alpha1.Sandbox, newStatus *agentsv1alpha1.SandboxStatus) error {
+	rtOpts, err := utilruntime.TransportOptionsFor(box, d.tlsBundle)
```
furykerry: Transport is resolved unconditionally, so a misconfigured cluster fails initialization even for sandboxes with no CSI mounts. Consider build runtime options only for sandbox with csi mounts (at least for now), we can add a TODO to build options for every requests

cmd/agent-sandbox-controller/main.go
```diff
@@ -122,6 +124,8 @@ func main() {
 		"supporting three states: ip, memory, and filesystem. Format: comma-separated, e.g.: memory,filesystem")
 	flag.StringVar(&metricLabelsAllowlist, "metric-labels-allowlist", "",
 		"Comma-separated list of Sandbox label keys to expose as sandbox_labels metric labels (e.g., app,env,version)")
+	flag.StringVar(&runtimeClientCertDir, "runtime-client-cert-dir", "",
```
furykerry, line 127: with this new flag, is the ENABLE_AGENTRUNTIME_TLS really needed? we can stamp the sandbox if runtime-client-cert-dir is provided, and don't have to dealt with the case that ENABLE_AGENTRUNTIME_TLS is set to true but no runtime-client-cert-dir given.

furykerry, line 127: --runtime-client-cert-dir is not in the manager args, there is no Secret volume/mount, and no doc or overlay describes the expected directory. Combined with Critical #1, an operator has no in-repo reference for the layout to create.

Plz add a kustomize overlay (mirroring config/sandbox-gateway-runtime-mtls/) with the volume, mount path, and flag — even if disabled by default — or note in the PR body which follow-up delivers it.

pkg/controller/sandbox/core/interface.go
```diff
@@ -28,6 +28,7 @@ import (
 	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
 	"github.com/openkruise/agents/pkg/utils/expectations"
 	"github.com/openkruise/agents/pkg/utils/inplaceupdate"
+	utilruntime "github.com/openkruise/agents/pkg/utils/runtime"
```
furykerry: the three new imports use utilruntime for pkg/utils/runtime, which is the well-established Kubernetes alias for k8s.io/apimachinery/pkg/util/runtime;  consider using runtimeclient to reduce friction for future readers.

pkg/controller/controllers.go
```diff
@@ -25,17 +25,21 @@ import (
 	"github.com/openkruise/agents/pkg/controller/sandboxset"
 	"github.com/openkruise/agents/pkg/controller/sandboxupdateops"
 	"github.com/openkruise/agents/pkg/controller/securitytokenrefresh"
+	utilruntime "github.com/openkruise/agents/pkg/utils/runtime"
```
furykerry: the three new imports use utilruntime for pkg/utils/runtime, which is the well-established Kubernetes alias for k8s.io/apimachinery/pkg/util/runtime;  consider using runtimeclient to reduce friction for future readers.

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -46,6 +46,7 @@ import (
 	"github.com/openkruise/agents/pkg/utils"
 	"github.com/openkruise/agents/pkg/utils/expectations"
 	utilfeature "github.com/openkruise/agents/pkg/utils/feature"
+	utilruntime "github.com/openkruise/agents/pkg/utils/runtime"
```
furykerry: the three new imports use utilruntime for pkg/utils/runtime, which is the well-established Kubernetes alias for k8s.io/apimachinery/pkg/util/runtime;  consider using runtimeclient to reduce friction for future readers.

pkg/utils/runtime/transport.go
```diff
@@ -43,6 +45,13 @@ const (
 
 	// pinnedDialTimeout bounds a single TCP dial to the sandbox Pod IP.
 	pinnedDialTimeout = 5 * time.Second
+
+	// Well-known file names inside a runtime client certificate directory, as
```
furykerry, line 49: the constants are file names but are named ...DataKey (Secret-data semantics), while the comment says "file names". Consider renamed as XXXFile

cmd/agent-sandbox-controller/main.go
```diff
@@ -122,6 +124,8 @@ func main() {
...
+		"Directory holding the agent-runtime client TLS material (ca.crt, client.crt, client.key); empty disables runtime TLS.")
```
furykerry: the flag help string ends with a period while its neighbors don't. plz remove the ending period

pkg/utils/runtime/csi.go
```diff
@@ -105,7 +116,7 @@ func ProcessCSIMounts(ctx context.Context, sbx *agentsv1alpha1.Sandbox, opts con
 		go func(opt config.MountConfig) {
 			defer wg.Done()
 			defer func() { <-sem }()
-			mountDuration, err := doCSIMount(ctx, sbx, opt)
+			mountDuration, err := doCSIMount(ctx, sbx, opt, rtOpts...)
```
furykerry, line 123: ProcessCSIMounts logs "mountOptionConfig", opt on error and success (csi.go:118,123), and opt.RequestRaw is the base64 CSI payload — the very payload storage.go:60-64 says "must never be logged" because it carries Secrets and PublishContext in clear text. Worth a separate fix (log opt.Driver and the target path only).

pkg/utils/runtime/csi.go
```diff
@@ -127,13 +143,59 @@ func ProcessCSIMounts(ctx context.Context, sbx *agentsv1alpha1.Sandbox, opts con
...
+// "rtOpts non-empty": TransportOptionsFor is the single oracle deciding whether a
+// sandbox is served over TLS, so this function stays free of annotation parsing.
+func doCSIMount(ctx context.Context, sbx *agentsv1alpha1.Sandbox, opts config.MountConfig, rtOpts ...Option) (time.Duration, error) {
 	ctx = logs.Extend(ctx, "action", "csiMount")
 	start := time.Now()
-	err := CSIMount(ctx, sbx, opts.Driver, opts.RequestRaw)
+	if len(rtOpts) > 0 {
+		// New path: the runtime storage API carries the typed CSI message, so the
+		// request travels as-is and protojson owns its encoding.
+		_, err := NewRuntime(sbx, rtOpts...).Storage().Mount(ctx, CreateMountRequest{
```
furykerry, line 158: Each concurrent mount goroutine calls NewRuntime(sbx, rtOpts...), which rebuilds the TLS config (parses CA bundle, creates cert pool) each time. For sandboxes with many CSI mounts, this is redundant work. consider constructing the Runtime client once in ProcessCSIMounts and passing it to doCSIMount(no rtOpts required anymore)

PR #721, opened 2026-08-02
fix(e2e): add readiness probe and health check to TLS registry setup

furykerry:

The TLS registry Deployment deployed by `hack/setup-local-registry.sh` had no readiness probe. Without one, `kubectl wait --for=condition=available` returned as soon as the container process started — before the `registry:2` process opened its listening socket. The commit e2e test (`Commit to TLS registry`) then raced the startup and intermittently timed out (180s) waiting for `CommitPhase=Succeeded`, causing CI failures like the one in [PR #700's e2e run](https://github.com/openkruise/agents/actions/runs/30742415236/job/91482055892).

This PR adds two layers of protection:

1. **TCPSocket readiness probe** on port 5443 — gates the Deployment's `Available` condition on the port actually accepting connections, so `kubectl wait` only returns once the registry is listening.

2. **End-to-end post-setup verification** — after all node plumbing (certs.d CA, `/etc/hosts`) is in place, runs `curl --cacert <ca.crt> https://<fqdn>:5443/v2/` from a kind node. This validates DNS resolution, TLS trust, and HTTP serving in one shot. If it fails, the `Setup TLS Registry` CI step fails with a clear message instead of letting the commit test fail mysteriously minutes later.

Fixes the intermittent e2e failure observed in PR #700 CI run (run 30742415236, job 91482055892).

Run the e2e-1.32 CI workflow. The `Setup TLS Registry` step should:
- Wait for the registry pod's TCP readiness probe before proceeding
- Print `==> Verifying TLS registry reachability from kind nodes...` and succeed
- Print `==> TLS registry is ready at tls-registry.e2e-registry.svc.cluster.local:5443`

If the registry is unhealthy, the step fails immediately with a diagnostic message rather than causing a 180s timeout in the commit test.

- Only `hack/setup-local-registry.sh` is modified — no production code or test code changes.
- The readiness probe uses `failureThreshold: 20` with `periodSeconds: 3` (up to 60s of retry) to accommodate slow `registry:2` image pulls on first CI run.
- The `curl` verification relies on the CA cert already installed in `/etc/containerd/certs.d/` by the preceding loop, and the `/etc/hosts` entry added by the immediately preceding step.

### DIFF
--- a/hack/setup-local-registry.sh
+++ b/hack/setup-local-registry.sh
@@ -16,9 +16,11 @@
 # Deploy a self-signed TLS registry in the kind cluster for e2e tests.
 # This creates:
 #   1. A self-signed CA + server certificate
-#   2. A registry:2 Deployment with TLS enabled (port 5443)
+#   2. A registry:2 Deployment with TLS enabled (port 5443) and a TCP readiness
+#      probe so kubectl wait only returns once the registry is actually listening
 #   3. Configures kind nodes with certs.d so containerd/nerdctl trusts the CA
 #   4. Adds /etc/hosts entries on kind nodes for hostNetwork pod DNS resolution
+#   5. Verifies end-to-end reachability (DNS + TLS + HTTP) from a kind node
 
 set -o errexit
 set -o pipefail
@@ -97,6 +99,18 @@ spec:
           value: "/certs/tls.crt"
         - name: REGISTRY_HTTP_TLS_KEY
           value: "/certs/tls.key"
+        # Without a readiness probe the pod is marked Ready as soon as the
+        # container starts, but the registry:2 process needs a brief moment
+        # to open the listening socket. kubectl wait --for=condition=available
+        # would return before the port is actually accepting connections, causing
+        # an intermittent race with the commit e2e test that pushes to this
+        # registry. The TCP probe gates availability on the port being open.
+        readinessProbe:
+          tcpSocket:
+            port: ${REGISTRY_PORT}
+          initialDelaySeconds: 2
+          periodSeconds: 3
+          failureThreshold: 20
         volumeMounts:
         - name: certs
           mountPath: /certs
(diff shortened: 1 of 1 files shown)


HARSHRAJ2789 (commented):
Read this against master at `43765ba` and the workflow that runs it.

The readiness probe is the right shape for the race. `kubectl wait --for=condition=available`
at `setup-local-registry.sh:163` gates on the Deployment condition, and without a probe that
condition is satisfied as soon as the container starts, so the wait can return before
`registry:2` has opened the socket. A `tcpSocket` probe moves availability onto the port
itself, which is the thing the commit e2e actually needs.

One thing about the new verification block. Both loops above it run on every node:
`setup-local-registry.sh:146-160` installs the CA and `hosts.toml` per node, and
`:171-173` appends the `/etc/hosts` entry per node. The new loop breaks on the first node
that answers, so it reports success while some other node is still missing either piece.

That is not hypothetical here. `test/kind-conf.yaml` declares two nodes, a `control-plane`
and a `worker`, and `.github/workflows/e2e-1.32.yaml:37` passes that file to the kind action.
The `/etc/hosts` entry exists because the commit Job pods run with `hostNetwork: true` and so
resolve through the node's own DNS, which makes the node that matters whichever one those pods
land on. A check that stops at the first responder can pass entirely on the node that
never runs them.

Dropping the `break` and requiring every node to answer would make the check cover the
per-node plumbing it is verifying, at the cost of one extra `curl` in the two-node case.

Smaller point on the failure path: the `curl` carries `-s` and `2>/dev/null`, so when the
block fails the error message asks the reader to check `/etc/hosts`, the `certs.d` CA and the
pod logs, while the one line that would have said which of the three failed has been
discarded. Re-running the same `curl` without `-s`/`2>/dev/null` inside the failure branch
would print the reason next to the instruction to go find it.

I checked the `--cacert` path since it looked like a host path at first glance. It is correct:
`CERTS_DIR` is `/etc/containerd/certs.d/...` on the node, and `setup-local-registry.sh:151`
puts `ca.crt` there with `docker cp`, so the path resolves inside the `docker exec`.

PR #723, opened 2026-08-03
feat(runtime): add envd Filesystem ListDir/Remove atomic operations

BH4AWS:
Introduce the generated Go bindings for the envd Filesystem service and two control-plane helpers built on top of them, so callers can list a directory and remove a path inside a sandbox over the native envd protocol instead of shelling out via the Process service.

- proto/envd/filesystem: copy the generated filesystem.pb.go and filesystem.connect.go from the upstream envd repository, rewriting the connect import to the in-repo package path (per proto/envd/README.md).
- ListDirWithRuntime / RemovePathWithRuntime: connect/gRPC helpers that reuse the Process transport contract (X-Access-Token plus a caller supplied Basic identity). AuthUser is required because the service resolves every path per authenticated user. Failures are classified into ErrRuntimePathNotFound (NotFound) and ErrRuntimeFilesystemUnsupported (Unimplemented) so callers can treat a missing path or an older runtime as benign.
- Document the caller-visible service contract: Remove is recursive (os.RemoveAll) and idempotent, and ListDir excludes the directory itself while reporting each entry's type.


pkg/utils/runtime/filesystem.go
```diff
@@ -220,3 +225,183 @@ func buildRuntimeFilesEndpoint(runtimeURL, filePath, username string) string {
...
+// Two properties of the underlying service are load-bearing for callers:
+//
+//   - Removal is idempotent. A path that does not exist is not an error, so a
+//     caller retrying after a partial failure does not need to re-check first.
+//   - Removal is RECURSIVE. Passing a directory deletes its whole subtree, so a
+//     caller that means to delete only files must check the entry type (which
+//     ListDirWithRuntime reports) before calling this.
+//
+// A runtime that does not serve the Filesystem service yields an error wrapping
+// ErrRuntimeFilesystemUnsupported; every other failure is a plain wrapped error.
```
furykerry: the doc is incorrect, remove will also yields ErrRuntimePathNotFound

pkg/utils/runtime/filesystem.go
```diff
@@ -220,3 +225,183 @@ func buildRuntimeFilesEndpoint(runtimeURL, filePath, username string) string {
...
+	}
+	url := GetRuntimeURL(sbx)
+	if url == "" {
+		return nil, nil, nil, fmt.Errorf("runtime url not found on sandbox")
+	}
+	if timeout <= 0 {
+		timeout = defaultRuntimeFilesystemTimeout
+	}
+
+	client := filesystemconnect.NewFilesystemClient(http.DefaultClient, url, connect.WithGRPC())
```
furykerry: newFilesystemCall uses http.DefaultClient directly, while WriteFileWithRuntime (same file, line 168) uses the package-level runtimeFilesHTTPClient variable specifically introduced. Consider reuse runtimeFilesHTTPClient or introduce a similar package level client variable

pkg/utils/runtime/filesystem.go
BH4AWS: done

pkg/utils/runtime/filesystem.go
BH4AWS: done

PR #724, opened 2026-08-03
perf(sandbox-gateway): reduce informer cache memory

chengzhycn:

- Configure a controller-runtime cache transform only for the `Sandbox` GVK in sandbox-gateway.
- Project each cached Sandbox to the metadata, annotations, spec, and status fields required by route and state calculation, dropping Pod templates, PVC templates, managed fields, unrelated annotations, and unused status data.
- Allocate independent maps, slices, owner references, and timestamps so projected objects do not retain mutable source collections or large backing storage.
- Add contract tests covering lifecycle states, SandboxSet ownership, deletion and shutdown handling, authentication fields, IPv4/IPv6 routes, idempotence, and source-data independence.

The same production-shaped Sandbox model was tested with 200,000 objects on Kubernetes `v1.36.2-aliyun.1`:

| Metric | Before | After | Change |
|---|---:|---:|---:|
| 0-object Working Set | 40.41 MiB | 39.05 MiB | Comparable baseline |
| 200k Working Set | 3215.94 MiB | 1118.17 MiB | -65.23% |
| 200k incremental Working Set | 3175.53 MiB | 1079.11 MiB | -66.02% |
| 200k startup peak RSS | 3860.41 MiB | 1115.23 MiB | -71.11% |
| Increment per Sandbox | ~16.26 KiB | ~5.53 KiB | ~-66% |
| Increment per 10k Sandboxes | 157.15 MiB | 53.96 MiB | -103.19 MiB |

All 63 route samples across three independent 200k cold starts returned HTTP 200. The gateway had zero restarts and no OOMs, and the test namespace remained at zero Pods and PVCs.

NONE

Run the sandbox-gateway unit and race tests:

```bash
go test ./pkg/sandbox-gateway/... -count=1
go test -race ./pkg/sandbox-gateway/controller -count=1
```

The projection contract tests verify that full and projected Sandboxes produce identical Route and State results while unused fields are dropped and retained reference data is independently copied.

For a scale verification:

1. Run one gateway replica with the Sandbox controller disabled and prohibit Pods/PVCs in the test namespace.
2. Create 200,000 production-shaped `Running` and `Ready=True` Sandbox objects with non-empty `status.podInfo.podIP`.
3. Cold-start the gateway and sample `container_memory_rss` and `container_memory_working_set_bytes` from kubelet cAdvisor.
4. Verify representative routes, restart count, and absence of runtime Pods/PVCs.
5. Wait for a 10-minute window with less than 1% memory variation before comparing results.

The measured image digest was `sha256:5ad695f3bfa6873b4e6ddc81db2ff0d0ca44e92928b371208f6939e42e576a64`.

- The projection allowlist is now part of the gateway Route/State contract. Future route or state dependencies must update the transform and parity tests together.
- `UnsafeDisableDeepCopy` remains disabled, and paused/non-running route behavior is unchanged.
- The transform reduces steady-state memory substantially, but the measured 200k increment is still 1079.11 MiB. It does not yet meet the initial `<=600 MiB` target or the `>=70%` steady-state reduction target. Follow-up Go heap profiling and registry/cache attribution are required.
- The 200k startup peak reduction does meet the 70% target and increases headroom under the existing 4 GiB limit.
- WatchList is enabled by the current client-go version, but clusters that fall back to a traditional LIST still require startup headroom testing.


furykerry:
@chengzhycn is it possible to only strip managedFields which is big and not used by oka anyway?

furykerry:
@chengzhycn consider only strip away managedFields and EmbeddedSandboxTemplate, gateway feature may require other gateway fields, e.g. in https://github.com/openkruise/agents/pull/586, 

#586 consumer | Field read from gateway cache | Dropped by #724?
-- | -- | --
GetRouteFromSandbox → Route.WakeOnTraffic (primary wake check in shouldWakeSandbox) | annotations["agents.kruise.io/wake-on-traffic"] | ✅ yes
wake.Waker.HasWakeAnnotation (registry-sync-gap fallback, reads via mgr.GetClient() — the same transformed cache, since #586 does cache.NewCache(mgr, true) on this manager) | annotations["agents.kruise.io/wake-on-traffic"] | ✅ yes
wake.Waker.wakeInternal wake-timeout selection | annotations["agents.kruise.io/wake-timeout-seconds"] | ✅ yes
wake.Waker.wakeInternal auto-pause detection: autoPause := sbx.Spec.PauseTime != nil | Spec.PauseTime | ✅ yes

chengzhycn:
Follow-up 200k Sandbox memory benchmark for the final conservative transform that strips only `metadata.managedFields` and `spec.embeddedSandboxTemplate`.

Tested image:

```text
registry-cn-hangzhou.ack.aliyuncs.com/acs/ack-sandbox-gateway:feat_sandbox-gateway-informer-cache_ae56eee8_202608140149
sha256:311c7bd44a0ea04129a5a75bb5d4322761e6c06346baef977b1e0e8e12fb63ab
```

The benchmark used the same 200,000 full simulated Running Sandbox objects and cAdvisor sampling method as the previous original, allowlist, and managedFields-only runs. Each result below is the median of three independent cold starts.

| Metric | Original | Allowlist | managedFields only | managedFields + EmbeddedSandboxTemplate |
|---|---:|---:|---:|---:|
| 0-object working set | 40.41 MiB | 39.05 MiB | 40.30 MiB | 42.21 MiB |
| 200k working set | 3215.94 MiB | 1118.17 MiB | 2814.09 MiB | **1265.83 MiB** |
| 200k steady-state delta | 3175.53 MiB | 1079.11 MiB | 2773.79 MiB | **1223.62 MiB** |
| Startup peak RSS | 3860.41 MiB | 1115.23 MiB | 3155.39 MiB | **1484.39 MiB** |
| Per-Sandbox delta | 16.26 KiB | 5.53 KiB | 14.20 KiB | **6.26 KiB** |
| Steady CPU | 0.0145 core | 0.00513 core | 0.01302 core | **0.00781 core** |

Three final runs:

| Run | Startup peak RSS | Steady RSS | Working set | CPU |
|---:|---:|---:|---:|---:|
| 1 | 1504.14 MiB | 969.96 MiB | 1325.73 MiB | 0.00794 core |
| 2 | 1484.39 MiB | 1027.12 MiB | 1242.46 MiB | 0.00745 core |
| 3 | 1408.64 MiB | 955.31 MiB | 1265.83 MiB | 0.00781 core |

Compared with the original implementation, the final transform reduced the 200k working set by **60.64%**, the steady-state delta by **61.47%**, and startup peak RSS by **61.55%**. Compared with managedFields-only, it reduced working set by **55.02%** and startup peak RSS by **52.96%**.

The final working set is 13.21% above the allowlist projection, but this approach preserves annotations, `spec.pauseTime`, and all other Sandbox fields. It therefore captures most of the allowlist memory benefit without maintaining a field allowlist that can silently omit fields needed by future gateway features.

`EmbeddedSandboxTemplate` was only 547 bytes in the serialized 3524-byte test object, yet removing it saved about 1548 MiB of working set relative to managedFields-only at 200k objects. The dominant cost is therefore the decoded nested Pod/PVC Go object graph, DeepCopy allocations, and long-lived cache objects rather than serialized JSON size.

Correctness checks: the final namespace contained exactly 200,000 expected Sandbox names, 9 sampled objects remained `Running / Ready=True / PodIP=127.0.0.1`, all **63/63** route probes returned HTTP 200 across three cold starts, gateway restart count remained 0, and no Pod or PVC was created in the benchmark namespace.

PR #725, opened 2026-08-03
fix(runtime): harden command RPC against lost grpc-status trailers

BH4AWS:
A CSI mount claim failed with "internal: protocol error: no Grpc-Status trailer: unexpected EOF" although the mount command had run to completion and NodePublishVolume had succeeded: the gRPC response stream ended cleanly but carried no grpc-status, so the claim was failed and the sandbox was reserved for debugging instead of being handed out.

- Give the Process/Start RPC its own http.Client with DisableKeepAlives instead of sharing runtimeGRPCHTTPClient, so a command RPC never inherits a pooled or half-closed connection whose tail bytes carry the RPC status. RunCmdFuncArgs.HTTPClient allows overriding it per call for tests and transport experiments.
- Record whether a Process End event arrived and, when it did, tolerate the missing-trailer protocol error: the exit status is already known, so the caller decides based on ExitCode. Calls without an End event keep failing, because then ExitCode == 0 only means "never observed".
- Log pid/exitCode/exited/endReceived on command failure so a transport problem is distinguishable from a command that really failed.

runtimeGRPCHTTPClient now backs only the Filesystem service, and its comment records why the Process path deliberately diverges on transport. The Filesystem RPCs carry their result in trailers too, so they may want the same treatment later; that is left out here to keep this fix scoped to the failure actually observed.

### DIFF
--- a/pkg/utils/runtime/command_transport.go
+++ b/pkg/utils/runtime/command_transport.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2026.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package runtime
+
+import (
+	"net"
+	"net/http"
+	"strings"
+	"time"
+
+	"connectrpc.com/connect"
+)
+
+// runtimeCommandHTTPClient is the HTTP client used by RunCommandWithRuntime for the
+// streaming Process/Start RPC.
+//
+// It deliberately does not use http.DefaultClient: the default client shares a
+// process-wide connection pool with every other net/http user in the binary, so a
+// long-lived streaming RPC whose result is carried in HTTP trailers becomes sensitive
+// to unrelated traffic and to stale pooled connections. DisableKeepAlives keeps each
+// command on its own connection, which makes a failure attributable to that single call.
+//
+// Intentionally no http.Client.Timeout is set: RunCommandWithRuntime wraps the caller's
+// context with args.Timeout, which stays the single source of truth for deadlines
+// (mirrors runtimeFilesHTTPClient).
+//
+// It is a variable so tests can substitute their own transport.
+var runtimeCommandHTTPClient = &http.Client{
+	Transport: newRuntimeCommandTransport(),
+}
+
+// newRuntimeCommandTransport builds the transport backing runtimeCommandHTTPClient.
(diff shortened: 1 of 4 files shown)


HARSHRAJ2789 (commented):
This has been red since master moved, so I resolved the conflict locally against `43765ba` to
see what the rebase actually costs. It is one file and three edits, and two of them are quiet
enough to survive a careless merge.

The conflict is entirely with #734, your own, merged 7 Aug 10:25. It turned
`RunCommandWithRuntime` into `processAPI.Run` plus `resolveTransport`. This branch was pushed
7 Aug 13:28 from a base of `ce3b859` (#723), so it still carries the pre-#734 shape of that
function. The trailer-tolerance half applies cleanly; `process.go` is the only file git stops
on.

First, `HTTPClient` has to move from `RunCmdFuncArgs` to `RunCommandRequest`.
`processAPI.Run` never sees `RunCmdFuncArgs`, so `httpClient := args.HTTPClient` does not
compile. Worth flagging because `command_transport_test.go` only asserts on the package-level
`runtimeCommandHTTPClient` and never sets the override, so no test fails if the field ends up
somewhere it cannot be read from. Moving it also means `RunCommandWithRuntime` has to forward
it explicitly, which is your call on whether the wrapper should expose it at all.

Second, the command client belongs inside `resolveTransport`, as its `plainClient` argument.
The textual merge leaves `httpClient` assigned after the call, which overwrites what
`resolveTransport` returned and sends TLS sandboxes down a client with no pinned transport.
Passing it in reads:

```go
plainClient := req.HTTPClient
if plainClient == nil {
    plainClient = runtimeCommandHTTPClient
}
base, httpClient, err := p.r.resolveTransport(sbx, plainClient)
```

Third, `baseLog` is gone. The new form defines only `log`, already at
`V(utils.DebugLogLevel)` (`process.go:133`). Mapping `baseLog` onto it demotes the
tolerated-trailer message to debug, which the comment directly above that call says it does
not want. It needs a fresh logger from `ctx`.

With those three, the merge builds and `./pkg/utils/runtime/...` is green, including all of
the new tests here.

One thing the rebase confirms about scope, since I went looking for a hole and found the
opposite. `resolveTransport` (`client.go:293-296`) uses `plainClient` only when there is no
dial IP; with one it builds `newPinnedTransport(...)`, which already sets
`DisableKeepAlives: true` (`transport.go:310`). So `runtimeCommandHTTPClient` changes
behaviour on the plaintext path alone, where `runtimeGRPCHTTPClient` is a bare
`&http.Client{}` (`process.go:53`) sitting on `http.DefaultTransport`.

That is the path the reported failure was on. `doCSIMount` documents the split at
`csi.go:146-151` and dispatches on `rtOpts` being non-empty: the runtime storage API over
HTTPS, and the legacy `sandbox-storage` CLI over the plaintext envd process protocol for
everything else. `CSIMount` at `:178` is the legacy arm and forwards no `rtOpts`, so the mount
command runs on `runtimeGRPCHTTPClient` and its pooled connections. The fix lands on exactly
the transport that produced the trailer loss, and the TLS arm needs nothing because it already
has the property.

PR #726, opened 2026-08-04
feat: add InPlace update strategy proposal for SandboxUpdat…

Liquorice-Ma:
## feat(sandboxupdateops): add InplaceUpdate update strategy

This PR adds the `InplaceUpdate` strategy to SandboxUpdateOps (SUO), enabling batch in-place Pod updates for container images, resources, and template metadata without recreating Pods.

`InplaceUpdate` uses the existing SUO upgrade lifecycle. A target Sandbox enters `Upgrading`, runs `PreUpgrade` and `PostUpgrade` hooks when configured, performs the `UpgradePod` step by patching the existing Pod, and returns to `Running` only after the target configuration and Pod readiness are observed.

Unlike the SandboxClaim delivery path, which has no `upgradePolicy` and remains in `Running`, SUO reports progress and outcomes through the `Upgrading` condition.

### I. What this PR does

1. **API support**
   - Adds `InplaceUpdate` to both `SandboxUpdateOpsStrategyType` and `SandboxUpgradePolicyType`.
   - Updates the API validation markers for the new enum value.

2. **SUO validation and patching**
   - Prevents switching `spec.updateStrategy.type` to or from `InplaceUpdate` after creation. Existing `Recreate` ↔ `CheckpointRestore` behavior is unchanged.
   - Validates an `InplaceUpdate` patch before applying it. The patch must preserve the Sandbox immutable template hash, which restricts it to container image changes, container resource changes, and template metadata changes.
   - Records a `ValidationFailed` warning event on the SUO when a target Sandbox cannot be updated in place.
   - Sets `spec.upgradePolicy.type: InplaceUpdate` when applying an InplaceUpdate SUO patch, which routes the Sandbox through the explicit upgrade lifecycle.
   - Preserves the existing two-phase handling for paused Sandboxes: resume first, then apply the template patch.

3. **Sandbox upgrade lifecycle**
   - Treats `InplaceUpdate` as an explicit upgrade policy, so Sandboxes transition through `Upgrading` instead of using the Claim-time `Running`-phase update path.
   - Runs the shared upgrade lifecycle, including `PreUpgrade` and `PostUpgrade` hooks.
   - Executes the `UpgradePod` step through `performInplaceUpgrade`, which patches the existing Pod instead of deleting and recreating it.
   - Keeps existing `Recreate` and `CheckpointRestore` lifecycle behavior unchanged.

4. **Shared in-place update engine**
   - Uses one shared in-place update engine for Claim and SUO paths, without Claim/SUO mode parameters.
   - Applies the same QoS pre-check and resource-coverage semantics to both paths. Environment-provided resource increases that cover the requested target are treated as satisfied.
   - Supports `latest-target-wins`: a new target revision can supersede a previous pending or completed in-place update state.
   - Records the target image in in-place state so a corrective update that returns to an earlier image can complete even when the image ID does not change.

5. **Status and observability**
   - SUO classifies target Sandboxes using the `Upgrading` condition for all upgrade strategies.
   - The SUO in-place path does not write the Claim-specific `InplaceUpdate` condition; its outcome is reported through `Upgrading`.
   - Emits `ValidationFailed` and `PatchFailed` warning events on the SUO for validation and patch failures.
   - Emits a `SandboxUpgrading` normal event after a target Sandbox is successfully patched.

### II. Does this pull request fix one issue?

NONE

### III. How to verify

1. Create a `SandboxUpdateOps` object with:

   ```yaml
   spec:
     updateStrategy:
       type: InplaceUpdate


docs/proposals/20260804-suo-inplace-strategy.md
```diff
@@ -0,0 +1,317 @@
...
+(container resources for QoS, `inplace-update-state` annotation). These checks
+are handled by the sandbox controller, which reports results via the
+`InplaceUpdate` condition on the Sandbox status.
+
+### Change 1: API Types — New InPlace Enum
+
+**File**: `api/v1alpha1/sandboxupdateops_types.go`
+
+Add `SandboxUpdateOpsStrategyInPlace` constant and update the kubebuilder enum
+validation annotation to include `InPlace`. Run `make generate manifests` to
```
zmberg: InPlace -> InplaceUpdate

Liquorice-Ma: done，and add a Flow Diagrams

pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -39,11 +39,16 @@ const (
 	EventUpgradeResumed           = "UpgradeResumed"
 	EventUpgradePreUpgradeFailed  = "PreUpgradeFailed"
 	EventUpgradePodReplaced       = "UpgradePodReplaced"
+	EventUpgradePodUpdatedInPlace = "UpgradePodUpdatedInPlace"
```
zmberg: UpgradePodUpdatedInPlace -> UpgradePodInplaceUpdate

pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -95,14 +103,48 @@ func (r *UpgradeControl) recordUpgradeEvent(box *agentsv1alpha1.Sandbox, eventTy
...
+// sandbox through the Upgrading phase and its lifecycle state machine
+// (PreUpgrade → Checkpointing → UpgradePod → PostUpgrade).
+//
+// All explicit upgrade policies qualify, including InplaceUpdate: even though
+// an in-place upgrade keeps the pod, running it through the Upgrading phase
+// makes the upgrade observable (Phase=Upgrading, Ready=False) and lets lifecycle
+// hooks run. A sandbox without an upgrade policy does not enter the phase; a
+// template change on that path is applied in place from the Running phase,
+// which is what the SandboxClaim delivery flow depends on.
+func RequiresUpgradePhase(box *agentsv1alpha1.Sandbox) bool {
```
zmberg: RequiresUpgradePhase -> RequiresUpgradeSandbox

pkg/controller/sandbox/core/upgrade_control.go
Liquorice-Ma: done

pkg/controller/sandbox/core/upgrade_control.go
Liquorice-Ma: done

pkg/controller/sandbox/core/common_inplace_update_handler.go
```diff
@@ -128,18 +128,27 @@ func handleInPlaceUpdateCommon(
 		return false, err
 		// state!=nil indicates that an in-place upgrade has already been performed previously.
 	} else if state != nil {
```
zmberg: The code at L72-L82 should no longer simply return true — it would be better to return a specific error type and let the caller handle it:

- In the Claim phase, this kind of error should be handled as: Condition Ready=True, Condition InplaceUpdate=False, with the Message explaining the reason for the error.
- In the Upgrading phase, this kind of error should be reflected as Reason UpgradePodFailed, with the Message explaining the reason.

Besides, for case 2, how is the user supposed to proceed afterwards? There are two scenarios:

1. The upgrade must go through — in that case a Recreate upgrade has to be used, and we need to check whether it can still work.
2. Roll back — give up on the upgrade for now; what should the user do to get back to the Running state?

Both scenarios need to be considered in the code and covered by tests to verify they behave as expected.

pkg/controller/sandbox/core/common_inplace_update_handler.go
```diff
@@ -128,18 +128,27 @@ func handleInPlaceUpdateCommon(
 		return false, err
 		// state!=nil indicates that an in-place upgrade has already been performed previously.
```
zmberg: L93-L95 currently uses the InplaceUpdate Condition. I'd prefer to keep the two paths separate:

- For the Claim scenario, keep it as is — use the Ready and InplaceUpdate Conditions. 
- For the Upgrading scenario, use the Upgrading Condition instead. Don't couple it with the InplaceUpdate Condition in any way.

So the handleInPlaceUpdateCommon function still needs to be refactored — simply invoking it isn't enough.

pkg/controller/sandbox/core/common_inplace_update_handler.go
```diff
@@ -20,237 +20,340 @@ import (
...
+// classifyInplaceError:
+//   - inplaceClassUntrackedPod, inplaceClassUnsupportedChange: the template
+//     change cannot be applied in place; the pod is untouched.
+//   - inplaceClassTerminalRejected: retrying the same patch cannot succeed
+//     (QoS change, kubelet-side terminal failure, resize unsupported).
+//   - inplaceClassStateCorrupted: the state annotation cannot be parsed.
+//   - inplaceClassMetadataTransient: transient metadata fast-path failure.
+//   - inplaceClassTransient (any other error): a transient patch failure
+//     worth a requeue.
+func runInplaceUpdateStep(
```
zmberg, line 186: runInplaceUpdateStep -> handleInPlaceUpdateCommon

pkg/controller/sandbox/core/common_control.go
```diff
@@ -187,6 +192,16 @@ func (r *commonControl) EnsureSandboxUpdated(ctx context.Context, args EnsureFun
 			// Ready=False/InplaceUpdate conditions set during the update.
 			return nil
 		}
+		if isMetadataOnly {
```
zmberg, line 195: This code only prints logs and serves no real purpose — let's remove it.

pkg/utils/inplaceupdate/inplace_update.go
```diff
@@ -53,6 +53,26 @@ func (e *ResizeNotSupportedError) Unwrap() error {
 	return e.Err
 }
 
+// ImagePullFailedError indicates that a container tracked by an in-flight
+// in-place update cannot pull its target image (ErrImagePull /
+// ImagePullBackOff). Sandboxes must fail fast: even a pull failure that could
+// eventually self-heal (registry rate limit, network jitter) resolves slower
+// than the customer-facing deadline of a few seconds, so callers treat every
+// pull failure as terminal instead of waiting out the kubelet's retry backoff.
+type ImagePullFailedError struct {
```
zmberg, line 62: ImagePullFailedError -> ContainerStartFailedError

pkg/utils/inplaceupdate/inplace_update.go
```diff
@@ -509,20 +529,26 @@ func (c *InPlaceUpdateControl) patchPodResources(ctx context.Context, logger klo
...
 	logger := logf.FromContext(ctx).WithValues("pod", klog.KObj(pod))
 
-	state, err := GetPodInPlaceUpdateState(pod)
-	if state == nil || err != nil {
+	if state == nil {
 		return true, nil
 	}
 	if state.UpdateImages {
 		if !isPodImageUpdateCompleted(pod, state) {
+			if terminalErr := checkPodImagePullFailed(pod, state); terminalErr != nil {
```
zmberg, line 549: checkPodImagePullFailed -> checkContainerStartFailed

pkg/controller/sandbox/core/common_inplace_update_handler.go
```diff
@@ -20,237 +20,340 @@ import (
...
-			Message:            "",
+		if terminalErr != nil {
+			return inplaceUpdateStepInProgress, "", wrapInplaceError(inplaceClassTerminalRejected,
+				"in-place pod update failed", terminalErr)
 		}
-		utils.SetSandboxCondition(newStatus, cond)
-		return true, nil
+		// Patch applied, waiting for the kubelet to finish; report what the pod
+		// status says is holding the round up (e.g. ImagePullBackOff).
+		return inplaceUpdateStepInProgress, describeInplaceWaitReason(pod), nil
```
zmberg, line 234: describeInplaceWaitReason(pod) can never actually reach this point — the if terminalErr != nil check above would have already returned early.

pkg/controller/sandbox/core/common_inplace_update_handler.go
```diff
@@ -20,237 +20,340 @@ import (
...
-		// currently, multiple in-place updates are not supported.
-		klog.FromContext(ctx).Info("currently, multiple in-place updates are not supported", "sandbox", klog.KObj(box))
-		handler.GetRecorder().Eventf(box, corev1.EventTypeWarning, "InplaceUpdateForbidden",
-			"currently, multiple in-place updates are not supported")
-		completed, terminalErr := inplaceupdate.IsInplaceUpdateCompleted(ctx, pod)
-		if !completed {
-			if terminalErr != nil {
-				klog.FromContext(ctx).Info("previous in-place resize is infeasible, skipping", "sandbox", klog.KObj(box), "error", terminalErr)
+	// The target revision differs from the pod's: a still-running previous
+	// round blocks the new one (fail-stop; see the MAINTAINERS note above).
```
zmberg, line 238: I'm torn on this logic — I'd like to remove it but I'm not entirely sure it's safe. There's one scenario that needs testing first:
1. In-place upgrade to an incorrect image
2. Then in-place upgrade to a correct image

If this scenario works properly, let's keep the code as is.

pkg/controller/sandbox/core/common_inplace_update_handler.go
```diff
@@ -20,237 +20,340 @@ import (
...
+	// directly without starting a full in-place round, so callers do not have
+	// to treat the change as an update in progress.
 	if isMetadataOnlyChange(pod, box) {
-		klog.FromContext(ctx).Info("metadata-only change detected, patching pod metadata directly", "sandbox", klog.KObj(box))
-		opts := inplaceupdate.InPlaceUpdateOptions{
-			Pod:      pod,
-			Box:      box,
-			Revision: newStatus.UpdateRevision,
+		logger.Info("metadata-only change detected, patching pod metadata directly", "sandbox", klog.KObj(box))
+		if _, err := deliverInplacePatch(ctx, control, pod, box, targetRevision); err != nil {
```
zmberg, line 277: inplaceClassMetadataTransient is overly complex — just have it return a simple bool, true or false.

pkg/controller/sandbox/core/common_inplace_update_handler.go
```diff
@@ -20,237 +20,340 @@ import (
...
+}
+
+// inplaceUpdateStepResult describes a non-failure outcome of one engine pass.
+type inplaceUpdateStepResult int
+
+const (
+	// inplaceUpdateStepInProgress: a delivered round is still being applied by
+	// the kubelet (or a previous round must finish before a new one may
+	// start). The pod was not mutated in this pass.
+	inplaceUpdateStepInProgress inplaceUpdateStepResult = iota
```
zmberg, line 140: As far as I can tell, only inplaceUpdateStepInProgress and inplaceUpdateStepSucceeded serve a real purpose here — I honestly don't see the point of the rest.

pkg/controller/sandbox/core/common_inplace_update_handler.go
```diff
@@ -269,24 +372,9 @@ func isMetadataOnlyChange(pod *corev1.Pod, box *agentsv1alpha1.Sandbox) bool {
 		if origin.Image != container.Image {
 			return false
 		}
-		if !inplaceupdate.IsResourceSatisfied(origin.Resources, container.Resources) {
+		if !inplaceupdate.ResourcesExactlyEqual(origin.Resources, container.Resources) {
```
zmberg, line 375: ResourcesExactlyEqual -> IsResourceSatisfied

pkg/controller/sandbox/core/common_inplace_update_handler.go
Liquorice-Ma: done

PR #727, opened 2026-08-04
fix(sandboxset): only emit success event when dead sandbox deletion succeeds

### LINKED ISSUE #722: SandboxSet: deleteDeadSandboxes emits Normal "deleted" event even when sandbox deletion fails
### What happened

`deleteDeadSandboxes` in `pkg/controller/sandboxset/sandboxset_controller.go` records a `Normal` `"Sandbox X deleted"` event even when `r.Delete` returns an error. The success log and event sit outside the error branch with no `else`, so they fire on every iteration:

```go
if err := r.Delete(ctx, sbx); err != nil {
    log.Error(err, "failed to delete sandbox")
    failNum++
}
// these two lines run even when Delete failed:
log.Info("sandbox deleted", "sandbox", klog.KObj(sbx))
r.Recorder.Eventf(sbx, corev1.EventTypeNormal, EventFailedSandboxDeleted, "Sandbox %s deleted", klog.KObj(sbx))
```

The failure is only logged at debug verbosity (`V(DebugLogLevel)`), so at default log levels the misleading `Normal` event is the only user-visible signal for a failed deletion.

---

### What you expected to happen

- A failed deletion emits a `Warning` event that includes the error.
- The `Normal` `"deleted"` event is emitted only on success.
- The two paths are mutually exclusive.

---

### How to reproduce it

1. Create a `SandboxSet` and let a sandbox reach the `Dead` state (e.g., `SandboxFailed` or `SandboxSucceeded` phase).
2. Block the `Delete` call — e.g., via a validating webhook that denies deletion on `Sandbox` objects.
3. Run `kubectl get events -n <namespace> --field-selector reason=FailedSandboxDeleted`.
4. Observe a `Normal` `"Sandbox X deleted"` event even though the sandbox still exists and the delete failed.

---

### Anything else we need to know?

The reconcile loop is not broken — the error is still aggregated and returned (lines 435–436), so `controller-runtime` requeues correctly. This is purely an observability-correctness issue: success and failure currently look identical to the operator.

**Design note:** I plan to keep the existing `FailedSandboxDeleted` reason for both the `Warning` (failure) and `Normal` (success) events, since changing the reason string would alter user-visible output that dashboards/alerts may already match on. If you'd prefer a distinct success reason, I'm happy to do that instead. 

The fix itself is small — an `else` branch plus a `Warning` event with error detail, mirroring `createSandbox` (lines 376, 387). I can send a PR; does this direction look right, or would you rather handle events differently?

---

### Environment

- **OpenKruiseAgent version:** `master` (3e10d58)
- **Kubernetes version:** N/A (reproducible in any)
- **Install details:** N/A
- **Others:** —

AdeshDeshmukh:

`deleteDeadSandboxes` in `pkg/controller/sandboxset/sandboxset_controller.go` recorded a `Normal` `"Sandbox X deleted"` event and a success log even when `r.Delete` returned an error, because those statements sat outside the error branch with no `else`.

This PR:

1. Wraps the success log and event in an `else` branch so the two paths are mutually exclusive.
2. Records a `Warning` event with the error detail on failure, matching the existing `createSandbox` failure pattern (lines 376, 387).
3. Reuses the existing `FailedSandboxDeleted` reason for both severities to avoid changing user-visible event reasons that dashboards/alerts may match on.

---

Fixes #722

---

```bash
go test ./pkg/controller/sandboxset/ -run 'TestDeleteDeadSandboxes_DeleteFailure|TestReconcile_DeleteDead|TestReconcile_BasicScale' -count=1
go test ./pkg/controller/sandboxset/ -count=1
go vet ./pkg/controller/sandboxset/...
```

---

- Production change is ~6 lines in one function; no new constants or imports.
- New `TestDeleteDeadSandboxes_DeleteFailure` injects a `Delete` error via `interceptor.Funcs` and asserts a `Warning` event is emitted with no `Normal` event on the failure path.
- Existing GC-event tests are unchanged.

### DIFF
--- a/pkg/controller/sandboxset/sandboxset_controller.go
+++ b/pkg/controller/sandboxset/sandboxset_controller.go
@@ -426,11 +426,17 @@ func (r *Reconciler) deleteDeadSandboxes(ctx context.Context, dead []*agentsv1al
 			continue
 		}
 		if err := r.Delete(ctx, sbx); err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
 			log.Error(err, "failed to delete sandbox")
+			r.Recorder.Eventf(sbx, corev1.EventTypeWarning, EventFailedSandboxDeleted,
+				"Failed to delete sandbox %s: %v", klog.KObj(sbx), err)
 			failNum++
+		} else {
+			log.Info("sandbox deleted", "sandbox", klog.KObj(sbx))
+			r.Recorder.Eventf(sbx, corev1.EventTypeNormal, EventFailedSandboxDeleted, "Sandbox %s deleted", klog.KObj(sbx))
 		}
-		log.Info("sandbox deleted", "sandbox", klog.KObj(sbx))
-		r.Recorder.Eventf(sbx, corev1.EventTypeNormal, EventFailedSandboxDeleted, "Sandbox %s deleted", klog.KObj(sbx))
 	}
 	if failNum > 0 {
 		return fmt.Errorf("failed to delete %d sandboxes", failNum)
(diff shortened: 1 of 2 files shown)


AdeshDeshmukh:
@furykerry PTAL!

PR #728, opened 2026-08-05
fix(sandbox-manager): issue traffic tokens for cloned sandboxes

chengzhycn:

- Applies explicit three-state JWT auth merging during clone: an explicitly provided request value overrides the checkpoint value, while an absent request inherits the checkpoint unchanged.
- Issues a fresh traffic access token for the cloned Sandbox identity, validates the token and RFC3339 expiration, and keeps both values only on the transient response wrapper.
- Fails and cleans up clones when token issuance or response validation fails, and records clone traffic-token latency.
- Adds table-driven merge coverage plus clone lifecycle and E2B response tests.

NONE

```bash
go test ./pkg/identity -count=1
go test ./pkg/sandbox-manager/infra/sandboxcr -count=1
go test ./pkg/sandbox-manager/infra -count=1
go test ./pkg/servers/e2b -run "^TestCloneSandbox$|^TestConvertToE2BSandbox" -count=1
```

- Traffic tokens and expirations are never persisted to Sandbox or Checkpoint metadata.
- This PR intentionally excludes token-validity configuration, refresh APIs, SDK refresh behavior, and Gateway/JWKS changes.

### DIFF
--- a/pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
@@ -86,17 +86,3 @@ func RestoreAnnotationsFromCheckpoint(cp *v1alpha1.Checkpoint, sbx *v1alpha1.San
 	}
 	sbx.SetAnnotations(copyPreservedAnnotations(cpAnnotations, sbx.GetAnnotations()))
 }
-
-// restoreAnnotationsFromCheckpointForClone applies generic checkpoint
-// restoration plus clone request precedence for JWT authentication. An
-// explicitly provided request value overrides the restored checkpoint value;
-// an absent request inherits the checkpoint unchanged.
-func restoreAnnotationsFromCheckpointForClone(cp *v1alpha1.Checkpoint, sbx *v1alpha1.Sandbox) {
-	requestValue, requestProvided := sbx.GetAnnotations()[identity.AnnotationEnableJwtAuth]
-	RestoreAnnotationsFromCheckpoint(cp, sbx)
-	if requestProvided {
-		annotations := sbx.GetAnnotations()
-		annotations[identity.AnnotationEnableJwtAuth] = requestValue
-		sbx.SetAnnotations(annotations)
-	}
-}
--- a/pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
@@ -17,19 +17,10 @@ limitations under the License.
 package sandboxcr
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/openkruise/agents/api/v1alpha1"
 	"github.com/openkruise/agents/pkg/identity"
-)
-
-type jwtAuthSetting int
-
-const (
-	jwtAuthUnset jwtAuthSetting = iota
-	jwtAuthDisabled
-	jwtAuthEnabled
 )
 
 // necessaryAnnotationKeys defines the exact annotation keys that need to be
@@ -97,41 +88,15 @@ func RestoreAnnotationsFromCheckpoint(cp *v1alpha1.Checkpoint, sbx *v1alpha1.San
 }
 
 // restoreAnnotationsFromCheckpointForClone applies generic checkpoint
-// restoration plus the clone-specific JWT inheritance rule: an explicit
-// request may enable inherited JWT auth, but cannot disable it.
-func restoreAnnotationsFromCheckpointForClone(cp *v1alpha1.Checkpoint, sbx *v1alpha1.Sandbox) error {
-	checkpointSetting := parseJWTAuthSetting(cp.GetAnnotations())
-	requestSetting := parseJWTAuthSetting(sbx.GetAnnotations())
-
-	if checkpointSetting == jwtAuthEnabled && requestSetting == jwtAuthDisabled {
-		return fmt.Errorf("cannot disable JWT authentication inherited from checkpoint")
-	}
-
+// restoration plus clone request precedence for JWT authentication. An
+// explicitly provided request value overrides the restored checkpoint value;
+// an absent request inherits the checkpoint unchanged.
+func restoreAnnotationsFromCheckpointForClone(cp *v1alpha1.Checkpoint, sbx *v1alpha1.Sandbox) {
+	requestValue, requestProvided := sbx.GetAnnotations()[identity.AnnotationEnableJwtAuth]
 	RestoreAnnotationsFromCheckpoint(cp, sbx)
-	finalSetting := checkpointSetting
-	if requestSetting == jwtAuthEnabled || checkpointSetting == jwtAuthUnset {
-		finalSetting = requestSetting
+	if requestProvided {
+		annotations := sbx.GetAnnotations()
+		annotations[identity.AnnotationEnableJwtAuth] = requestValue
+		sbx.SetAnnotations(annotations)
 	}
-	annotations := sbx.GetAnnotations()
-	switch finalSetting {
-	case jwtAuthEnabled:
-		annotations[identity.AnnotationEnableJwtAuth] = v1alpha1.True
-	case jwtAuthDisabled:
-		annotations[identity.AnnotationEnableJwtAuth] = v1alpha1.False
-	default:
-		delete(annotations, identity.AnnotationEnableJwtAuth)
-	}
-	sbx.SetAnnotations(annotations)
-	return nil
-}
-
-func parseJWTAuthSetting(annotations map[string]string) jwtAuthSetting {
-	value, present := annotations[identity.AnnotationEnableJwtAuth]
-	if !present {
-		return jwtAuthUnset
-	}
-	if value == v1alpha1.True {
-		return jwtAuthEnabled
-	}
-	return jwtAuthDisabled
 }
--- a/pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
@@ -17,10 +17,19 @@ limitations under the License.
 package sandboxcr
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/openkruise/agents/api/v1alpha1"
 	"github.com/openkruise/agents/pkg/identity"
+)
+
+type jwtAuthSetting int
+
+const (
+	jwtAuthUnset jwtAuthSetting = iota
+	jwtAuthDisabled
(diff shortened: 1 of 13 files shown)


pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
```diff
@@ -86,3 +95,53 @@ func RestoreAnnotationsFromCheckpoint(cp *v1alpha1.Checkpoint, sbx *v1alpha1.San
...
+	case jwtAuthDisabled:
+		annotations[identity.AnnotationEnableJwtAuth] = v1alpha1.False
+	default:
+		delete(annotations, identity.AnnotationEnableJwtAuth)
+	}
+	sbx.SetAnnotations(annotations)
+	return nil
+}
+
+func parseJWTAuthSetting(annotations map[string]string) (jwtAuthSetting, error) {
```
furykerry: the contract documented on AnnotationEnableJwtAuth (pkg/identity/types.go:60-15, which this PR itself edits) states that any value other than exactly "true" — including "1" or "True" — simply leaves the sandbox out of the issuance path. The claim path honors that (silent no-op). The new clone path instead treats any value other than "true"/"false" as a hard error. 

consider following the claim path and treat invalid value as false

pkg/sandbox-manager/infra/sandboxcr/clone.go
```diff
@@ -191,6 +191,25 @@ func CloneSandbox(ctx context.Context, opts infra.CloneSandboxOptions, cache inf
...
 
+	// Mint a new traffic access token for the cloned sandbox. The token is bound
+	// to the clone's identity and stays only on the transient wrapper returned to
+	// the API layer; it is never persisted to the Sandbox or Checkpoint.
+	if identity.IsAccessTokenRequested(sbx.Sandbox) {
+		start := time.Now()
+		accessResp, issueErr := identity.IssueSandboxAccessToken(ctx, sbx.Sandbox)
+		metrics.TrafficToken = time.Since(start)
+		metrics.Total += metrics.TrafficToken
+		if issueErr != nil {
```
furykerry, line 203: the security-token block six lines above deliberately preserves context cancellation (if !wait.Interrupted(err)), but the new access-token block unconditionally wraps any issue error as retriableError. 

consider mirror the sibling pattern so as to avoid  context retry if canceled/expired inside IssueSandboxAccessToken

pkg/sandbox-manager/infra/sandboxcr/clone.go
```diff
@@ -191,6 +191,25 @@ func CloneSandbox(ctx context.Context, opts infra.CloneSandboxOptions, cache inf
...
+			err = retriableError{Message: issueErr.Error()}
+			return
+		}
+		if validationErr := validateTrafficTokenResponse(accessResp); validationErr != nil {
```
furykerry: Issue: Clone now rejects nil/empty-token/invalid-expiration responses, but claim (claim.go:380-382) still accepts any non-nil response silently — an empty AccessToken would be returned to the client as a "successful" claim. The identical provider misbehavior thus fails clones but succeeds claims. The validation also logically belongs to the identity contract, not the clone flow.

Recommendation: Move the validation next to the issuance helper (e.g. into pkg/identity, called from or exported alongside IssueSandboxAccessToken) and apply it on both paths, so the two issuance sites stay behaviorally identical — this PR's stated goal ("mirroring the claim flow ordering").

pkg/sandbox-manager/infra/sandboxcr/clone.go
```diff
@@ -191,6 +191,25 @@ func CloneSandbox(ctx context.Context, opts infra.CloneSandboxOptions, cache inf
 		metrics.Total += metrics.SecurityToken
 	}
 
+	// Mint a new traffic access token for the cloned sandbox. The token is bound
+	// to the clone's identity and stays only on the transient wrapper returned to
+	// the API layer; it is never persisted to the Sandbox or Checkpoint.
+	if identity.IsAccessTokenRequested(sbx.Sandbox) {
```
furykerry, line 198: consider renumber the step as step 7

pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
```diff
@@ -86,3 +95,53 @@ func RestoreAnnotationsFromCheckpoint(cp *v1alpha1.Checkpoint, sbx *v1alpha1.San
...
+	requestSetting, err := parseJWTAuthSetting(sbx.GetAnnotations())
+	if err != nil {
+		return fmt.Errorf("invalid clone JWT authentication setting: %w", err)
+	}
+
+	if checkpointSetting == jwtAuthEnabled && requestSetting == jwtAuthDisabled {
+		return fmt.Errorf("cannot disable JWT authentication inherited from checkpoint")
+	}
+
+	RestoreAnnotationsFromCheckpoint(cp, sbx)
```
furykerry: consider just add AnnotationEnableJwtAuth to necessaryAnnotationKeys, and simply using RestoreAnnotationsFromCheckpoint instead. customized func restoreAnnotationsFromCheckpointForClone is not required

pkg/sandbox-manager/infra/sandboxcr/clone.go
chengzhycn: Addressed in 0415b662. The access-token branch now mirrors the sibling security-token path and preserves errors recognized by wait.Interrupted instead of wrapping cancellation/deadline errors as retriableError. Added a context.Canceled regression case.

pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
chengzhycn: Addressed in 0415b662. A present value now maps to enabled only when it is exactly "true"; every other present value maps to explicit disabled, while absence remains unset for clone inheritance. Added matrix coverage for non-standard request/checkpoint values, including rejection when such a request would weaken an enabled checkpoint.

pkg/sandbox-manager/infra/sandboxcr/clone.go
chengzhycn: Addressed in 0415b662. Traffic access-token issuance is now Step 7 and CSI mount is Step 8.

pkg/sandbox-manager/infra/sandboxcr/clone.go
chengzhycn: Addressed in 0415b662. Response validation now lives in identity.IssueSandboxAccessToken, so claim and clone both reject nil responses, empty tokens, and non-RFC3339 expirations through the shared identity contract. Added table-driven identity tests and removed clone-local validation.

pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
chengzhycn: `AnnotationEnableJwtAuth` is already preserved by `necessaryAnnotationKeyPrefixes`: its value is `security.agents.kruise.io/enable-jwt-auth`, which matches `identity.SecurityMetadataPrefix` (`security.agents.kruise.io/`). Adding the same key to `necessaryAnnotationKeys` would therefore not change whether it is copied.

The issue here is merge precedence rather than selection. `newSandboxFromTemplate` runs the request modifier first, so the new Sandbox already contains the Clone request metadata. `RestoreAnnotationsFromCheckpoint` then calls `copyPreservedAnnotations(cpAnnotations, sbxAnnotations)`, whose `dst[key] = val` assignment unconditionally lets the Checkpoint overwrite a same-key request value. That would make a checkpoint value of `false` silently override an explicit request value of `true`, preventing a security upgrade; conversely, checkpoint `true` would silently ignore request `false` instead of returning the intended downgrade error.

The clone-specific wrapper captures the request/checkpoint three-state values before generic restoration, restores all other annotations through the existing helper, then applies the explicit policy: inherit when unset, allow `false -> true`, and reject `true -> false`. Keeping this policy outside the generic Restore also avoids making that mechanical copy helper interpret every pre-existing Sandbox annotation as Clone request intent.

chengzhycn: Follow-up in 4bd23a38: the JWT merge policy now matches the CSI mount precedence model. An explicitly provided Clone request value always wins, including `false` overriding checkpoint `true`; an absent request inherits the checkpoint unchanged.

The helper has been reduced to capture the request value/presence, call the generic Restore, and reapply the request value when present. We still cannot use only `RestoreAnnotationsFromCheckpoint`, because its source-over-destination assignment would make the checkpoint overwrite request metadata. Adding the exact key is also still redundant because the security prefix already selects it.

pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
chengzhycn: Follow-up in 4bd23a38: after aligning JWT merge semantics with CSI mount precedence, any explicitly provided request value is now preserved verbatim and overrides the checkpoint. Only exact `"true"` enables issuance through `IsAccessTokenRequested`; other explicit values remain non-enabling. An absent request still inherits the checkpoint unchanged. The merge matrix and E2B test were updated accordingly.

pkg/sandbox-manager/infra/sandboxcr/checkpoint_utils.go
```diff
@@ -86,3 +86,17 @@ func RestoreAnnotationsFromCheckpoint(cp *v1alpha1.Checkpoint, sbx *v1alpha1.San
...
 }
+
+// restoreAnnotationsFromCheckpointForClone applies generic checkpoint
+// restoration plus clone request precedence for JWT authentication. An
+// explicitly provided request value overrides the restored checkpoint value;
+// an absent request inherits the checkpoint unchanged.
+func restoreAnnotationsFromCheckpointForClone(cp *v1alpha1.Checkpoint, sbx *v1alpha1.Sandbox) {
+	requestValue, requestProvided := sbx.GetAnnotations()[identity.AnnotationEnableJwtAuth]
+	RestoreAnnotationsFromCheckpoint(cp, sbx)
+	if requestProvided {
```
furykerry: consider move the annotation overriding logic to prepareSandboxFromCheckpoint, along with AnnotationCSIVolumeConfig overriding logic

chengzhycn: Addressed in e9872e00. `prepareSandboxFromCheckpoint` now captures the request JWT value before generic checkpoint restoration and reapplies it immediately afterward, next to the existing CSI request override. The clone-specific restore helper was removed from `checkpoint_utils.go`; behavior and merge-matrix coverage remain unchanged.

PR #734, opened 2026-08-05
    feat(identity): deliver security tokens over the resolved runtime transport

BH4AWS:
…nsport

The claim, clone and post-resume flows already resolved a per-sandbox transport with TransportOptionsFor and applied it to the /init handshake and the CSI mounts, but the security-token step stayed on the plaintext runtime URL, so a TLS-capable sandbox received its identity-provider credential in clear text. Forwarding the transport alone would not have helped: only the JSON capability group understood WithTLS, while the multipart /files route and the connect-gRPC Filesystem and Process services resolved GetRuntimeURL directly, which is why Initialize still carried a TODO for this step.

- runtime: add Filesystem() (Write/ListDir/Remove) and Process() (Run/Chmod) capability groups to the Runtime interface and move the existing helper implementations into them. All three wire protocols (JSON, multipart, connect-gRPC) now resolve their endpoint and HTTP client through a single resolveTransport, so no path can drift back to plaintext unnoticed. TLS keeps HTTP/1.1, matching the protocol the plaintext connect path already uses. The former free functions remain as thin wrappers that accept rtOpts, leaving their call sites and the recycle test seam untouched.
- identity: thread rtOpts through SecurityTokenPropagator, IdentityProvider.PropagateSecurityToken, PropagateSandboxToken and ProcessSandboxToken. The package only carries the value and never resolves or overrides it, so the TLS decision stays at the caller's boundary and identity remains transport-neutral.
- claim / clone / post-resume: pass the already resolved rtOpts into ProcessSandboxToken, which retires the TODO in Initialize.
- securitytokenrefresh: plumb the runtime TLS bundle from Deps into NewDefaultRefresher and resolve the transport per refresh, because the Pod IP and the advertised capability change across pause/resume cycles. The resolution runs before issuance, so a sandbox advertising TLS while the controller holds no bundle fails without burning an issuance call rather than downgrading the delivery to plaintext.
- WriteFileArgs.Permissions: correct the comment, which claimed the mode travels in an X-File-Mode header that is never set. Behaviour is unchanged; ChmodFileOnRuntime stays the way to enforce a mode until the runtime honours the header natively.

### DIFF
--- a/pkg/utils/runtime/transport_test.go
+++ b/pkg/utils/runtime/transport_test.go
@@ -32,6 +32,7 @@ import (
 	"testing"
 	"time"
 
+	"connectrpc.com/connect"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
@@ -41,6 +42,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
+	"github.com/openkruise/agents/proto/envd/filesystem/filesystemconnect"
+	"github.com/openkruise/agents/proto/envd/process"
+	"github.com/openkruise/agents/proto/envd/process/processconnect"
 )
 
 // sandboxWithPodIP returns a sandbox whose Pod IP is set (but with no runtime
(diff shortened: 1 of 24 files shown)


pkg/utils/runtime/transport_test.go
```diff
@@ -458,3 +462,204 @@ func TestNewTLSBundleFromSecret(t *testing.T) {
...
+		{
+			name: "process run",
+			invoke: func(ctx context.Context) error {
+				_, err := rt.Process().Run(ctx, RunCommandRequest{
+					ProcessConfig: &process.ProcessConfig{Cmd: "true"}, Timeout: time.Second,
+				})
+				return err
+			},
+		},
+	}
```
furykerry, line 674: plz add  Remove and Chmod.

furykerry:
Issue: The PR description says "no path can drift back to plaintext unnoticed" — true within the runtime client, but one in-repo callers still pass no options: ExecuteLifecycleHook (upgrade lifecycle hooks, which also hard-gates on GetRuntimeURL) ， consider add an TODOI in ExecuteLifecyleHook

```
// TODO: resolve the sandbox transport via TransportOptionsFor once the
// upgrade/recycle flows carry the runtime TLS bundle, mirroring the
// claim/clone/post-resume flows (see PR #734).
```

pkg/utils/runtime/transport_test.go
BH4AWS: done

PR #740, opened 2026-08-06
fix(sandbox-manager): correct E2B traffic policy precedence

chengzhycn:

- Changes per-Sandbox E2B TrafficPolicy priority from 1000 to 100.
- Ensures per-Sandbox allow/deny rules are evaluated before matching E2B global fallback policies.
- Adds coverage for newly created policies and runtime updates that migrate a legacy priority-1000 policy to priority 100.

NONE

Run:

```bash
go test ./pkg/sandbox-manager/infra/sandboxcr -count=1
go test ./pkg/servers/e2b -run TestUpdateSandboxNetwork_Success -count=1
```

Verify that newly created TrafficPolicies use priority 100 and updating an existing priority-1000 policy changes it to priority 100 without changing its rules.

This quick fix intentionally changes only the per-Sandbox priority. E2B global fallback policies are deployment prerequisites and must use a priority greater than 100. Their concrete priority is operator-defined; 900 is a recommended value, not a code dependency.

For narrow-deny semantics, an internet-enabled sandbox is expected to match a global allow fallback. For internet-disabled semantics, it is expected to match a global deny fallback. Runtime mutation of `allow_internet_access` remains outside this PR.

### DIFF
--- a/pkg/sandbox-manager/infra/sandboxcr/network.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/network.go
@@ -30,6 +30,10 @@ import (
 	"github.com/openkruise/agents/pkg/utils/network"
 )
 
+// E2B global fallback policies must use a priority greater than this value.
+// TrafficPolicy priorities are evaluated in ascending order.
+const e2bPerSandboxTrafficPolicyPriority int32 = 100
+
 // sandboxOwnerRef returns an OwnerReference that points to the given Sandbox CR.
 // Setting this on TrafficPolicy CRs ensures they are garbage-collected
 // when the owning Sandbox is deleted (including timeout-driven deletion by the controller).
@@ -102,7 +106,7 @@ func buildTrafficPolicy(allowOutCIDRs, allowOutDomains, denyOut []string, namesp
 			OwnerReferences: []metav1.OwnerReference{sandboxOwnerRef(sandbox)},
 		},
 		Spec: agentsv1alpha1.TrafficPolicySpec{
-			Priority: 1000,
+			Priority: e2bPerSandboxTrafficPolicyPriority,
 			Selector: metav1.LabelSelector{
 				MatchLabels: map[string]string{
 					agentsv1alpha1.LabelSandboxName: sandbox.Name,
--- a/pkg/sandbox-manager/infra/sandboxcr/network_test.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/network_test.go
@@ -235,7 +235,7 @@ func TestBuildTrafficPolicy(t *testing.T) {
 			assert.Equal(t, "tp-", tp.GenerateName)
 			assert.Equal(t, "default", tp.Namespace)
 			assert.Equal(t, "test-sandbox", tp.Spec.Selector.MatchLabels[agentsv1alpha1.LabelSandboxName])
-			assert.Equal(t, int32(1000), tp.Spec.Priority)
+			assert.Equal(t, e2bPerSandboxTrafficPolicyPriority, tp.Spec.Priority)
 			// Verify OwnerReference is set
 			require.Len(t, tp.OwnerReferences, 1)
 			assert.Equal(t, "Sandbox", tp.OwnerReferences[0].Kind)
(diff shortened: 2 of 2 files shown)


pkg/sandbox-manager/infra/sandboxcr/network.go
```diff
@@ -30,6 +30,10 @@ import (
 	"github.com/openkruise/agents/pkg/utils/network"
 )
 
+const (
+	e2bPerSandboxTrafficPolicyPriority int32 = 100
```
furykerry: The policy composition assumption is undocumented.   consider add a short comment ,e,g. 
```

 // e2bPerSandboxTrafficPolicyPriority must stay below the priority of
  // operator-managed GlobalTrafficPolicy guardrails (typically the CRD
  // default 1000): rules are evaluated priority ASC, so per-sandbox
  // allow/deny rules decide before the global fallback applies.
```

pkg/servers/e2b/network_test.go
```diff
@@ -541,41 +541,48 @@ func TestUpdateSandboxNetwork_Success(t *testing.T) {
...
+			expectCode:     http.StatusNoContent,
+			expectTP:       true,
+			expectInternet: agentsv1alpha1.False,
 		},
 		{
-			name:       "update with empty config clears TP",
-			req:        models.SandboxNetworkUpdateConfig{},
-			expectCode: http.StatusNoContent,
-			expectTP:   false,
+			name: "update with allowInternetAccess true clears TP",
```
furykerry: Issue: The previous "update with empty config clears TP" case (empty SandboxNetworkUpdateConfig{}) was replaced by the allowInternetAccess: true case instead of keeping both. The empty-config path — nil internet flag leaves the label untouched while the TP is deleted — is now uncovered at the API level.

Recommendation: Restore the empty-config case alongside the new one, asserting expectTP: false and that the label is unchanged from its initial value.

furykerry (commented):
## Review: does `updateInternetAccessLabel` accidentally trigger sandbox rebuild/upgrade?

I traced the controller-side reaction to the two writes this function performs
(`sandbox.Labels` and `sandbox.Spec.Template.Labels`).

### TL;DR

For ordinary E2B sandboxes: **no rebuild is triggered** — the change takes the
designed metadata-only pod patch path. However, there is **one combination that
does trigger a full pod recreation**, plus two silent edge cases where the label
lands on the CR but not on the pod.

### Why the normal path is safe

Writing `Spec.Template.Labels` always changes `Status.UpdateRevision`
(`core.HashSandbox` hashes the whole template including labels), so the
controller does notice the revision mismatch. But:

1. The immutable-hash guard does not trip: `HashSandbox` computes
   `hashImmutablePart` after zeroing labels/annotations/image/resources, so a
   label-only change is permitted in-place
   (`pkg/controller/sandbox/core/util.go`, `common_inplace_update_handler.go`).
2. The recreate-upgrade gate stays closed: entering `SandboxUpgrading` requires
   both a revision mismatch and `RequiresPodReplacementUpgrade(box)`, i.e.
   `Spec.UpgradePolicy` of type Recreate/CheckpointRestore. The E2B path never
   sets `UpgradePolicy` and no webhook defaults it.
3. `isMetadataOnlyChange(pod, box)` returns true (image/resources unchanged),
   so `handleInPlaceUpdateCommon` patches pod metadata directly and bumps
   `pod-template-hash` without setting `InplaceUpdate`/`Ready=False` and
   without deleting the pod. This is the mechanism `create.go` already relies
   on ("the controller patches the pod metadata directly for metadata-only
   changes"), covered by `TestHandleInPlaceUpdateCommon_MetadataOnlyChange`.
4. Pool recycle is unaffected: the recycle staleness check compares
   `box.Labels[LabelTemplateHash]`, which this function never touches.

### ⚠️ The case that DOES trigger recreation

`Spec.UpgradePolicy` is set by SandboxUpdateOps (`sandboxupdateops/patch.go`)
and is never cleared afterwards. For any sandbox that has ever gone through a
SandboxUpdateOps batch upgrade:

```
network update API → updateInternetAccessLabel → template label change
→ UpdateRevision changes → RequiresPodReplacementUpgrade == true
→ phase → SandboxUpgrading → performRecreateUpgrade deletes & recreates the pod
```

(With CheckpointRestore policy it would even checkpoint first.) So a user
merely toggling `allow_internet_access` on such a sandbox would destroy and
rebuild its pod. This latent trap predates this PR, but this PR introduces the
first routine user-facing API that mutates template labels, making the trap
reachable. Possible mitigations: clear `UpgradePolicy` once an upgrade
completes, or gate the recreate decision on the immutable-part hash so
metadata-only diffs never enter the recreate path. At minimum this interaction
should be documented.

### Silent edge cases (no rebuild, but inconsistent state)

1. **Previously in-place-upgraded pods**: if the pod already carries an
   in-place update state, `handleInPlaceUpdateCommon` short-circuits with
   "multiple in-place updates are not supported" *before* the metadata-only
   check — the new label lands on the Sandbox CR but is never patched onto the
   pod, so the GlobalTrafficPolicy selector keeps enforcing the old mode until
   the pod is next recreated.
2. **Legacy pods without the `pod-template-hash` label**: same outcome — CR
   updated, pod untouched.

pkg/sandbox-manager/infra/sandboxcr/network.go
chengzhycn: Added a concise invariant above the constant: E2B global fallback policies must use a priority greater than 100 because priorities are evaluated in ascending order. I intentionally avoided naming a concrete fallback priority because 900 is recommended but not required.

pkg/servers/e2b/network_test.go
chengzhycn: Restored the original empty-config API test. The allowInternetAccess runtime label update has been removed from this quick fix, so the API test is back to covering TrafficPolicy deletion without changing Sandbox labels.

chengzhycn:
Addressed the review by narrowing this PR to the priority quick fix only. The `allow_internet_access` runtime label/PodTemplate mutation and its controller interactions have been removed, so the rebuild and stale-Pod-label risks described in the review are no longer introduced by this PR. The final diff only changes per-Sandbox TrafficPolicy priority from 1000 to 100 and tests create/update behavior.

PR #742, opened 2026-08-06
feat: rotate traffic access tokens

chengzhycn:

- Makes TrafficAccessToken validity a sandbox-manager policy and passes the requested lifetime through identity providers.
- Adds Infra issuance, Manager refresh/bootstrap orchestration, per-Sandbox in-flight result coalescing, and refresh interval limiting.
- Adds a Kruise-specific E2B refresh endpoint, plus fresh token bootstrap on Connect.
- Adds an independent, opt-in Traffic JWT monkey patch for sync/async automatic refresh across envd HTTP/RPC and Jupyter requests.
- Preserves the legacy, approximately 100-year binary default validity while allowing operators to configure short-lived tokens after clients support refresh.

NONE

- Run `go test ./pkg/identity ./pkg/sandbox-manager/... ./pkg/servers/e2b ./cmd/sandbox-manager -count=1`.
- Run `go test -race ./pkg/sandbox-manager -run 'TestSandboxManagerTrafficAccessTokenLimitIsPerSandbox|TestSandboxManagerBootstrapTrafficAccessTokenJoinsRefresh|TestSandboxManagerRefreshTrafficAccessTokenRateLimit|TestTrafficTokenLimiter' -count=3`.
- Run the Python test suite from `sdk/customized_e2b`; it covers 23 sync/async refresh and transport cases.
- Run the `sandbox-gateway (JWT auth)` workflow to verify gateway JWT validation and SDK data-plane transport coverage. The Manager-driven rotation case is explicitly skipped until the open-source Manager has production issuer wiring.

- Refresh issues a replacement token but does not immediately revoke the previous token; overlap ends at the previous JWT expiration.
- Concurrent issuance for the same Sandbox waits for and reuses one result; different Sandboxes issue independently.
- `patch_e2b()` remains the existing URL/protocol patch. Automatic rotation requires the separate `patch_traffic_access_token()` opt-in patch.
- Automatic refresh supports `e2b>=2.35.0,<2.38.0` with `e2b-code-interpreter>=2.9.0,<2.10.0`; older SDKs retain the base patch.
- The open-source Manager currently has no production-configurable signed JWT issuer. This PR does not compile a test provider into the production entrypoint, and its Manager-driven rotation E2E remains skipped until that wiring exists.
- Deploy refresh-capable clients before configuring a shorter `--traffic-access-token-validity`; clients without refresh lose data-plane access when a short-lived token expires.


.github/workflows/e2e-sandbox-gateway-security.yaml
```diff
@@ -262,11 +281,14 @@ jobs:
...
             export TRAFFIC_ACCESS_TOKEN_JWT_E2E=true
             export JWT_E2E_TOKEN_COMMAND="kubectl exec -n sandbox-system deployment/jwt-e2e-oidc-provider -- /jwt-oidc-provider issue"
+            export JWT_E2E_TOKEN_VALIDITY_SECONDS=65
           fi
           # Temporarily pinned: e2b 2.38.0 removed the http2 kwarg of
           # get_transport() that e2b-code-interpreter 2.9.0 still passes.
           # 2.37.1 + 2.9.0 is the last verified working combination.
           bash hack/run-e2b-e2e-test.sh \
+            --e2b-version 2.37.0 \
+            --sdk-version 2.9.0 \
```
furykerry: why pass --e2b-version and --sdk-version twice? it is already in L293 and L294

pkg/servers/e2b/pause_resume.go
```diff
@@ -318,10 +319,27 @@ func (sc *Controller) ConnectSandbox(r *http.Request) (web.ApiResponse[*models.S
 		return web.ApiResponse[*models.Sandbox]{}, withSandboxResourceContext(err, sbx)
 	}
 	log.Info("sandbox timeout updated")
+	user := GetUserFromContext(ctx)
```
furykerry: Issue: user := GetUserFromContext(ctx); if user == nil { …401… } happens after resume and timeout update already executed. With CheckApiKey upstream this is practically unreachable, but if it ever fired, the sandbox would be resumed and its timeout changed before returning 401.

Recommendation: Resolve and nil-check the user at the top of the handler, before any mutation

pkg/sandbox-manager/traffic_token.go
```diff
@@ -0,0 +1,205 @@
...
+		},
+	})
+	if issueErr != nil && managererrors.GetErrCode(issueErr) == managererrors.ErrorUnknown {
+		if errors.Is(issueErr, cache.ErrSandboxNotFound) {
+			issueErr = managererrors.NewError(managererrors.ErrorNotFound, "sandbox %s not found", opts.SandboxID)
+		} else {
+			issueErr = managererrors.NewError(managererrors.ErrorUnavailable, "failed to issue traffic access token: %v", issueErr)
+		}
+	}
+	m.trafficTokenLimiter.complete(limiterKey, flight, result, issueErr)
```
furykerry: if the code return early in L91 or L96, the limiter is not completed, causing possible limitation token leakage

pkg/servers/e2b/routes.go
```diff
@@ -59,6 +59,7 @@ func (sc *Controller) registerRoutes() {
 	RegisterE2BRoute(sc.mux, http.MethodPost, "/sandboxes/{sandboxID}/pause", sc.PauseSandbox, sc.CheckApiKey)
 	RegisterE2BRoute(sc.mux, http.MethodPost, "/sandboxes/{sandboxID}/resume", sc.ResumeSandbox, sc.CheckApiKey)
 	RegisterE2BRoute(sc.mux, http.MethodPost, "/sandboxes/{sandboxID}/connect", sc.ConnectSandbox, sc.CheckApiKey)
+	RegisterE2BRoute(sc.mux, http.MethodPost, "/sandboxes/{sandboxID}/traffic-access-token", sc.RefreshTrafficAccessToken, sc.CheckApiKey)
```
furykerry: consider only add route for kruise specific route, /kruise/api/

cmd/sandbox-manager/e2e_identity_provider.go
```diff
@@ -0,0 +1,32 @@
...
+package main
+
+import (
+	"os"
+
+	"github.com/openkruise/agents/pkg/identity"
+	e2eidentityprovider "github.com/openkruise/agents/test/e2b/identityprovider"
+)
+
+func init() {
```
furykerry: Production entrypoint imports the test tree, is it possible to move this init to the test package ?

sdk/customized_e2b/setup.py
```diff
@@ -19,20 +19,26 @@
...
-        "Programming Language :: Python :: 3.7",
-        "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
     ],
     python_requires=">=3.9,<4.0",
     install_requires=[
         "e2b>=2.8.0",
+        "e2b-code-interpreter>=2.4.1",
```
furykerry, line 29: Python version statement inconsistency: sdk/customized_e2b/README.md says the refresh patch requires Python 3.10+, while setup.py declares python_requires=">=3.9". Align them.

pkg/sandbox-manager/infra/sandboxcr/traffic_token.go
```diff
@@ -0,0 +1,99 @@
...
+	response, err := identity.IssueSandboxAccessToken(ctx, before, opts.TokenOptions)
+	if err != nil {
+		return infra.TrafficAccessToken{}, err
+	}
+
+	after, err := i.getTrafficTokenSandbox(ctx, key, opts.SandboxID)
+	if err != nil {
+		return infra.TrafficAccessToken{}, err
+	}
+	if before.UID != after.UID {
```
furykerry: it is not necessary to read before/after the token insurance, the id token already encode the sandbox uid information.

pkg/sandbox-manager/traffic_token.go
```diff
@@ -0,0 +1,205 @@
...
+	return result, err
+}
+
+// BootstrapTrafficAccessToken issues a token for connect responses when the
+// Sandbox requires traffic authentication. Unprotected Sandboxes are skipped.
+func (m *SandboxManager) BootstrapTrafficAccessToken(ctx context.Context, opts RefreshTrafficAccessTokenOptions) (infra.TrafficAccessToken, bool, error) {
+	return m.issueTrafficAccessToken(ctx, opts, true, false)
+}
+
+func (m *SandboxManager) issueTrafficAccessToken(ctx context.Context, opts RefreshTrafficAccessTokenOptions, allowUnprotected, enforceInterval bool) (infra.TrafficAccessToken, bool, error) {
```
furykerry: plz leave code comment about the return value, especially the second bool param

pkg/sandbox-manager/traffic_token.go
```diff
@@ -0,0 +1,205 @@
...
+// RefreshTrafficAccessToken issues a new token without mutating the Sandbox.
+func (m *SandboxManager) RefreshTrafficAccessToken(ctx context.Context, opts RefreshTrafficAccessTokenOptions) (infra.TrafficAccessToken, error) {
+	result, _, err := m.issueTrafficAccessToken(ctx, opts, false, true)
+	return result, err
+}
+
+// BootstrapTrafficAccessToken issues a token for connect responses when the
+// Sandbox requires traffic authentication. Unprotected Sandboxes are skipped.
+func (m *SandboxManager) BootstrapTrafficAccessToken(ctx context.Context, opts RefreshTrafficAccessTokenOptions) (infra.TrafficAccessToken, bool, error) {
+	return m.issueTrafficAccessToken(ctx, opts, true, false)
```
furykerry: plz comment why bootstrap does not require interval enforcing

pkg/sandbox-manager/config/manager.go
```diff
@@ -27,8 +28,28 @@ import (
 const (
 	// DefaultMemberlistBindPort is the default port for memberlist gossip
 	DefaultMemberlistBindPort = 7946
+	// DefaultTrafficAccessTokenValidity preserves the existing traffic-token
+	// lifetime while clients migrate to automatic refresh.
+	DefaultTrafficAccessTokenValidity = 100 * 365 * 24 * time.Hour
+	// DefaultTrafficAccessTokenMinValidity prevents unusably short tokens.
+	DefaultTrafficAccessTokenMinValidity = 5 * time.Minute
```
furykerry, line 35: DefaultTrafficAccessTokenRefreshMinInterval  is actually smaller than DefaultTrafficAccessTokenMinValidity, consider making the default values consistent

pkg/sandbox-manager/traffic_token.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (l *trafficTokenLimiter) complete(key string, flight *trafficTokenFlight, result infra.TrafficAccessToken, err error) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	flight.result = result
+	flight.err = err
+	entry := l.entries[key]
+	if entry.flight == flight {
+		entry.flight = nil
+		l.entries[key] = entry
```
furykerry: Issue: acquire stamps entry.last = now before the attempt, and complete never rolls it back on failure. A single transient issuer blip therefore locks the client out with 429 for the full RefreshMinInterval. With the E2E-style config (65s validity, 30s default interval in production) one failed refresh plus one 429 can consume most of the remaining token lifetime, and the SDK's own backoff compounds it.

Recommendation: Only charge the interval for successful issuance, letting the client's own backoff govern retries:

```
func (l *trafficTokenLimiter) complete(key string, flight *trafficTokenFlight, result infra.TrafficAccessToken, err error) {
    ...
    if entry.flight == flight {
        entry.flight = nil
        if err != nil {
            // A failed attempt must not spend the caller's refresh budget.
            entry.last = time.Time{}
        }
        l.entries[key] = entry
    }
```

.github/workflows/e2e-sandbox-gateway-security.yaml
chengzhycn: Removed the duplicate 2.37.0/2.9.0 arguments and kept the existing verified 2.37.1/2.9.0 pair in c56853b8.

pkg/sandbox-manager/traffic_token.go
chengzhycn: Added the rationale in c56853b8: bootstrap bypasses the completed-refresh interval so a recent explicit refresh cannot make Connect fail, while it still joins an issuance already in flight for the same Sandbox.

pkg/sandbox-manager/traffic_token.go
chengzhycn: Documented the second return value and named it issued in the public method signature. It is false when the Sandbox does not require traffic authentication. Fixed in c56853b8.

pkg/servers/e2b/routes.go
chengzhycn: Changed the refresh endpoint to register only at /kruise/api/sandboxes/{sandboxID}/traffic-access-token and added a route registration test in c56853b8.

pkg/servers/e2b/pause_resume.go
chengzhycn: Moved user resolution and the nil check to the start of ConnectSandbox, before request parsing and any Sandbox mutation, in c56853b8.

sdk/customized_e2b/setup.py
chengzhycn: `python_requires` applies to the whole distribution and cannot vary by extra. The base `patch_e2b` path still supports Python 3.9, while only the optional `[traffic-token-refresh]` feature requires Python 3.10. I also verified that installing the refresh extra on Python 3.9 already fails dependency resolution because `e2b>=2.35.0` itself requires Python >=3.10, so the feature constraint is enforced even though the distribution-wide baseline remains >=3.9. Would you prefer that we keep the current split requirement and document it per feature, or raise the entire `kruise-agents` package to Python >=3.10?

pkg/sandbox-manager/traffic_token.go
chengzhycn: These returns are in the follower path, so they must not call `complete`: the leader owns the flight and always publishes the result, clears the entry, and closes `done` after issuance returns. A canceled follower only stops its own wait, and a follower receiving from `done` does so after leader completion; completing from either follower path could close the shared channel twice and panic. Added an explanatory ownership comment in e6f1798e.

cmd/sandbox-manager/e2e_identity_provider.go
chengzhycn: Agreed. The open-source sandbox-manager does not yet have a production-configurable signed JWT issuer implementation, so compiling a test-only provider into the Manager entrypoint would hide that product limitation and create the wrong dependency direction. I removed `e2e_identity_provider.go`, the test Provider package, the build-tag/Docker wiring, and the Manager issuer configuration in 79e01bdc. The Manager-driven rotation E2E is now explicitly skipped with this limitation as its reason; the existing gateway JWT tests continue to mint tokens directly through the external issuer CLI.

pkg/sandbox-manager/config/manager.go
chengzhycn: The two defaults intentionally operate on different scales. Concurrent requests for one Sandbox join the same in-flight issuance; the refresh interval limits staggered requests that arrive after that issuance completes. It must therefore remain shorter than the minimum token validity so a client receiving `429`/`Retry-After` still has time to retry before expiration. I made that relationship explicit in a5d13b7d by validating `RefreshMinInterval < MinValidity` and added equal/greater boundary tests.

pkg/sandbox-manager/traffic_token.go
chengzhycn: Fixed in 311c42c8. `complete` now clears the recorded attempt time when issuance fails, so only successful issuance consumes the server refresh interval and clients can apply their own retry backoff while the previous token remains valid. I extended the existing Manager table to verify an issuer failure can be retried immediately and added a direct limiter boundary test; the focused tests also pass under `-race -count=3`.

chengzhycn: Follow-up in 473d8ff7: simply clearing `last` on failure could erase a previous successful interval when a bootstrap attempt bypassed the interval and then failed. The limiter now tracks `lastSuccess` only: acquire never advances it, successful completion updates it, and failed completion preserves the prior value. This keeps an initial failure immediately retryable without allowing a later bootstrap failure to remove the previous success budget. Added a regression case for success -> failed bootstrap -> explicit refresh still rate-limited.

pkg/sandbox-manager/infra/sandboxcr/traffic_token.go
chengzhycn: Updated in ea6bda85. Traffic token issuance now uses the informer-cache Sandbox observation and validates policy once before calling the provider; the before/after APIReader reads and UID/owner/phase fencing have been removed. The signed JWT carries the Sandbox ID and UID, and the gateway verifies both against the selected route. The sequence-reader and transition-specific tests were removed, while cache lookup, policy rejection, provider failure, token options, and response handling remain covered.

pkg/sandbox-manager/infra/sandboxcr/traffic_token.go
```diff
@@ -0,0 +1,56 @@
...
+func (i *Infra) IssueTrafficAccessToken(ctx context.Context, opts infra.IssueTrafficAccessTokenOptions) (infra.TrafficAccessToken, error) {
+	if opts.SandboxID == "" {
+		return infra.TrafficAccessToken{}, fmt.Errorf("sandbox ID is required")
+	}
+	if opts.Validate == nil {
+		return infra.TrafficAccessToken{}, fmt.Errorf("traffic token validator is required")
+	}
+	lookup, err := i.lookupSandbox(ctx, infra.GetSandboxOptions{Namespace: opts.Namespace, SandboxID: opts.SandboxID})
+	if err != nil {
+		return infra.TrafficAccessToken{}, err
```
furykerry: plz use wrapGetSandboxError to translate backend specific error to general infra error

pkg/sandbox-manager/traffic_token.go
```diff
@@ -0,0 +1,216 @@
...
+
+	result, issueErr := m.infra.IssueTrafficAccessToken(ctx, infra.IssueTrafficAccessTokenOptions{
+		Namespace:    opts.Namespace,
+		SandboxID:    opts.SandboxID,
+		TokenOptions: m.trafficTokenOptions,
+		Validate: func(sandbox infra.Sandbox) error {
+			return validateTrafficTokenSandbox(sandbox, opts.User)
+		},
+	})
+	if issueErr != nil && managererrors.GetErrCode(issueErr) == managererrors.ErrorUnknown {
```
furykerry, line 98: plz translate backend specific error (cache.ErrSandboxNotFound ) to general infra error (infra.ErrSandboxNotFound) with wrapGetSandboxError

pkg/sandbox-manager/traffic_token.go
```diff
@@ -0,0 +1,216 @@
...
+		return infra.TrafficAccessToken{}, false, managererrors.NewError(managererrors.ErrorInternal, "traffic access token limiter is not configured")
+	}
+
+	limiterKey := string(sandbox.GetUID())
+	if limiterKey == "" {
+		limiterKey = opts.SandboxID
+	}
+	flight, leader, ok := m.trafficTokenLimiter.acquire(limiterKey, enforceInterval)
+	if !ok {
+		return infra.TrafficAccessToken{}, false, managererrors.NewError(managererrors.ErrorRateLimited, "traffic access token refresh is rate limited")
```
furykerry: plz log the rate-limit rejection, it is now not logged in anywhere

pkg/sandbox-manager/traffic_token.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	if !leader {
+		// Only the leader owns flight completion. A follower may stop waiting or
+		// consume the published result, but must not clear or close the shared
+		// flight; the leader does that after issuance returns.
+		select {
+		case <-ctx.Done():
+			return infra.TrafficAccessToken{}, false, managererrors.NewError(managererrors.ErrorUnavailable, "waiting for traffic access token issuance: %v", ctx.Err())
```
furykerry: plz log wait error

pkg/servers/e2b/pause_resume.go
```diff
@@ -318,10 +323,23 @@ func (sc *Controller) ConnectSandbox(r *http.Request) (web.ApiResponse[*models.S
 		return web.ApiResponse[*models.Sandbox]{}, withSandboxResourceContext(err, sbx)
 	}
 	log.Info("sandbox timeout updated")
+	trafficToken, issued, err := sc.manager.BootstrapTrafficAccessToken(ctx, sandboxmanager.RefreshTrafficAccessTokenOptions{
```
furykerry: plz remove BootstrapTrafficAccessToken in resume path, identity provider may not be configured or running.

cmd/sandbox-manager/main.go
```diff
@@ -211,6 +219,11 @@ func main() {
 	if quotaRedisOperationTimeout <= 0 {
 		klog.Fatalf("--quota-redis-operation-timeout must be greater than 0")
 	}
+	trafficTokenOpts := config.TrafficAccessTokenOptions{
+		Validity: trafficTokenValidity, MinValidity: trafficTokenMinValidity, MaxValidity: trafficTokenMaxValidity,
+		RefreshMinInterval: trafficTokenRefreshMinInterval,
+	}
+	mustValidateTrafficAccessTokenOptions(trafficTokenOpts)
```
furykerry: is mustValidateTrafficAccessTokenOptions necessary? Build() (core.go:179) already validates the normalized values with ValidateTrafficAccessTokenOptions.

pkg/sandbox-manager/traffic_token.go
```diff
@@ -0,0 +1,205 @@
...
+			return infra.TrafficAccessToken{}, managererrors.NewError(managererrors.ErrorUnavailable, "waiting for traffic access token issuance: %v", ctx.Err())
+		case <-flight.done:
+			if flight.err != nil {
+				return infra.TrafficAccessToken{}, flight.err
+			}
+			return flight.result, nil
+		}
+	}
+
+	result, issueErr := m.infra.IssueTrafficAccessToken(ctx, infra.IssueTrafficAccessTokenOptions{
```
furykerry, line 90: The leader calls m.infra.IssueTrafficAccessToken(ctx, ...) with its own request context. Waiters that joined via acquire (L88-98) receive flight.err verbatim. If the leader's HTTP client hangs up (or its request times out) mid-issuance, the provider call returns context.Canceled, which gets wrapped as ErrorUnavailable and handed to every waiter — even those whose own contexts are perfectly healthy.

Consider detach the issuance from the leader's cancellation and bound it explicitly:
```
// The minted token is shared with every coalesced waiter, so the leader's
// own cancellation must not fail issuance for callers that are still waiting.
issueCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), m.trafficTokenIssueTimeout)
defer cancel()
result, issueErr := m.infra.IssueTrafficAccessToken(issueCtx, infra.IssueTrafficAccessTokenOptions{...})

```

pkg/sandbox-manager/errors/error.go
```diff
@@ -31,6 +31,8 @@ const (
 	ErrorUnknown       = ErrorCode("Unknown")
 	ErrorBadRequest    = ErrorCode("BadRequest")
 	ErrorQuotaExceeded = ErrorCode("QuotaExceeded")
+	ErrorRateLimited   = ErrorCode("RateLimited")
```
furykerry: No production code produces ErrorRateLimited any more , — the limiter returns a cached token instead.  consider remove the ErrorRateLimited case

sdk/customized_e2b/kruise_agents/patch_traffic_token.py
```diff
@@ -0,0 +1,416 @@
...
+
+
+def _install_lazy_token_manager_if_required(sandbox, info) -> None:
+    if not sandbox.traffic_access_token and _requires_traffic_token(info):
+        _install_lazy_token_manager(sandbox)
+
+
+def _connect_sandbox_sync(cls, *args, **kwargs):
+    sandbox = _original_sync_cls_connect_sandbox(cls, *args, **kwargs)
+    if not sandbox.connection_config.debug and not sandbox.traffic_access_token:
```
furykerry, line 213: For non-JWT sandboxes — the common case — traffic_access_token is always falsy, so sandbox.get_info() fires on every connect just to discover that no manager is needed.

Recommendation: The metadata you need is already in the connect/get_info response the SDK just fetched — reuse it rather than issuing a second call,

pkg/sandbox-manager/traffic_token.go
```diff
@@ -0,0 +1,205 @@
...
+			// connect without triggering duplicate issuance or receiving 429.
+			entry.lastSuccess = l.now()
+			entry.result = result
+		}
+		l.entries[key] = entry
+	}
+	close(flight.done)
+}
+
+func (l *trafficTokenLimiter) cleanup(now time.Time) {
```
furykerry: Issue: trafficTokenLimitEntry now stores the minted token itself. cleanup runs only from inside acquire, so entries are purged only when some other refresh happens. If refresh traffic stops — process idle, sandbox deleted, tenant gone — the last batch of signed JWTs stays resident in the manager's heap indefinitely. This sits uneasily beside the explicit contract on infra.TrafficAccessToken ("must never be persisted") and turns a bounded coalescing map into an unbounded-lifetime credential store.

Recommendation: Evict on a timer rather than opportunistically, and drop the cached material as soon as it is unusable:

pkg/sandbox-manager/traffic_token.go
```diff
@@ -0,0 +1,205 @@
...
+	if limiterKey != "" {
+		// A reusable Sandbox CR can serve multiple deliveries. Include the
+		// delivery ID so a recycled CR cannot receive its previous token.
+		limiterKey += "\x00" + opts.SandboxID
+	} else {
+		limiterKey = opts.SandboxID
+	}
+	flight, cached, leader := m.trafficTokenLimiter.acquire(limiterKey)
+	if flight == nil {
+		return cached, nil
```
furykerry: refresh_traffic_access_token(force=True) — advertised in the README as the explicit escape hatch — will silently receive the cached token for up to RefreshMinInterval.  A caller trying to recover from a suspected-compromised or rejected token cannot get a new one for 30 s, and the response gives no way to tell cached from fresh.

Recommendation: Either document the bound explicitly in the README (force is capped by the server's --traffic-access-token-refresh-min-interval; it guarantees a valid token, not a new one), or plumb an explicit bypass for genuinely forced refreshes. Documenting is sufficient for me.

pkg/identity/sandbox_token_helper.go
```diff
@@ -116,7 +116,7 @@ func IssueSandboxToken(ctx context.Context, sbx *agentsv1alpha1.Sandbox) (*Token
 	log := klog.FromContext(ctx).WithValues("sandbox", klog.KObj(sbx), "action", "IssueSandboxToken")
 	start := time.Now()
 
-	tokenResp, err := IssueToken(ctx, sbx, TokenKindIDToken)
+	tokenResp, err := IssueToken(ctx, sbx, TokenOptions{Kind: TokenKindIDToken})
```
HARSHRAJ2789, line 119: This response is returned to callers unchecked, and this is the path whose value gets persisted.

`refresher.go:92` calls `IssueSandboxToken`, `patchAnnotation` writes `AccessTokenExpiration`
into `AgentKeyTokenRefreshStatus` (`refresher.go:124`), and `token_refresh_controller.go:253`
reads that annotation back to derive the next refresh. A provider returning an empty or
unparseable expiration therefore has it recorded as a successful issuance.

That failure is terminal, not noisy. `decodeTokenRefreshStatus` then yields an empty
`AccessTokenExpiration`, `isRefreshTarget` returns false, and the Sandbox is never enqueued
again. No event, no requeue.

`validateAccessTokenResponse` at `:161` already rejects exactly those shapes, and
`IssueSandboxAccessToken` calls it at `:153`. Taking the token kind as a parameter for the
error message would let both paths share the one validator.

(Same point as my 14 August review, on the line this time.)

cmd/sandbox-manager/main.go
chengzhycn: Fixed in 44695d59. Removed the entrypoint-level `mustValidateTrafficAccessTokenOptions` call and helper. `SandboxManagerBuilder.Build` remains the single validation point for normalized TrafficAccessToken options.

pkg/sandbox-manager/errors/error.go
chengzhycn: Fixed in 44695d59. Removed the unused `ErrorRateLimited` domain code together with the unreachable E2B 429/Retry-After mapping and its test case.

pkg/sandbox-manager/traffic_token.go
chengzhycn: Fixed in 44695d59. A canceled follower now emits a structured sandbox-manager error log with the context error and Sandbox ID before returning the unavailable domain error.

pkg/sandbox-manager/traffic_token.go
chengzhycn: The rate-limit rejection path no longer exists. Commit 44695d59 removes the remaining `ErrorRateLimited` code and E2B 429 mapping, so there is no rejection event to log.

pkg/sandbox-manager/traffic_token.go
chengzhycn: Fixed in 44695d59. The sandboxcr implementation now converts backend lookup errors at the Infra boundary, and Manager classifies only the neutral `infra.ErrSandboxNotFound`; the direct dependency on `cache.ErrSandboxNotFound` was removed.

pkg/sandbox-manager/infra/sandboxcr/traffic_token.go
chengzhycn: Fixed in 44695d59. `IssueTrafficAccessToken` now translates lookup failures with `wrapGetSandboxError`, and the table-driven Infra test verifies a cache miss is exposed as `infra.ErrSandboxNotFound` without invoking validation or the provider.

pkg/sandbox-manager/traffic_token.go
chengzhycn: Fixed in 8712eb14. Removed completed Traffic JWT result caching, lazy cleanup, and `RefreshMinInterval` entirely. The map now contains only active in-flight issuances; `complete` publishes the result to already-joined waiters and deletes the map entry immediately, so sandbox-manager retains no completed credential cache. Added success/failure singleflight and sequential re-issuance coverage, including race runs.

pkg/sandbox-manager/traffic_token.go
chengzhycn: Fixed in 8712eb14. Since completed results and `RefreshMinInterval` were removed, a later explicit refresh now reaches the issuer and receives a newly issued token. Only requests that overlap an issuance already in progress share that singleflight result. README and the user guide were updated to document this behavior.

DahuK:
As I read it, the token lifetime is a single process-wide constant with no way to vary it per request, per tenant, or per sandbox. Please correct me if I've missed a path, Could we make the token TTL a configurable parameter through RequestedValidity when a user requests a token, rather than an operator-level global parameter? This might better fit the onboarding requirements in multi-tenant scenarios, and also make the later migration from 100-year tokens to short-lived tokens smoother.

pkg/sandbox-manager/traffic_token.go
chengzhycn: Fixed in 7a13403a. Shared Traffic JWT issuance now runs on a detached context with an independent 30-second bound, while the leader and every follower wait using their own request contexts. A canceled leader returns immediately without canceling issuance, and healthy waiters still receive the shared result. The detached work is panic-safe, preserves cancellation/timeout causes, and is canceled and drained by `SandboxManager.Stop` before Infra stops. Added repeated race coverage for leader cancellation, shared timeout propagation, provider panic cleanup/retry, and manager lifecycle shutdown.

furykerry:
> As I read it, the token lifetime is a single process-wide constant with no way to vary it per request, per tenant, or per sandbox. Please correct me if I've missed a path, Could we make the token TTL a configurable parameter through RequestedValidity when a user requests a token, rather than an operator-level global parameter? This might better fit the onboarding requirements in multi-tenant scenarios, and also make the later migration from 100-year tokens to short-lived tokens smoother.

i think as an improvement, per sandbox RequestedValidity can be implemented in a separate patch, which can address specific issue of migration from 100-year tokens. Process-wide RequestedValidity is acceptable

PR #750, opened 2026-08-07
feat: implement two-phase upgrade for paused sandboxes

zmberg:
## Summary

This PR implements a two-phase upgrade flow for paused sandboxes via SandboxUpdateOps, plus comprehensive unit tests (92.1% coverage on the `sandboxupdateops` package).

## Background

Previously, SandboxUpdateOps could only upgrade sandboxes in the Running phase. When a paused sandbox needed upgrading, it had to be manually resumed first. This PR enables direct upgrade of paused sandboxes through an automated two-phase process.

### Two-phase upgrade flow (`sandboxupdateops_controller.go`, `patch.go`)

**Phase 1 — Resume trigger:**
- `applySandboxPatch` detects `Phase == Paused` and sets the `AnnotationUpgradeResumeTrigger` annotation **without** patching the template
- The sandbox controller resumes with the OLD template, then transitions to `ResumeSucceed` state

**Phase 2 — Template patch:**
- New `applyTemplatePatch` method patches the template and removes the resume trigger annotation
- `Reconcile` processes `resumeSucceedCandidates` (no concurrency limit) before handling new candidates

### Classification changes (`classifySandbox`)
- New `sandboxResumeSucceed` state for sandboxes with `Upgrading` condition `Reason=ResumeSucceed`
- Paused phase validation: requires `Paused condition=True` AND `Spec.Paused=true` to be an upgrade candidate
- `CheckpointFailed` added to failure reason list
- `isStateIncluded` default now includes `Paused` (alongside `Running`) when `StateFilter` is nil or empty

### Upgrade control (`upgrade_control.go`, `common_control.go`)
- New `Resuming → ResumeSucceed` states in the upgrade state machine
- `ResumeFunc` type extracted from `EnsureSandboxResumed` for reuse in upgrade path
- `SkipInitializeOnResume` support to skip post-resume Initialize on the old pod
- Event recording for `UpgradeResuming` / `UpgradeResumed` transitions

## Test Coverage

- Fixed 2 failing tests (`TestClassifySandbox_StateFilter`, `StateFilterPausedOnly`) — added `Spec.Paused=true` to match new Paused validation
- Added 3 Paused validation branch tests (no condition, condition=False, spec.Paused=false)
- Added ResumeSucceed state classification test
- Added CheckpointFailed to failed reasons table
- Phase 1: `TestApplySandboxPatch_PausedSetsResumeTriggerAnnotation`, `TestApplySandboxPatch_SkipInitializeOnResume`
- Phase 2: `TestApplyTemplatePatch_Success`, `TestApplyTemplatePatch_PatchError`
- Reconcile: `TestReconcile_Phase2PatchForResumeSucceed`, `TestReconcile_Phase2PatchPausedOpsSkipsPhase2`
- `TestIsStateIncluded_DefaultIncludesPaused` (11 sub-tests)

**Coverage: 92.1% of statements**

## Review Guidance

Focus review on:
1. `classifySandbox` Paused validation logic — ensure the `condition=True + spec.Paused=true` guard is correct
2. `applySandboxPatch` phase 1 / `applyTemplatePatch` phase 2 split — annotation set/remove correctness
3. `Reconcile` phase 2 loop — verify `resumeSucceedCandidates` are processed before new candidates

### DIFF
--- a/pkg/controller/sandbox/core/pod_control.go
+++ b/pkg/controller/sandbox/core/pod_control.go
@@ -46,6 +46,12 @@ type PodGenerateArgs struct {
 	Client    client.Client
 	Box       *agentsv1alpha1.Sandbox
 	NewStatus *agentsv1alpha1.SandboxStatus
+	// IsResume indicates that this pod creation is a resume from a paused
+	// state (deep hibernation). The generator uses this to inject resume-only
+	// annotations (recover-from-instance-id, source-pod-uid, recreating)
+	// from the sandbox status PodInfo, instead of relying on the sandbox
+	// phase which may be Upgrading during an upgrade's Resuming stage.
+	IsResume bool
 }
 
 // PodGenerateFunc generates a Pod from a Sandbox spec.
@@ -65,6 +71,12 @@ type CreatePodArgs struct {
 	// checkpoint delta applied on top of the template may override the sidecar
 	// configuration, so the current template is not authoritative there.
 	AdvertiseRuntimeTLS bool
+	// IsResume indicates that this pod creation is a resume from a paused
+	// state (deep hibernation). The generator uses this to inject resume-only
+	// annotations (recover-from-instance-id, source-pod-uid, recreating)
+	// from the sandbox status PodInfo, instead of relying on the sandbox
+	// phase which may be Upgrading during an upgrade's Resuming stage.
+	IsResume bool
 }
 
 // PodControl manages Pod creation for sandbox controllers.
@@ -126,7 +138,7 @@ func (c *PodControl) CreatePod(ctx context.Context, args CreatePodArgs) (*corev1
 		}
 	}
 
-	pod, err := c.generatePod(ctx, PodGenerateArgs{Client: c.Client, Box: box, NewStatus: args.NewStatus})
+	pod, err := c.generatePod(ctx, PodGenerateArgs{Client: c.Client, Box: box, NewStatus: args.NewStatus, IsResume: args.IsResume})
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/controller/sandbox/core/upgrade_control.go
+++ b/pkg/controller/sandbox/core/upgrade_control.go
@@ -167,61 +167,7 @@ func (r *UpgradeControl) EnsureSandboxUpgraded(ctx context.Context, args EnsureF
 	// will only be re-triggered after the upgrade succeeds and the Sandbox
 	// enters Running state.
 	case agentsv1alpha1.SandboxUpgradingReasonResuming:
-		// The sandbox was paused — resume it before proceeding with the upgrade.
-		pausedCond := utils.GetSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionPaused))
-		if pausedCond == nil || pausedCond.Status == metav1.ConditionFalse {
-			// Sandbox is still pausing (or pause condition missing), wait for it to complete.
-			klog.InfoS("Sandbox is still pausing, waiting before upgrade", "sandbox", klog.KObj(box))
-			return nil
-		}
-		// Record the resume event only on the first entry into Resuming;
-		// subsequent reconciles see a non-empty Message and skip the
-		// duplicate event while waiting for pod readiness.
-		if upgradeCond.Message == "" {
-			r.recordUpgradeEvent(box, corev1.EventTypeNormal, EventUpgradeResuming, "Resuming paused sandbox for upgrade")
-			upgradeCond.Message = "Resume triggered, waiting for pod readiness"
-			utils.SetSandboxCondition(newStatus, *upgradeCond)
-		}
-		if err := r.resumeFunc(ctx, args); err != nil {
-			return err
-		}
-		// Check if resume succeeded by looking at the Resumed condition.
-		resumedCond := utils.GetSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionResumed))
-		if resumedCond == nil || resumedCond.Status != metav1.ConditionTrue {
-			klog.InfoS("Sandbox resume in progress, waiting before upgrade", "sandbox", klog.KObj(box))
-			return nil
-		}
-		// pod may be nil here if resumeFunc just created it but the local
-		// args.Pod copy is stale. Re-fetch is not needed — the controller will
-		// re-reconcile with the updated pod. Just wait for the next cycle.
-		if pod == nil {
-			klog.InfoS("Pod not yet available after resume, waiting for next reconcile", "sandbox", klog.KObj(box))
-			return nil
-		}
-		pCond := utils.GetPodCondition(&pod.Status, corev1.PodReady)
-		if pCond == nil || pCond.Status != corev1.ConditionTrue {
-			klog.InfoS("Waiting for pod ready before initialization", "sandbox", klog.KObj(box))
-			return nil
-		}
-		// SkipInitializeOnResume allows users to skip the post-resume Initialize
-		// call (runtime re-init, security token, CSI re-mount) on the old pod
-		// that is about to be deleted in the UpgradePod step. The new pod
-		// created in performRecreateUpgrade is always initialized regardless.
-		if box.Spec.UpgradePolicy != nil && box.Spec.UpgradePolicy.SkipInitializeOnResume {
-			klog.InfoS("Skipping Initialize on resume due to SkipInitializeOnResume", "sandbox", klog.KObj(box))
-		} else if err := r.initializer.Initialize(ctx, box, newStatus); err != nil {
-			return err
-		}
-		r.syncStatusFromPod(pod, newStatus, false)
-		// Resume succeeded. Transition to ResumeSucceed and wait for
-		// SandboxUpdateOps to patch the template before proceeding.
-		klog.InfoS("Sandbox resumed successfully, waiting for template patch", "sandbox", klog.KObj(box))
-		r.recordUpgradeEvent(box, corev1.EventTypeNormal, EventUpgradeResumed, "Sandbox resumed, waiting for template patch")
-		upgradeCond.Reason = agentsv1alpha1.SandboxUpgradingReasonResumeSucceed
-		upgradeCond.Message = ""
-		utils.SetSandboxCondition(newStatus, *upgradeCond)
-		utils.RemoveSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionPaused))
-		return nil
+		return r.handleResuming(ctx, args, upgradeCond, newStatus)
 	// ResumeSucceed is a transient waiting state. When the template is
 	// patched, calculateStatus detects the hash change and calls
 	// determineUpgradeResumeReason to transition to PreUpgrade.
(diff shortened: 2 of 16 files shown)


pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -358,7 +387,10 @@ func (r *UpgradeControl) performRecreateUpgrade(ctx context.Context, args Ensure
 
 	// Step 2: Create new Pod (old pod deleted)
 	if pod == nil {
-		klog.InfoS("Creating new pod for upgrade", "sandbox", klog.KObj(box))
+		klog.InfoS("Creating new pod for upgrade, waiting 1s to mitigate VK status reporting issues", "sandbox", klog.KObj(box))
+		// Wait 1 second before creating the pod to mitigate status
+		// reporting issues caused by the virtual kubelet.
+		time.Sleep(time.Second)
```
furykerry: An unconditional 1s blocking sleep is added to every recreate upgrade (not just paused ones) to mitigate VK status reporting. It holds a reconcile worker and ignores ctx.Done(), contrary to the project guideline "check cancellation in retrying or long-running work". Consider reenqueue for 1 seconds and leave a code comment about why the wait is necessary.

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -190,14 +190,36 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			if err := r.applyTemplatePatch(ctx, sbx, ops); err != nil {
+				klog.ErrorS(err, "Failed to apply template patch (phase 2)",
+					"sandbox", klog.KObj(sbx), "ops", klog.KObj(ops))
+				r.Recorder.Eventf(ops, v1.EventTypeWarning, "PatchFailed",
+					"Failed to patch sandbox %s: %v", sbx.Name, err)
+				if patchErr == nil {
+					patchErr = err
+				}
+			} else {
+				r.Recorder.Eventf(ops, v1.EventTypeNormal, "SandboxUpgrading",
```
furykerry: phase 2 is an terminology for detail implementation, consider rephase the event message as "Patching sandbox %s after resuming.

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -240,7 +259,7 @@ func isStateIncluded(ops *agentsv1alpha1.SandboxUpdateOps, phase agentsv1alpha1.
 		states = ops.Spec.StateFilter.States
 	}
 	if len(states) == 0 {
-		return phase == agentsv1alpha1.SandboxRunning
+		return phase == agentsv1alpha1.SandboxRunning || phase == agentsv1alpha1.SandboxPaused
```
furykerry: why change the  default behavior, which  change may breaks existing installation

pkg/controller/sandbox/core/pod_control.go
```diff
@@ -46,6 +46,12 @@ type PodGenerateArgs struct {
 	Client    client.Client
 	Box       *agentsv1alpha1.Sandbox
 	NewStatus *agentsv1alpha1.SandboxStatus
+	// IsResume indicates that this pod creation is a resume from a paused
+	// state (deep hibernation). The generator uses this to inject resume-only
+	// annotations (recover-from-instance-id, source-pod-uid, recreating)
+	// from the sandbox status PodInfo, instead of relying on the sandbox
+	// phase which may be Upgrading during an upgrade's Resuming stage.
+	IsResume bool
```
furykerry, line 54: IsResume is unread in the OSS tree: [pod_control.go#L46-L79](file:///Users/shouchen/git/kruise-agents/agents/pkg/controller/sandbox/core/pod_control.go) plumbs IsResume into PodGenerateArgs, but GeneratePodFromSandbox/generateBasePodFromSandbox never read it

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -190,14 +191,36 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 	}
 
+	// Process phase 2: patch template for ResumeSucceed sandboxes (no concurrency
+	// limit). These sandboxes already resumed with the OLD template and just need
+	// the template patch + annotation removal to proceed with the actual upgrade.
+	var patchErr error
+	if newStatus.Phase == agentsv1alpha1.SandboxUpdateOpsUpdating && !ops.Spec.Paused {
+		for i := range resumeSucceedCandidates {
+			sbx := resumeSucceedCandidates[i]
```
furykerry: sbx := resumeSucceedCandidates[i] ([L199](file:///Users/shouchen/git/kruise-agents/agents/pkg/controller/sandboxupdateops/sandboxupdateops_controller.go#L198-L199)) can be for _, sbx := range resumeSucceedCandidates.

pkg/controller/sandbox/sandbox_controller.go
```diff
@@ -609,6 +599,15 @@ func (r *SandboxReconciler) calculateStatus(ctx context.Context, args core.Ensur
 					LastTransitionTime: metav1.Now(),
 				}
 				utils.SetSandboxCondition(newStatus, rCond)
+			} else if box.Annotations[agentsv1alpha1.AnnotationUpgradeResumeTrigger] == agentsv1alpha1.True {
```
furykerry, line 602: consider add ut for this branch for calculateStatus

pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -166,48 +167,13 @@ func (r *UpgradeControl) EnsureSandboxUpgraded(ctx context.Context, args EnsureF
...
-		upgradeCond.Reason = agentsv1alpha1.SandboxUpgradingReasonPreUpgrade
-		upgradeCond.Message = ""
-		utils.SetSandboxCondition(newStatus, *upgradeCond)
-		utils.RemoveSandboxCondition(newStatus, string(agentsv1alpha1.SandboxConditionPaused))
-		fallthrough
+		return r.handleResuming(ctx, args, upgradeCond, newStatus)
+	// ResumeSucceed is a transient waiting state. When the template is
+	// patched, calculateStatus detects the hash change and calls
+	// determineUpgradeResumeReason to transition to PreUpgrade.
+	case agentsv1alpha1.SandboxUpgradingReasonResumeSucceed:
```
furykerry, line 180: Issue: Phase 1 patches label/policy/annotation without the template. If the user sets spec.paused=false while the sandbox is still Paused/Resuming (before the sandbox reaches ResumeSucceed), the sandbox resumes normally with the old template and ends up Running with the ops label, cond == nil, and a non-matching template. In classifySandbox this falls through to sandboxUpdating (L442). It is no longer a candidate (it has the ops label), and phase 2 only processes sandboxResumeSucceed — no code path ever applies the template patch. The ops reports updating ≥ 1 forever and can never reach Completed. Before this PR applySandboxPatch always patched the template, so this state was unreachable.

Consider route labeled sandboxes that are past phase 1 but lost the paused state back into the patching path, e.g. in classifySand

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -390,6 +447,14 @@ func (r *Reconciler) classifySandbox(ctx context.Context, sbx *agentsv1alpha1.Sa
 		return sandboxNoNeedUpdate
 	}
 
+	// Phase 2 of two-phase upgrade: resume succeeded, now ready for template
+	// patch. The sandbox controller resumed with the OLD template; ops must
+	// now patch the template and remove the resume trigger annotation to
+	// proceed with the actual pod replacement.
+	if cond != nil && cond.Reason == agentsv1alpha1.SandboxUpgradingReasonResumeSucceed {
+		return sandboxResumeSucceed
+	}
```
furykerry: Issue: Deletion cleanup only removes LabelSandboxUpdateOps. If an ops is deleted during the phase-1→phase-2 window, the sandbox keeps AnnotationUpgradeResumeTrigger=true. If that sandbox is later paused and re-paused completes, [calculateStatus L602](file:///Users/shouchen/git/kruise-agents/agents/pkg/controller/sandbox/sandbox_controller.go#L602-L611) fires the two-phase flow again with no ops to perform phase 2 → stuck in ResumeSucceed ("waiting for template patch" forever).

Fix: in handleDeletion, extend the merge patch to null out the annotation alongside the label:

pkg/controller/sandbox/core/pod_control.go
zmberg: This is used for the ACS scenario, so it needs to be kept for now.

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -390,6 +441,22 @@ func (r *Reconciler) classifySandbox(ctx context.Context, sbx *agentsv1alpha1.Sa
 		return sandboxNoNeedUpdate
 	}
 
+	// Phase 1 labeled the sandbox but it resumed normally (e.g., user set
```
furykerry: the comment is quite confusing, consider rephase it like this
```
    // The sandbox is labeled by this updateops and Running, yet it never entered
	// the upgrade flow (no Upgrading condition) and its template still
	// differs from the patch target. This happens when phase 1 set only the
	// resume trigger and the sandbox was un-paused (spec.Paused=false)
	// before it reached ResumeSucceed, so the template patch was never
	// applied. Treat it as resume-succeeded so Reconcile applies the
	// pending template patch.

``

File: pkg/controller/sandboxupdateops/sandboxupdateops_controller.go [hunk 10]
```diff
@@ -352,6 +390,19 @@ func (r *Reconciler) classifySandbox(ctx context.Context, sbx *agentsv1alpha1.Sa
 		if !isStateIncluded(ops, sbx.Status.Phase) {
 			return sandboxNoNeedUpdate
 		}
+		// Only a fully paused sandbox (Paused condition=True and
```
furykerry: the comment is quite confusing, consider rephase it like this
```
		// Only a fully paused sandbox (Paused condition=True with
		// spec.Paused=true) can enter the two-phase upgrade. Transient
		// states — pause still in progress, or an un-pause already
		// requested — are skipped here; the sandbox will be picked up by a
		// later reconcile once it stabilizes.
```

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -390,6 +441,22 @@ func (r *Reconciler) classifySandbox(ctx context.Context, sbx *agentsv1alpha1.Sa
...
 
+	// Phase 1 labeled the sandbox but it resumed normally (e.g., user set
+	// spec.Paused=false before reaching ResumeSucceed). The template was
+	// never patched, so route it back to the template patching path so
+	// applyTemplatePatch can apply the template in phase 2.
+	if cond == nil && sbx.Status.Phase == agentsv1alpha1.SandboxRunning {
+		return sandboxResumeSucceed
+	}
+
+	// Phase 2 of two-phase upgrade: resume succeeded, now ready for template
```
furykerry: the comment is quite confusing, consider rephase it like this:
```
	// Resume with the OLD template succeeded and the sandbox now waits for
	// the template patch (phase 2 of the two-phase upgrade). Reconcile
	// applies the patch and removes the resume trigger annotation, which
	// kicks off the actual pod-replacement upgrade.
```

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -390,6 +441,22 @@ func (r *Reconciler) classifySandbox(ctx context.Context, sbx *agentsv1alpha1.Sa
...
+	// patch. The sandbox controller resumed with the OLD template; ops must
+	// now patch the template and remove the resume trigger annotation to
+	// proceed with the actual pod replacement.
+	if cond != nil && cond.Reason == agentsv1alpha1.SandboxUpgradingReasonResumeSucceed {
```
furykerry: consider move L456 to L426, and combine three `cond != nil` check in a single if clause to make the code more reasonable
```
if cond != nil {
     if cond.Reason == agentsv1alpha1.SandboxUpgradingReasonSucceeded && cond.Status == metav1.ConditionTrue {
		return sandboxUpdated
	} else if cond.Reason == agentsv1alpha1.SandboxUpgradingReasonResumeSucceed {
        return sandboxResumeSucceed
    } else if.cond.Status == metav1.ConditionFalse &&
		(cond.Reason == agentsv1alpha1.SandboxUpgradingReasonPreUpgradeFailed ||
			cond.Reason == agentsv1alpha1.SandboxUpgradingReasonPostUpgradeFailed ||
			cond.Reason == agentsv1alpha1.SandboxUpgradingReasonUpgradePodFailed ||
			cond.Reason == agentsv1alpha1.SandboxUpgradingReasonCheckpointFailed) {
		return sandboxFailed
	}
}

```

pkg/controller/sandboxupdateops/sandboxupdateops_controller.go
```diff
@@ -250,7 +281,7 @@ func isStateIncluded(ops *agentsv1alpha1.SandboxUpdateOps, phase agentsv1alpha1.
 	return false
 }
 
-func (r *Reconciler) classifySandboxes(ctx context.Context, sandboxList *agentsv1alpha1.SandboxList, ops *agentsv1alpha1.SandboxUpdateOps) (updated, failed, updating int32, candidates []*agentsv1alpha1.Sandbox, requeueResult *ctrl.Result) {
+func (r *Reconciler) classifySandboxes(ctx context.Context, sandboxList *agentsv1alpha1.SandboxList, ops *agentsv1alpha1.SandboxUpdateOps) (updated, failed, updating int32, candidates, resumeSucceedCandidates []*agentsv1alpha1.Sandbox, requeueResult *ctrl.Result) {
 	for i := range sandboxList.Items {
 		sbx := &sandboxList.Items[i]
 		if !sbx.DeletionTimestamp.IsZero() || !isStateIncluded(ops, sbx.Status.Phase) {
```
furykerry: A paused upgrade requires StateFilter=[Running,Paused], and isStateIncluded only implicitly adds Upgrading. So a labeled sandbox that transits through any other phase becomes invisible to the ops — it drops out of total, and the terminal check updated+failed == total && len(candidates) == 0 (L185) can fire with the claimed sandbox still mid-flight.

Concrete trace (reachable race):
1. ops with StateFilter=[Running,Paused] applies phase 1 to a paused sandbox (label + annotation, no template).
2. The user un-pauses (spec.paused=false) before the sandbox controller consumes the annotation. calculateStatus gives un-pause priority → sandbox enters phase Resuming (normal resume, annotation ignored).
3. SandboxEventHandler enqueues the ops on that phase change; ResourceVersionExpectations is satisfied once the informer catches up.
4. classifySandboxes: isStateIncluded(Resuming) = false → continue → sandbox uncounted. If it's the only in-flight sandbox: total==0, updated+failed==0==total, candidates==0 → ops flips to Completed.
5. The sandbox finishes resuming with the OLD template; stale label + annotation remain; the upgrade is silently lost (a later ops would re-claim it, but this ops reports success without upgrading).

Fix: State filter should gate new candidate selection, not tracking of already-claimed sandboxes:
```
// classifySandboxes
if !sbx.DeletionTimestamp.IsZero() {
    continue
}
// The state filter only gates new candidates. A sandbox already claimed by
// this ops must stay tracked regardless of its phase (e.g. Resuming after a
// user un-pause mid two-phase upgrade), otherwise the ops could complete
// while the sandbox is still pending its template patch.
if sbx.Labels[agentsv1alpha1.LabelSandboxUpdateOps] != ops.Name &&
    !isStateIncluded(ops, sbx.Status.Phase) {
    continue
}

```

PR #751, opened 2026-08-07
fix(sandboxcr): return definitive cache misses immediately

### LINKED ISSUE #741: bug: E2B DELETE /sandboxes returns 204 without deleting the sandbox when sandbox lookup fails
### Background

The E2B-compatible DELETE sandbox handler returns `204 No Content` whenever the sandbox lookup fails, without ever invoking the delete path. Clients therefore observe a successful delete while the sandbox pod is never deleted. The condition is transient but its consequence is durable: orphaned sandboxes remain with no further delete request from the client.

### Issue Type

bug

### Relevant Code

`pkg/servers/e2b/services.go` — `DeleteSandbox` (around lines 91-97): any error from `getSandboxOfUser` is logged as `"failed to get sandbox, just return success"` and the handler returns 204 without calling `manager.DeleteSandbox`.

`getSandboxOfUser` (`pkg/servers/e2b/sandbox.go`) contributes to the trigger surface:

- It applies a hard 2-second timeout to the lookup.
- The lookup (`Infra.GetSandbox` → `lookupSandbox` in `pkg/sandbox-manager/infra/sandboxcr/infra.go`) polls the informer cache only; on a cache miss there is no APIReader fallback (the fallback applies only to stale cache hits).
- It only accepts states `{Running, Paused, Dead}`; any other state yields an error.

### Notes

- Under high concurrency or informer resync storms, multiple manager replicas may have lagging caches simultaneously. A burst of concurrent DELETEs can all hit the 2s lookup timeout, each returning a false-success 204 while no pod is actually deleted.
- This differs semantically from the other 204 paths in the handler (sandbox not viewable / genuinely gone), where returning success is expected. Here the 204 is returned because the system failed to determine the sandbox's existence, yet it is indistinguishable to the caller from a real delete.
- Mechanism is confirmed from code; observed frequency is likely load-correlated.

AiRanthem:
## What changed

- Return a claimed-sandbox cache miss immediately when there is no existing route indicating informer propagation.
- When a route exists, poll the informer cache only within the caller-provided deadline and retain the newest observed route resource version.
- Refresh stale cache hits through the APIReader using the cached object's unambiguous key.
- Remove `singleflight` from `GetClaimedSandbox` so every polling attempt performs an independent read of the current informer cache.
- Cover the observable no-route, route-driven polling, stale-delete, and claimed-state delete behavior.

## Root cause and impact

The E2B DELETE handler intentionally treats a genuinely absent or non-viewable sandbox as an idempotent `204 No Content`. However, the lookup path could also fail while the informer cache was still catching up, making that transient lookup failure indistinguishable from a real absence and allowing DELETE to return success without invoking the delete path.

This change uses route presence as evidence that a cache miss may be transient. Only those misses are retried; definitive misses return immediately. Once the cache catches up, a newer route resource version still triggers the existing APIReader refresh before state validation and deletion.

## Reviewer note: why remove `singleflight`

`GetClaimedSandbox` is a cheap local informer-cache read, while its caller deliberately polls to observe cache progress. Coalescing these reads can make concurrent callers and retries share an in-flight result created with another caller's context, so they may all observe the same stale miss or context failure instead of independently reading the latest cache state. Removing `singleflight` keeps cancellation scoped to each caller and makes every retry a fresh cache observation.

No dedicated regression test is added solely for the `singleflight` removal. `singleflight` is an internal coalescing detail, and a timing-based test that tries to prove two cache reads do not overlap would be brittle while adding no user-visible contract. The deterministic lookup tests instead cover the observable behavior: definitive misses, route-backed retries, caller cancellation, retained route resource versions, and stale APIReader fallback.

## Validation

```text
go test ./pkg/cache ./pkg/sandbox-manager/infra/sandboxcr ./pkg/servers/e2b ./pkg/utils
```

Fixes #741


pkg/sandbox-manager/api.go
```diff
@@ -210,6 +210,9 @@ func (m *SandboxManager) CloneSandbox(ctx context.Context, opts CloneSandboxOpti
 // When expectedStates is empty, ownership is still checked but any claimed state
 // is accepted. When expectedStates is non-empty, the sandbox state must match one
 // of the provided values.
+//
+// The infra lookup may wait while ctx is live, so pass a context with a
+// deadline (e.g. context.WithTimeout).
```
furykerry, line 220: for GetSandbox errors other than ErrSandboxNotFound, return err will cause delete operation return 204, we should return internal error to the user

pkg/sandbox-manager/infra/sandboxcr/infra.go
```diff
@@ -413,42 +413,74 @@ type claimedSandboxLookup struct {
...
+// until the cache observes the Sandbox or the caller's context ends.
 func (i *Infra) lookupSandbox(ctx context.Context, opts infra.GetSandboxOptions) (claimedSandboxLookup, error) {
 	var lookup claimedSandboxLookup
-	err := wait.PollUntilContextCancel(ctx, RetryInterval, true, func(ctx context.Context) (bool, error) {
-		got, err := i.Cache.GetClaimedSandbox(ctx, cache.GetClaimedSandboxOptions{Namespace: opts.Namespace, SandboxID: opts.SandboxID})
+	err := i.readClaimedSandbox(ctx, opts, &lookup)
+	if err == nil {
+		return lookup, nil
+	}
+	if !errors.Is(err, cache.ErrSandboxNotFound) || !lookup.observeRoute(i.Proxy, opts.SandboxID) {
```
furykerry: for errors that is not ErrSandboxNotFound, we should pollClaimedSandbox instead

pkg/utils/utils.go
```diff
@@ -296,6 +296,17 @@ func GetSandboxID(sbx *agentsv1alpha1.Sandbox) string {
 	return sbx.Namespace + sandboxIDSeparator + sbx.Name
 }
 
+// ParseSandboxID decodes a legacy Sandbox ID produced by GetSandboxID.
+// known-limit: this is only correct while legacy namespaces cannot contain
+// "--"; remove it after Route carries Namespace/Name in the follow-up PR.
+func ParseSandboxID(sandboxID string) (namespace, name string, err error) {
```
furykerry: ParseSandboxID is not used anywhere, consider delete it

PR #757, opened 2026-08-08
fix(controller): reject leftover pod from a previous same-name sandbox

### LINKED ISSUE #756: bug: sandbox controller adopts leftover pod after same-name sandbox recreation
### Background

The sandbox controller fetches the pod purely by the sandbox's NamespacedName
and never verifies the pod's owner reference. When a Sandbox is deleted and
recreated with the same name (new UID), a leftover pod from the previous
generation can still exist, and its ownerReference UID points to the old,
already-deleted Sandbox. `EnsureSandboxRunning` then trusts this pod: the new
Sandbox transitions to Running and syncs PodIP/PodUID from the previous
generation's pod.

This is commonly hit in the snapshot workflow: take a snapshot (Checkpoint)
of a sandbox, delete the sandbox, then create a new sandbox with the same
name (e.g., restored from the snapshot). At the moment the new same-named
Sandbox is created, the previous generation's pod has not been fully removed
yet, so the new Sandbox adopts it.

In addition, the pod-delete validating webhook denies deletion/eviction for
any requester except the sandbox controller's service account as long as a
same-named Sandbox exists and is not being deleted. The garbage collector is
therefore denied when trying to reclaim such an orphaned pod, so the leftover
pod is never cleaned up while the same-named Sandbox exists.

### Issue Type

bug

### Relevant Code

- `pkg/controller/sandbox/sandbox_controller.go` — `Reconcile` gets the pod
  with the same NamespacedName as the Sandbox, with no ownership validation.
- `pkg/controller/sandbox/core/common_control.go` — `EnsureSandboxRunning`
  transitions the Sandbox to Running based on any existing pod, regardless of
  the pod's owner reference UID.
- `pkg/controller/sandbox/core/pod_control.go` — pods are created with a
  controller reference to the Sandbox, so a recreated Sandbox (new UID) no
  longer matches the leftover pod's ownerReference.
- `pkg/webhook/pod/validating/pod_delete.go` — denies pod deletion for all
  users except the sandbox controller when a same-named non-deleting Sandbox
  exists; the GC controller is not exempted.

### Relation to PR #646 (lazy sandbox finalizer)

This issue is closely related to https://github.com/openkruise/agents/pull/646
(merged):

- Before PR #646, every Sandbox carried the `agents.kruise.io/sandbox`
  finalizer. Deleting a Sandbox went through `EnsureSandboxTerminated`, which
  deletes the pod and waits until the pod is gone before removing the
  finalizer. The Sandbox object only disappeared after its pod was fully
  removed, so a same-name recreation could almost never observe a leftover
  pod.
- After PR #646, Running sandboxes carry no finalizer: deleting a Running
  Sandbox removes the object immediately and pod cleanup is delegated to
  asynchronous owner-reference GC. This widens the race window exactly into
  the snapshot workflow (snapshot -> delete -> recreate with the same name
  before GC removes the orphaned pod), so the new Sandbox adopts the
  previous generation's pod.
- Compounding effect: once the new same-named Sandbox exists, the pod-delete
  webhook denies the GC's deletion request, so what used to be a narrow
  cache-lag race becomes a permanent mis-adoption with a lingering pod.

### Notes

- Common reproduction workflow: take a snapshot (Checkpoint) of a sandbox,
  delete the sandbox, then create a new sandbox with the same name before the
  previous pod is fully removed.
- Observed effects: the new Sandbox reports Running with the previous
  generation's pod (stale workload state, PodIP, PodUID in status), and the
  leftover pod persists indefinitely because GC deletion is denied by the
  webhook.
- The same ownership assumption also affects the Pending-phase terminal pod
  detection in `calculateStatus` (`updateStatusIfPodCompleted`), which can
  complete the new Sandbox based on the leftover pod's phase.

furykerry:

The sandbox controller fetches the pod purely by the sandbox's NamespacedName
and never verifies the pod's owner reference. When a Sandbox is deleted and
recreated with the same name (new UID) — commonly hit in the snapshot workflow
(snapshot -> delete -> recreate) — a leftover pod from the previous generation
can still exist, and its ownerReference UID points to the old Sandbox.
`EnsureSandboxRunning` then adopts it: the new Sandbox transitions to Running
with the previous generation's pod (stale workload state, PodIP, PodUID).

This PR adds an ownership guard:

- `EnsureSandboxRunning` no longer adopts a pod whose Sandbox ownerRef UID
  differs from the current sandbox. It stays Pending, emits a
  `Warning/PodOwnerMismatch` event, and waits for the GC controller to reclaim
  the pod; the pod watch retriggers the reconcile afterwards, then a fresh pod
  is created.
- `calculateStatus` treats such a stale pod as absent in the Pending phase, so
  a terminal-phase leftover pod cannot complete the new sandbox.
- The pod-delete validating webhook now allows deletion/eviction when the
  pod's Sandbox ownerRef UID differs from the same-named existing Sandbox.
  Without this, the GC is denied forever and the leftover pod is never
  reclaimed.

fixes #756

1. Unit tests: `go test ./pkg/controller/sandbox/... ./pkg/webhook/pod/... -count=1`
2. New table cases: a running pod owned by a different sandbox UID keeps the
   sandbox Pending (with event); a pending sandbox with a terminal stale pod
   stays Pending; the webhook allows deleting a pod whose Sandbox ownerRef UID
   mismatches the same-named sandbox.
3. Manual: delete a sandbox and recreate one with the same name while the old
   pod still exists — the new sandbox stays Pending until GC removes the old
   pod, then creates its own.

- The controller intentionally does not delete the stale pod itself: it stays
  Pending and relies on the GC, keeping deletion authority in one place. The
  webhook change is required for this to converge, because the webhook
  previously denied GC deletions whenever a same-named sandbox exists (only
  the sandbox controller's service account is exempted).
- Pods without any Sandbox owner reference keep the previous behavior (no
  mismatch check), so pods owned by other controllers are untouched.
- Related background: PR #646 (lazy finalizer) moved Running-sandbox pod
  cleanup from the controller's synchronous termination path to async
  owner-reference GC, which widened this race window.

### DIFF
--- a/pkg/controller/sandbox/core/util.go
+++ b/pkg/controller/sandbox/core/util.go
@@ -78,3 +78,18 @@ func GeneratePVCName(templateName, sandboxName string) (string, error) {
 func GetControllerKey(obj client.Object) string {
 	return types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}.String()
 }
+
+// StaleSandboxPodOwner returns the UID of the Sandbox owner reference on pod
+// when it does not match the current sandbox, i.e. the pod is a leftover from
+// a previous sandbox generation with the same name (see issue #756).
+// Returns "", false when the pod is owned by the current sandbox or carries
+// no Sandbox owner reference.
+func StaleSandboxPodOwner(pod *corev1.Pod, box *agentsv1alpha1.Sandbox) (types.UID, bool) {
+	for i := range pod.OwnerReferences {
+		ref := &pod.OwnerReferences[i]
+		if ref.Kind == "Sandbox" && ref.APIVersion == agentsv1alpha1.SchemeGroupVersion.String() {
+			return ref.UID, ref.UID != box.UID
+		}
+	}
+	return "", false
+}
--- a/pkg/controller/sandbox/core/util.go
+++ b/pkg/controller/sandbox/core/util.go
@@ -88,7 +88,10 @@ func StaleSandboxPodOwner(pod *corev1.Pod, box *agentsv1alpha1.Sandbox) (types.U
 	for i := range pod.OwnerReferences {
 		ref := &pod.OwnerReferences[i]
 		if ref.Kind == "Sandbox" && ref.APIVersion == agentsv1alpha1.SchemeGroupVersion.String() {
-			return ref.UID, ref.UID != box.UID
+			if ref.UID == box.UID {
+				return "", false
+			}
+			return ref.UID, true
 		}
 	}
 	return "", false
--- a/pkg/controller/sandbox/core/common_control.go
+++ b/pkg/controller/sandbox/core/common_control.go
@@ -41,6 +41,10 @@ import (
 
 const CommonControlName = "common"
 
+// eventReasonPodOwnerMismatch is the event reason emitted when an existing
+// pod is owned by a previous sandbox generation with the same name.
+const eventReasonPodOwnerMismatch = "PodOwnerMismatch"
+
 // Container waiting reasons defined by kubelet (not exported as public constants in K8s API).
 const (
 	// WaitingReasonPodInitializing indicates init containers are still running.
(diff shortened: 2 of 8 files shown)


pkg/controller/sandbox/core/util.go
```diff
@@ -78,3 +78,18 @@ func GeneratePVCName(templateName, sandboxName string) (string, error) {
 func GetControllerKey(obj client.Object) string {
 	return types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}.String()
 }
+
+// StaleSandboxPodOwner returns the UID of the Sandbox owner reference on pod
+// when it does not match the current sandbox, i.e. the pod is a leftover from
+// a previous sandbox generation with the same name (see issue #756).
+// Returns "", false when the pod is owned by the current sandbox or carries
```
zmberg, line 85: **[Suggestion]** The doc comment says it returns `"", false` when the pod is owned by the current sandbox, but the code actually returns `(ref.UID, false)` — the first return value is the current sandbox's UID, not an empty string. Consider aligning the comment with the actual behavior (or returning `""` when the UIDs match) so callers are not misled about the return semantics.

pkg/controller/sandbox/core/util.go
```diff
@@ -78,3 +78,18 @@ func GeneratePVCName(templateName, sandboxName string) (string, error) {
...
+// no Sandbox owner reference.
+func StaleSandboxPodOwner(pod *corev1.Pod, box *agentsv1alpha1.Sandbox) (types.UID, bool) {
```
zmberg, line 87: **[Suggestion]** `StaleSandboxPodOwner` is currently only covered indirectly through the table-driven cases of `EnsureSandboxRunning` / `calculateStatus`. Consider adding direct table tests in `util_test.go` for the edge cases: no Sandbox ownerRef (e.g. owned by another controller), matching UID, mismatched APIVersion, and a Sandbox ownerRef that is not the first entry in the list.

pkg/controller/sandbox/core/util.go
furykerry: Fixed in 605130ca: now returns ("", false) when the UIDs match, so the return semantics align with the doc comment.

pkg/controller/sandbox/core/util.go
furykerry: Fixed in 605130ca: added a direct table test TestStaleSandboxPodOwner covering no owner references, ownership by another controller, matching UID, mismatched UID, mismatched APIVersion, and a Sandbox owner reference that is not the first entry.

PR #761, opened 2026-08-08
fix(e2b): correct malformed klog structured logging in csiMountOptionsConfigRecord

### LINKED ISSUE #758: [BUG] Malformed klog.Info call in csiMountOptionsConfigRecord causes broken structured logging
**What happened**:
In `pkg/servers/e2b/create.go`, inside the `csiMountOptionsConfigRecord` function, a `klog` structured logging method (`log.Info`) is called with an odd number of arguments for its key-value pairs. 

Specifically, at line 472:
```go
	csiMountConfigRaw, err := json.Marshal(request.Extensions.CSIMount.MountConfigs)
	if err != nil {
		log.Info("failed to marshal csi mount config", err)
		return
	}
```

Because `log.Info` expects a message followed by `keysAndValues ...interface{}`, providing just the `err` object results in an incomplete key-value pair. This breaks structured logging for this error, typically emitting something like `(EXTRA *errors.errorString=...)` or `(MISSING)` instead of a properly formatted error log, making it difficult to debug CSI mount configuration marshaling failures.

The error should be logged properly using `log.Error` or by providing a key string before the error value in `log.Info`. For example:
```go
		log.Error(err, "failed to marshal csi mount config")
```
OR
```go
		log.Info("failed to marshal csi mount config", "error", err)
```

1. Inspect `pkg/servers/e2b/create.go` at line 472.
2. Trigger a failure when marshaling `request.Extensions.CSIMount.MountConfigs` to JSON during sandbox creation.
3. Observe the malformed log output in the sandbox-manager controller logs.

This appears to be the only instance of this specific malformed logging pattern in the `pkg` directory.

**Environment**:
- OpenKruiseAgent version: master
- Kubernetes version (use `kubectl version`): N/A (Codebase issue)
- Install details (e.g. helm install args): N/A
- Others: N/A

vishalmore90:

This PR fixes a bug in `pkg/servers/e2b/create.go` where a `log.Info` call inside the `csiMountOptionsConfigRecord` function was provided with an odd number of arguments for its key-value pairs (`log.Info("failed to marshal csi mount config", err)`). This broke the structured logging, producing malformed output when trying to log an error.

The PR changes `log.Info` to `log.Error(err, "failed to marshal csi mount config")`, ensuring that the error is correctly handled and logged in the expected structured format.

fixes #758

This can be verified by running the unit tests for the `pkg/servers/e2b` package:
```bash
make test
# OR
go test ./pkg/servers/e2b/...
```
And manually reviewing the code change to ensure the `log.Error` signature conforms to the expected standard.

```
NONE

```

### DIFF
--- a/pkg/servers/e2b/create.go
+++ b/pkg/servers/e2b/create.go
@@ -469,7 +469,7 @@ func (sc *Controller) csiMountOptionsConfigRecord(ctx context.Context, sbx infra
 	// marshal the csi mount confit to json
 	csiMountConfigRaw, err := json.Marshal(request.Extensions.CSIMount.MountConfigs)
 	if err != nil {
-		log.Info("failed to marshal csi mount config", err)
+		log.Error(err, "failed to marshal csi mount config")
 		return
 	}
 	annotations := sbx.GetAnnotations()
--- a/pkg/servers/e2b/create.go
+++ b/pkg/servers/e2b/create.go
@@ -42,6 +42,8 @@ import (
 	"github.com/openkruise/agents/pkg/utils/timeout"
 )
 
+var jsonMarshal = json.Marshal
+
 // noServerTimeout is used when the server should not impose its own deadline on
 // the claim/clone/wait-ready phases of CreateSandbox. The operation is then
 // bounded only by the client request context (cancellation). It is a far-future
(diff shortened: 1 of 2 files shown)


vishalmore90:
@AiRanthem 
/review

Issue #659, opened 2026-07-17
feat: standards-based inbound authentication (signed tokens, identity, RBAC)


furykerry:
### Update (2026-07-25): Interfaces landed, implementations needed

With the merge of #648 (gateway JWT verification via OIDC) and #671 (TokenKind-based access/ID token issuance interface), the OpenKruise Agents codebase now provides the **interfaces and scaffolding** for standards-based token issuance and verification, but **does not yet ship an open-source implementation** that mints real signed tokens or propagates them into sandboxes.

#### What has been delivered

| Capability | PR | What landed |
|---|---|---|
| Access token issuance interface | #671 | `IdentityProvider.IssueToken(ctx, sbx, kind)` with `TokenKindAccessToken`; `IssueSandboxAccessToken` helper; `IsAccessTokenRequested` gating via `security.agents.kruise.io/enable-jwt-auth` annotation; access token held transiently in memory and surfaced on the E2B response. |
| ID token issuance interface | #671 | Same `IssueToken` with `TokenKindIDToken`; `IssueSandboxToken` helper; `IsIDTokenRequested` gating via `security.agents.kruise.io/agent-name` annotation; refresh lifecycle via `SecurityTokenRefreshReconciler`. |
| JWT verification (data plane) | #648 | `pkg/identity/oidc` verifier: OIDC discovery, JWKS snapshot, signature validation, required-claims enforcement (`exp`/`iat`/`nbf`/`iss`/`sub`), sandbox ID/UID binding; `pkg/sandbox-gateway/jwtauth` manager with async initialization and readiness gating. |

#### What is still missing — the community default is a placeholder

The community default provider (`pkg/identity/common_provider.go` → `defaultTokenProvider`) mints **random UUID placeholder tokens** for both `TokenKindIDToken` and `TokenKindAccessToken`:

```go
func (u *defaultTokenProvider) IssueToken(...) (*TokenResponse, error) {
    return &TokenResponse{
        AccessToken:           uuid.NewString(),  // ← not a signed JWT
        SandboxClientID:       uuid.NewString(),
        AccessTokenExpiration: time.Now().Add(expiry).Format(time.RFC3339),
    }, nil
}
```

`PropagateSecurityToken` is a no-op. There is no in-tree provider that:
1. Signs JWTs with a real asymmetric key (RSA/ECDSA/Ed25519).
[...]

PR #772, opened 2026-08-09
feat(identity): add a signed JWT issuer for traffic access tokens

AnshulPatil2005:
Opening this as a draft against #659. It is deliberately only the first piece, and it stops short of the parts that depend on decisions the mentors haven't made yet.

`pkg/identity/oidc` is a complete verifier with nothing in tree on the other side of it. `defaultTokenProvider.IssueToken` returns `uuid.NewString()` for both token kinds, so there is no way to produce a token that verifier would accept.

This adds `pkg/identity/jwtissuer`:

- signs traffic access tokens with the claim shape from `20260713-traffic-access-token-jwt-verification.md`: `iss`, `sub`, `exp`, `iat`, `nbf`, and the `sandbox` claim carrying `sandboxId`/`sandboxUid`
- serves `/.well-known/openid-configuration` and `/jwks.json`, so a community deployment can bootstrap the gateway without an external IdP
- supports RSA, ECDSA P-256/384/521 and Ed25519, picking the algorithm from the key type, since the verifier rejects a token whose `alg` doesn't match its key
- publishes retained public keys alongside the active signing key, for rotation

Nothing is wired into the provider, the manager, or any server yet. This package alone is not user-visible.

Refs #659. First step only, not the whole issue.

`go test ./pkg/identity/jwtissuer/ -count=1 -race`

The tests don't mock the verifier. They stand up an HTTPS test server, publish real discovery and JWKS documents from it, hand `oidc.NewVerifier` the server's CA through a fake-client ConfigMap, and let it perform discovery exactly as the gateway does. What gets asserted is that the production verifier accepts the token. The claim shape and the JWKS rules it enforces (unique non-empty `kid`, public keys only, `alg` matching key type) are checked by the real consumer rather than by my reading of it.

I checked the tests aren't vacuous: dropping `nbf` from the minted claims fails `TestIssuedTokenPassesRealVerifier` with `token is missing nbf claim`.

`TestRotationOverlapWindow` walks the three rotation steps and ends by asserting the hazard is real: a gateway holding a pre-rotation JWKS snapshot rejects a token signed by the new key with `unknown kid`.

**Why this stops here.** @PRAteek-singHWY raised six design questions on #659 that are still open, and several of them change what comes next: whether access tokens get a re-issue path, whether `sub` carries the authenticated principal from the claiming API key or stays a fixed control-plane subject, and which runtime channel carries the ID token. I didn't want to guess at those, so I built the layer that is the same under every answer. Once the issuer exists, the provider wiring is small.

**Rotation.** The proposal states a published verifier is never replaced for the process lifetime, so rotation can only mean an overlap window: publish the new key, roll the gateways, then flip signing. `WithRetainedKeys` is what makes that expressible. Re-fetching JWKS on an unknown `kid` would put network I/O back on the request path and contradict the offline-after-initialization goal, so I've assumed that needs its own proposal.

**Claim struct duplication.** `trafficAccessTokenClaims` mirrors `oidc.TrafficAccessTokenClaims` rather than importing it. Importing would make the round-trip test tautological, since both sides would share one definition and a wrong JSON tag would still pass. The duplication is what gives the test something to catch. Happy to import it instead if you'd rather have the compile-time link.

**Not in scope here:** loading keys from a Secret, ID token issuance, `PropagateSecurityToken`, and the `IdentityProvider` implementation itself.

If the direction is wrong I would rather find out now than after building on top of it. I'm also happy to write this up under `docs/proposals/` first. Reading the two existing proposals together, the issuance side is the piece with no design doc yet.

### DIFF
--- a/pkg/identity/jwtissuer/endpoints.go
+++ b/pkg/identity/jwtissuer/endpoints.go
@@ -22,6 +22,8 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+
+	jose "github.com/go-jose/go-jose/v4"
 )
 
 // Endpoint paths served for OIDC discovery. The gateway is pointed at
@@ -57,10 +59,7 @@ func (i *Issuer) Handler() (http.Handler, error) {
 		return nil, fmt.Errorf("marshal JWKS: %w", err)
 	}
 
-	algorithms, err := i.signingAlgorithms()
-	if err != nil {
-		return nil, err
-	}
+	algorithms := signingAlgorithms(keySet)
 	discoveryBody, err := json.Marshal(discoveryDocument{
 		Issuer:                           i.issuerURL,
 		JWKSURI:                          strings.TrimSuffix(i.issuerURL, "/") + JWKSPath,
@@ -93,13 +92,10 @@ func issuerBasePath(issuerURL string) string {
 }
 
 // signingAlgorithms lists every algorithm a published key can be used with, so
-// the discovery document stays accurate across a rotation overlap window.
-func (i *Issuer) signingAlgorithms() ([]string, error) {
-	keySet, err := i.PublicKeySet()
-	if err != nil {
-		return nil, err
-	}
-
+// the discovery document stays accurate across a rotation overlap window. It
+// takes the key set the caller already built rather than rebuilding it, so the
+// two cannot describe different keys.
+func signingAlgorithms(keySet jose.JSONWebKeySet) []string {
 	seen := make(map[string]struct{}, len(keySet.Keys))
 	algorithms := make([]string, 0, len(keySet.Keys))
 	for _, key := range keySet.Keys {
@@ -109,7 +105,7 @@ func (i *Issuer) signingAlgorithms() ([]string, error) {
 		seen[key.Algorithm] = struct{}{}
 		algorithms = append(algorithms, key.Algorithm)
 	}
-	return algorithms, nil
+	return algorithms
 }
 
 func serveJSON(body []byte) http.HandlerFunc {
--- a/pkg/identity/jwtissuer/endpoints.go
+++ b/pkg/identity/jwtissuer/endpoints.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 )
 
@@ -71,10 +72,24 @@ func (i *Issuer) Handler() (http.Handler, error) {
 		return nil, fmt.Errorf("marshal discovery document: %w", err)
 	}
 
+	// Serve under the issuer URL's own path so the advertised jwks_uri is
+	// actually reachable when the issuer sits behind an ingress path prefix.
+	basePath := issuerBasePath(i.issuerURL)
+
 	mux := http.NewServeMux()
-	mux.HandleFunc(DiscoveryPath, serveJSON(discoveryBody))
-	mux.HandleFunc(JWKSPath, serveJSON(jwksBody))
+	mux.HandleFunc(basePath+DiscoveryPath, serveJSON(discoveryBody))
+	mux.HandleFunc(basePath+JWKSPath, serveJSON(jwksBody))
 	return mux, nil
+}
+
+// issuerBasePath returns the path portion of the issuer URL without a trailing
+// slash, or "" when the issuer is served from the root.
+func issuerBasePath(issuerURL string) string {
+	parsed, err := url.Parse(issuerURL)
+	if err != nil {
+		return ""
+	}
+	return strings.TrimSuffix(parsed.Path, "/")
 }
 
 // signingAlgorithms lists every algorithm a published key can be used with, so
--- a/pkg/identity/jwtissuer/endpoints.go
+++ b/pkg/identity/jwtissuer/endpoints.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


AnshulPatil2005:
flagging an overlap before someone else spots it. #742 also mints signed jwts with this claim shape, so it's worth being clear on what's different.

the provider there is `test/e2b/assets/jwt-oidc-provider`, a standalone binary under test/ with one hardcoded rsa key. it stands in for an external enterprise idp during e2e, which is the deployment shape #659 says the community doesn't have. this package is meant for the case where there is no external idp, so it carries key management, algorithm selection off the key type, and the rotation overlap.

the 1 hour default here happens to line up with the default in #742, which was accidental but convenient.

if #742 lands first i'm happy to rebase onto it, or fold its e2e provider onto this package so there's one signer instead of two, or drop this entirely if the community path should just point at keycloak or dex. mainly want to avoid two things minting the same tokens for different reasons.

pkg/identity/jwtissuer/issuer.go
```diff
@@ -0,0 +1,278 @@
...
+		Algorithm: string(algorithm),
+	}, nil
+}
+
+// signatureAlgorithmFor picks the JWS algorithm for a key type. The verifier
+// rejects a token whose algorithm does not match its key, so this is the single
+// place that decides the pairing.
+func signatureAlgorithmFor(publicKey crypto.PublicKey) (jose.SignatureAlgorithm, error) {
+	switch key := publicKey.(type) {
+	case *rsa.PublicKey:
```
HARSHRAJ2789, line 252: This arm returns RS256 for any `*rsa.PublicKey`, and go-jose adds no floor of its own, so a 1024-bit key loads and signs. RFC 7518 §3.3 requires 2048 or larger for RS256. `key.Size() < 256` here would reject it at `ParsePrivateKeyPEM` time with a clear message, since that path already calls through this function.

pkg/identity/jwtissuer/issuer_test.go
```diff
@@ -0,0 +1,443 @@
...
+// here must be accepted by the verifier the gateway actually runs, bootstrapped
+// the way the gateway actually bootstraps it: OIDC discovery over HTTPS, a JWKS
+// snapshot, and a CA read from a ConfigMap. A mock verifier would prove nothing.
+func TestIssuedTokenPassesRealVerifier(t *testing.T) {
+	tests := []struct {
+		name string
+		key  crypto.Signer
+	}{
+		{name: "ECDSA P-256", key: mustECDSAKey(t, elliptic.P256())},
+		{name: "ECDSA P-384", key: mustECDSAKey(t, elliptic.P384())},
```
HARSHRAJ2789, line 62: ES512 is the one branch of the curve switch with no end-to-end row. I added `{name: "ECDSA P-521", key: mustECDSAKey(t, elliptic.P521())},` locally and it passes against the real verifier, so it is a free line of coverage.

pkg/identity/jwtissuer/endpoints.go
```diff
@@ -0,0 +1,130 @@
...
+	keySet, err := i.PublicKeySet()
+	if err != nil {
+		return nil, fmt.Errorf("build JWKS: %w", err)
+	}
+	jwksBody, err := json.Marshal(keySet)
+	if err != nil {
+		return nil, fmt.Errorf("marshal JWKS: %w", err)
+	}
+
+	algorithms, err := i.signingAlgorithms()
```
HARSHRAJ2789, line 60: Nit: `signingAlgorithms()` calls `PublicKeySet()` again, so the key set is built twice per `Handler()` call. Passing `keySet` in costs one parameter and keeps the two from disagreeing later.

AnshulPatil2005:
thanks for these, all three passes were useful. rebased onto master and pushed, with the following.

**rsa floor.** fixed, and you were right that it belonged here rather than in the verifier. `signatureAlgorithmFor` now refuses anything under 2048 bits with the rfc 7518 reference in the comment, and since `ParsePrivateKeyPEM` routes through it a weak secret fails at load rather than at first issuance. `TestRejectsUndersizedRSAKey` pins 1024 refused and 2048 accepted.

**es512.** added, same row you suggested. it passes, so coverage rather than a defect as you said.

**double keyset build.** `signingAlgorithms` now takes the set the caller already built instead of rebuilding it.

**#796 and the rotation comment.** still open, so the one-shot snapshot holds and the comment was accurate, but you were right that it reads as permanent. reworded to describe current gateway behaviour and to say retention is correct either way: today it is the only thing preventing a fleet of 401s, and with a refresh it becomes a window for verifiers that have not caught up. no code change, the mechanism was already right under both.

**nbf and jti.** both documented rather than changed. nbf equals iat with a note pointing at the one minute `DefaultClockSkew`, so the next reader does not have to go find it. no jti was deliberate: it is only useful with a revocation store nothing maintains, and the trade is written down now, that a leaked token is good for the rest of its hour unless you rotate and drop the retained key.

**aud.** not settling this one here. i agree with your position in #775 that this token should carry none, since `sandbox.sandboxId` already does the binding and the verifier enforces it, but #697 wants the opposite for ingress and item 2 of #659 asks for it in the outbound sense. three readings, and it needs the maintainers rather than the two of us.

one thing from your side: #742 landed on the 22nd and changed `IssueToken` to take `TokenOptions` with a per-call `RequestedValidity`. this package still has a fixed lifetime set at construction, which is fine while nothing here implements `IdentityProvider`, but the wiring will need to honour it. leaving that until the provider is actually registered.

zmberg:
@AnshulPatil2005 Hello, I sent you an email(anshulpatil1022@gmail.com) and was wondering if you received it. I hope we can communicate via Slack.

AnshulPatil2005:
> @AnshulPatil2005 Hello, I sent you an email([anshulpatil1022@gmail.com](mailto:anshulpatil1022@gmail.com)) and was wondering if you received it. I hope we can communicate via Slack.

hi @zmberg,very sorry for the slow response. your mail went to my spam folder and i only found it just now, replying there.

zmberg:
refer: https://github.com/openkruise/agents/pull/1007 @AnshulPatil2005

PR #793, opened 2026-08-12
fix(controller): ops template patch sanitization and checkpoint resume selection

zmberg:

Two fixes for sandbox upgrades:

1. **Keep required containers from being deleted by ops template patch** (`pkg/controller/sandboxupdateops/patch.go`)
   An annotation-only `SandboxUpdateOps` patch was serialized with an empty containers slice, producing `"spec":{"containers":null}` in the strategic merge patch. The delete semantics of that field removed all existing containers from the sandbox pod template and blocked the upgrade. A new `sanitizeTemplatePatch` helper drops explicit nulls for fields a pod template can never drop before patching.

2. **Select checkpoint by ID when pod template delta is empty** (`pkg/controller/sandbox/core/checkpoint.go`)
   `GetCheckpointResumeData` only selected a checkpoint with a non-empty `PodTemplateDelta`, so checkpoints carrying a valid `checkpointId` but an empty delta were skipped and CheckpointRestore upgrades always failed with "checkpoint ID not found". A checkpoint is now selected when it carries either a delta or a checkpoint ID.

NONE

1. Run unit tests: `go test ./pkg/controller/sandboxupdateops/... ./pkg/controller/sandbox/core/...`
2. Verification scenario 1: create a `SandboxUpdateOps` with an annotation-only template patch; verify the upgrade proceeds and existing containers are preserved.
3. Verification scenario 2: upgrade a paused sandbox via CheckpointRestore whose checkpoint has an empty `PodTemplateDelta` but a valid `checkpointId`; verify the upgrade no longer fails with "checkpoint ID not found".

- `sanitizeTemplatePatch` returns the raw bytes unchanged when the patch carries no such nulls or cannot be parsed; parse errors are left to the strategic merge patch call which reports them.
- `containers` is required in a pod spec, so a null value can never be a legitimate delete request — it only appears when an empty PodSpec is marshaled into the patch.
- A sandbox holds at most one active checkpoint at a time, so selecting by "delta or ID" still resolves to a single checkpoint.

### DIFF
--- a/pkg/controller/sandboxupdateops/patch.go
+++ b/pkg/controller/sandboxupdateops/patch.go
@@ -17,6 +17,7 @@ limitations under the License.
 package sandboxupdateops
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -38,9 +39,13 @@ import (
 // containers of the sandbox template. The raw bytes are returned unchanged
 // when the patch does not carry such nulls or cannot be parsed (parse errors
 // are left to the strategic merge patch call, which reports them).
+// Numbers are decoded with UseNumber so the round trip keeps int64 values
+// (e.g. activeDeadlineSeconds, runAsUser) beyond float64 precision intact.
 func sanitizeTemplatePatch(raw []byte) []byte {
 	var obj map[string]interface{}
-	if err := json.Unmarshal(raw, &obj); err != nil {
+	dec := json.NewDecoder(bytes.NewReader(raw))
+	dec.UseNumber()
+	if err := dec.Decode(&obj); err != nil {
 		return raw
 	}
 	spec, ok := obj["spec"].(map[string]interface{})
--- a/pkg/controller/sandboxupdateops/patch.go
+++ b/pkg/controller/sandboxupdateops/patch.go
@@ -30,6 +30,41 @@ import (
 	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
 )
 
+// sanitizeTemplatePatch strips explicit nulls for fields a pod template can
+// never drop, so they cannot act as strategic-merge-patch delete directives.
+// Marshaling a PodTemplateSpec whose PodSpec is empty emits
+// "spec":{"containers":null} because PodSpec.Containers has no omitempty tag;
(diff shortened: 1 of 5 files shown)


pkg/controller/sandboxupdateops/patch.go
```diff
@@ -30,6 +30,41 @@ import (
 	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
 )
 
+// sanitizeTemplatePatch strips explicit nulls for fields a pod template can
+// never drop, so they cannot act as strategic-merge-patch delete directives.
+// Marshaling a PodTemplateSpec whose PodSpec is empty emits
+// "spec":{"containers":null} because PodSpec.Containers has no omitempty tag;
```
HARSHRAJ2789, line 37: Decoding into `map[string]interface{}` makes every JSON number a `float64`, so re-marshaling truncates past 2^53. Running the function as written:

```
in : {"spec":{"containers":null,"activeDeadlineSeconds":9007199254740993}}
out: {"spec":{"activeDeadlineSeconds":9007199254740992}}
```

Reachable via `activeDeadlineSeconds` and `runAsUser`/`runAsGroup`/`fsGroup`. A `json.Decoder` with `UseNumber()` keeps the literal and leaves the rest of the function unchanged.

pkg/controller/sandboxupdateops/patch.go
```diff
@@ -65,7 +105,7 @@ func mergeTemplate(modified *agentsv1alpha1.Sandbox, ops *agentsv1alpha1.Sandbox
 	if err != nil {
 		return fmt.Errorf("failed to marshal original template: %w", err)
 	}
-	mergedBytes, err := strategicpatch.StrategicMergePatch(originalBytes, ops.Spec.Patch.Raw, &v1.PodTemplateSpec{})
+	mergedBytes, err := strategicpatch.StrategicMergePatch(originalBytes, sanitizeTemplatePatch(ops.Spec.Patch.Raw), &v1.PodTemplateSpec{})
```
furykerry, line 108: can we validate the result sandbox before applying to apiserver ? if the patch is not valid semantically, we can capture such patch and expose information in the sandboxupdateops

PR #796, opened 2026-08-12
fix(gateway): refresh the OIDC verifier so provider key rotation is picked up

nishantbkl3345-ship-it:

`oidc.NewVerifier` builds an **immutable** JWKS snapshot, and `jwtauth.Manager.loadUntilReady` published that snapshot exactly once and then blocked on `<-ctx.Done()` for the rest of the process lifetime:

```go
verifier, err := m.loader(ctx, reader, options)
if err == nil && !isNilVerifier(verifier) {
    m.verifier.Store(&verifierHolder{verifier: verifier})
    ...
    <-ctx.Done()   // never reloaded
    return nil
}
```

The verifier is registered as a long-lived `manager.Runnable` in the sandbox-gateway process (`cmd/sandbox-gateway/main.go` → `controller.StartManager` → `addJWTAuthManager`), so the key set it loaded at startup was the only key set it would ever use.

Once the identity provider rotates its signing key, every newly-issued traffic access token references a `kid` that is not in the startup snapshot and is rejected by `verifier.Verify` with `token references unknown kid`. Recovery requires restarting every gateway pod — routine, expected key rotation turns into an authentication outage.

This PR keeps rebuilding the verifier on a refresh interval (default 10m) after the first successful load and hot-swaps it through the `atomic.Pointer` that `Current()` already reads lock-free, so the request path is unaffected.

Behaviour on a failed refresh is deliberate: the last published verifier **stays in place**, the manager **stays `Ready`**, and the refresh is retried with the existing backoff. Stale keys still verify unrotated tokens, so a provider blip must never revoke authentication that currently works.

NONE

```bash
go test ./pkg/sandbox-gateway/jwtauth/... -race -count=1
```

Two layers of tests were added, and **both were confirmed to fail against the pre-fix code** (by restoring the old publish-once-then-block behaviour) and pass with the fix:

1. `TestManagerKeyRotation` — end-to-end against a real `httptest` TLS JWKS endpoint using the production `oidc.NewVerifier` loader. It asserts the actual outage: a token signed with the not-yet-published key is first rejected with `unknown kid`, then the server rotates its key set, and the manager's verifier must accept the rotated token without a restart. Pre-fix result:

   ```
   --- FAIL: TestManagerKeyRotation/rotated_signing_key_is_accepted_without_a_restart
       Messages: verifier never picked up the rotated signing key
   ```

2. `TestManagerRefresh` — table-driven coverage of the manager's refresh contract: rotated keys are published; a failing refresh and a nil-verifier refresh both keep the last good verifier and keep the manager `Ready`; a non-positive interval disables refresh.

Also run: `go build ./...`, `go vet ./pkg/sandbox-gateway/... ./pkg/identity/...`, and the full `pkg/sandbox-gateway/...` + `pkg/identity/...` suites with `-race`. The timing-sensitive package was rerun with `-count=15 -race` to check for flakiness.

- **No behaviour change on the request path.** Publication still goes through the existing `atomic.Pointer`; `Current()` is untouched.
- The refresh interval is a package constant (`defaultRefreshInterval = 10 * time.Minute`) rather than new configuration, to keep this a focused bug fix. `NewManagerWithDependencies` keeps its existing signature and delegates to an unexported `newManager` that takes the interval, which is the seam the tests use. Happy to promote it to an `OIDC_*` env var if reviewers prefer it configurable.
- `loadUntilReady` was factored into `load`/`publish`/`sleep` helpers so the initial-load and refresh loops share one code path; the nil-loader and nil-verifier errors reported through `Ready()` are unchanged.
- One existing assertion message in `TestManagerRetryAndPublication` was reworded (`loader must not refresh after readiness` → `...before the refresh interval elapses`). The assertion itself is unchanged and still passes, since the production interval is far longer than that test's window.

### DIFF
--- a/pkg/sandbox-gateway/jwtauth/manager.go
+++ b/pkg/sandbox-gateway/jwtauth/manager.go
@@ -14,7 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package jwtauth manages process-wide initialization of the gateway JWT verifier.
+// Package jwtauth manages process-wide initialization and refresh of the
+// gateway JWT verifier.
 package jwtauth
 
 import (
@@ -35,6 +36,11 @@ import (
 const (
 	defaultInitialBackoff = time.Second
 	defaultMaxBackoff     = 30 * time.Second
+	// defaultRefreshInterval is how often a published verifier is rebuilt from
+	// the identity provider. A verifier holds an immutable JWKS snapshot, so
+	// without periodic refresh every token signed with a rotated key is
+	// rejected as an unknown kid until the process restarts.
+	defaultRefreshInterval = 10 * time.Minute
 )
 
 var (
@@ -68,13 +74,15 @@ type verifierHolder struct {
 	verifier oidc.Verifier
 }
 
-// Manager coordinates one-time, process-level OIDC verifier initialization.
+// Manager coordinates process-level OIDC verifier initialization and keeps the
+// published verifier refreshed so that provider key rotation is picked up.
 type Manager struct {
-	optionsSource  OptionsSource
-	loader         VerifierLoader
-	initialBackoff time.Duration
-	maxBackoff     time.Duration
-	wake           chan struct{}
+	optionsSource   OptionsSource
+	loader          VerifierLoader
+	initialBackoff  time.Duration
+	maxBackoff      time.Duration
+	refreshInterval time.Duration
+	wake            chan struct{}
 
 	state    atomic.Uint32
 	verifier atomic.Pointer[verifierHolder]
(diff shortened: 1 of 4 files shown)


nishantbkl3345-ship-it:
The earlier `golangci-lint` and `unit-tests` failures on this PR were **not** caused by these changes — `master` itself did not compile:

```
pkg/sandbox-manager/infra/sandboxcr/network_test.go:383:15: undefined: utils (typecheck)
```

Reproduced on `master` (9f59342) with no changes applied: #686 replaced `utils.GetSandboxID` with the `sandboxid` package, and #740 landed a call site written before that rename.

I opened #798 with the one-line fix, and rebased this branch onto it, so this PR is currently stacked on that commit. Once #798 merges, this PR will contain only its own commit. Happy to rebase again at any point.

PR #803, opened 2026-08-12
fix(sandboxset): sort old candidates by scale-down priority before deletion

### LINKED ISSUE #802: [BUG] SandboxSet controller fails to sort oldCandidates before scale down
**What happened**:
In the `SandboxSet` controller, when scaling down sandboxes, the controller correctly divides the candidates into `oldCandidates` (older template revision) and `updatedCandidates` (newer template revision). 

However, when picking which `oldCandidates` to delete first (if the required scale down count is less than the total number of old sandboxes), the controller simply takes a slice without sorting:
```go
	// Phase 1: Delete old revision sandboxes first
	oldToDelete := oldCandidates[:min(count, len(oldCandidates))]
```
Because the `oldCandidates` slice inherits the arbitrary order from the Kubernetes API list response, the controller ignores the intended scale-down priority (`Pending` > `Recycled` > `NotReady` > `Available`). This can result in deleting a perfectly healthy, `Available` older sandbox while leaving a `Pending` or `NotReady` older sandbox running.

By contrast, `updatedCandidates` is correctly sorted using `compareScaleDownPriority` before truncation:
```go
	// Phase 2: Delete updated revision sandboxes if more needed.
	// Priority: Pending > Recycled (recycledCount desc) > Running-NotReady > Available fresh.
	slices.SortFunc(updatedCandidates, compareScaleDownPriority)
	updatedToDelete := updatedCandidates[:min(remaining, len(updatedCandidates))]
```

The `oldCandidates` slice should be sorted using `slices.SortFunc(oldCandidates, compareScaleDownPriority)` before taking the subset to delete, ensuring that less desirable sandboxes (like `Pending` or `NotReady`) are cleaned up first.

1. Create a SandboxSet with multiple replicas.
2. Update the SandboxSet template to trigger a rolling update, but keep some old sandboxes running.
3. Scale down the SandboxSet such that only a fraction of the `oldCandidates` needs to be deleted (i.e. `count < len(oldCandidates)`).
4. Observe that the older sandboxes deleted are chosen arbitrarily, potentially killing `Available` sandboxes while leaving `Pending`/`NotReady` ones behind.

The fix is straightforward. In `pkg/controller/sandboxset/sandboxset_controller.go`:
```go
	// Phase 1: Delete old revision sandboxes first
+	slices.SortFunc(oldCandidates, compareScaleDownPriority)
	oldToDelete := oldCandidates[:min(count, len(oldCandidates))]
```

**Environment**:
- OpenKruiseAgent version: latest (`main` branch)
- Kubernetes version (use `kubectl version`): N/A
- Install details (e.g. helm install args): N/A
- Others: N/A

vishalmore90:

This PR fixes a bug in the `SandboxSet` controller where `oldCandidates` were not being sorted by scale-down priority prior to deletion. 

Previously, when scaling down sandboxes of an older template revision, the controller would simply slice the first `N` elements from the unsorted list of candidates. This caused an arbitrary deletion order that ignored the intended priority (`Pending` > `Recycled` > `NotReady` > `Available`), potentially deleting perfectly healthy, available sandboxes while leaving broken or pending sandboxes running.

This patch applies `slices.SortFunc(oldCandidates, compareScaleDownPriority)` before truncating the slice, ensuring that less desirable older sandboxes are always deleted first.

fixes #802

- A new unit test `TestReconcile_ScaleDown_OldCandidates_Priority` has been added to `sandboxset_controller_test.go` to explicitly cover the priority sorting logic for older sandbox revisions.
- Run `go test -v -count=1 ./pkg/controller/sandboxset` to confirm the controller respects the scale-down priority rules for old candidates.

- The PR is a minimal, single-line logic fix that aligns the treatment of `oldCandidates` with the existing (correct) treatment of `updatedCandidates`. No new priority logic was introduced.

### PR Checklist
- [x] Code follows the repository's existing patterns and style guidelines.
- [x] Test cases added to cover the bug and prevent regressions.
- [x] All unit tests pass successfully.
- [x] The commit message and PR title are descriptive and properly formatted.

### DIFF
--- a/pkg/controller/sandboxset/sandboxset_controller.go
+++ b/pkg/controller/sandboxset/sandboxset_controller.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"math"
 	"reflect"
-	"slices"
 	"time"
 
 	"github.com/google/uuid"
@@ -275,8 +274,12 @@ func (r *Reconciler) scaleDown(ctx context.Context, count int, sbs *agentsv1alph
 		return err
 	}
 
+	lessByScaleDownPriority := func(i, j *agentsv1alpha1.Sandbox) bool {
+		return compareScaleDownPriority(i, j) < 0
+	}
+
 	// Phase 1: Delete old revision sandboxes first
-	oldToDelete := oldCandidates[:min(count, len(oldCandidates))]
+	oldToDelete := findOldestSandboxes(oldCandidates, count, lessByScaleDownPriority)
 	var totalSuccesses int
 	successes, err := utils.DoItSlowlyWithInputs(oldToDelete, initialBatchSize, deleteFunc)
 	totalSuccesses += successes
@@ -292,8 +295,7 @@ func (r *Reconciler) scaleDown(ctx context.Context, count int, sbs *agentsv1alph
 
 	// Phase 2: Delete updated revision sandboxes if more needed.
 	// Priority: Pending > Recycled (recycledCount desc) > Running-NotReady > Available fresh.
-	slices.SortFunc(updatedCandidates, compareScaleDownPriority)
-	updatedToDelete := updatedCandidates[:min(remaining, len(updatedCandidates))]
+	updatedToDelete := findOldestSandboxes(updatedCandidates, remaining, lessByScaleDownPriority)
 	successes, err = utils.DoItSlowlyWithInputs(updatedToDelete, initialBatchSize, deleteFunc)
 	totalSuccesses += successes
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/sandboxset/sandboxset_controller.go
```diff
@@ -276,6 +276,7 @@ func (r *Reconciler) scaleDown(ctx context.Context, count int, sbs *agentsv1alph
 	}
 
 	// Phase 1: Delete old revision sandboxes first
+	slices.SortFunc(oldCandidates, compareScaleDownPriority)
```
furykerry: consider using findOldestSandboxes helper func to merge sort and top-k selection, the sort in phase 2 can also be changed.

vishalmore90: Thank you for the great suggestion, I've updated both Phase 1 and Phase 2 to use the findOldestSandboxes helper with a custom less function (compareScaleDownPriority(i, j) < 0). This successfully merges the sorting and top-k selection and improves performance O(N log K)

vishalmore90:
@furykerry 
/review

PR #818, opened 2026-08-13
fix(proxy): ignore reserved keys in the untrusted header modifier

nishantbkl3345-ship-it:
## What

`headerModifiers` in the ext-proc data plane reads the `request-header-modifier`
request header, unmarshals it as a JSON object, and emits every key/value as an
ext-proc `SetHeaders` mutation. That header arrives on the downstream request, so
its content is client input, but the mutations built from it were unrestricted.

Two things make those mutations privileged in the shipped configuration
(`config/sandbox-manager/envoy-config.yaml`):

- `mutation_rules: allow_envoy: true`, so this filter may set `x-envoy-*` headers
  that Envoy strips from untrusted downstream requests
- `sbx_original_dst_cluster` is `ORIGINAL_DST` with `use_http_header: true`, so
  the upstream is resolved from `x-envoy-original-dst-host` — the same header
  `handleRequestHeaders` sets from the resolved route

So a client could ask for the routing header itself and reach an address the
proxy never selected, bypassing the sandbox-ID-to-IP mapping and the
`State != Running` check; forge other `x-envoy-*` headers; or replace headers the
adapter derived, such as a rewritten `:path`. The modifier mutations were also
appended after the route-derived ones, leaving the effective value up to Envoy's
set-header semantics.

Nothing in the repository sets `request-header-modifier`, and it had no test
coverage, so the only source reaching this code is the client request.

The standalone gateway data plane (`pkg/sandbox-gateway/filter`) has no
equivalent path and is unaffected.

## How

`headerModifiers` now rejects reserved keys while building the mutations:

- headers this filter already derived from the route or the adapter
- HTTP/2 pseudo-headers (`:path`, `:authority`, ...)
- `Host`
- the `x-envoy-` namespace

Any other key keeps working exactly as before. Rejected keys are logged.

## Testing

`TestHandleRequestHeaders_RequestHeaderModifier` covers routing-header override,
other `x-envoy-*` keys, pseudo-headers, `Host`, adapter-derived headers, a benign
key, malformed JSON, and the absent header. Five of its cases fail on the current
code and pass with this change.

```
go test ./pkg/proxy/... -count=1 -race     # ok
go test ./pkg/sandbox-gateway/... ./pkg/sandboxroute/... -count=1   # ok
```

### DIFF
--- a/pkg/proxy/ext_proc.go
+++ b/pkg/proxy/ext_proc.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 
 	configPb "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	extProcPb "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
@@ -93,6 +94,14 @@ func (s *Server) Process(srv extProcPb.ExternalProcessor_ProcessServer) error {
 }
 
 var OrigDstHeader = "x-envoy-original-dst-host"
+
+const (
+	// requestHeaderModifier is the request header carrying extra header mutations
+	// as a JSON object. Its content is downstream input and never authoritative.
+	requestHeaderModifier = "request-header-modifier"
+	// envoyHeaderPrefix is the header namespace Envoy reserves for itself.
+	envoyHeaderPrefix = "x-envoy-"
+)
 
 func (s *Server) handleRequestHeaders(requestHeaders *extProcPb.ProcessingRequest_RequestHeaders, log logr.Logger) *extProcPb.ProcessingResponse {
 	// Step 1: Convert ext_proc headers to flat map[string]string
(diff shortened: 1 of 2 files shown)


nishantbkl3345-ship-it:
The failing quota Redis E2E check is unrelated to this PR.

This PR only changes `pkg/proxy/ext_proc.go` and its tests, while the failure is in the quota Redis E2E test:

```
test_redis_data_loss_rebuilds_from_live_crs_and_reenforces
AssertionError: count[all]=0, expected=2
```

The same failure occurs on master and unrelated Dependabot branches, and is tracked in #791.

Every other check passes at this PR HEAD (`ea19e61`), including the proxy package tests with race detection. The only other non-passing item is `tide`, which is pending the `approved`/`lgtm` labels rather than reporting a failure.

A separate root-cause fix for the quota rollup repair issue is being prepared independently rather than mixing it into this proxy change.

PR #820, opened 2026-08-14
fix(webhook): webhook registration for sandboxtemplate

googs1025:

This PR fixes SandboxTemplate admission webhook registration and validation:

- register SandboxTemplate mutating and validating webhook handlers at runtime
- return a validation error when `spec.template` is missing instead of panicking
- allow SandboxTemplate `volumeClaimTemplates` to satisfy pod volume mounts during validation
- add regression tests for missing `spec.template` and mounted volume claim templates

NONE

```shell
go test ./pkg/webhook/...
```

N/A

### DIFF
--- a/pkg/webhook/sandboxtemplate/validating/sandboxtemplate_create_update.go
+++ b/pkg/webhook/sandboxtemplate/validating/sandboxtemplate_create_update.go
@@ -22,7 +22,6 @@ import (
 	"net/http"
 	"strings"
 
-	apicorev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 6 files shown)


pkg/webhook/sandboxtemplate/validating/sandboxtemplate_create_update.go
```diff
@@ -102,6 +106,53 @@ func validateSandboxPodTemplateSpec(spec agentsv1alpha1.SandboxTemplateSpec, fld
 		errList = append(errList, field.Invalid(fldPath.Child("template"), spec.Template, fmt.Sprintf("Convert_v1_PodTemplateSpec_To_core_PodTemplateSpec failed: %v", err)))
 		return errList
 	}
+	if len(spec.VolumeClaimTemplates) != 0 {
```
furykerry: Issue: validateVolumeClaimTemplateMounts and the synthetic-volume injection loop are copied character-for-character (only the spec type differs). Any future fix (e.g., handling subPath semantics, new container kinds, duplicate-name collision messages) must now be applied in two places, and drift is guaranteed.

Recommendation: Extract a policy-neutral helper into pkg/webhook/utils that operates on *corev1.PodTemplateSpec + []corev1.PersistentVolumeClaim

furykerry (approved):
currently sandboxtemplate only validate create operation, the user can bypass the validation by create and update the sandboxtemplate later, as a TODO, consider add update validating for sandboxtemplate, sandboxtemplate spec should be immutable after creation

PR #828, opened 2026-08-15
Add SandboxUpdateOps metrics

googs1025:

Add Prometheus metrics for SandboxUpdateOps reconciliation:

- `sandbox_updateops_status_phase{namespace,name,phase}` reports the active phase as `1`.
- `sandbox_updateops_replicas{namespace,name,type}` reports total, updated, updating, and failed replica counts from status.
- `sandbox_updateops_total{namespace,result}` counts terminal results by namespace with `completed`, `failed`, and `zero_target`.

The terminal result counter is recorded from observed terminal status with a per-UID in-process guard, so it can still count objects first seen after restart or leader handoff.

NONE

```bash
go test ./pkg/controller/sandboxupdateops -count=1
git diff --check
```

Gauge series are cleaned up when the SandboxUpdateOps finalizer cleanup completes. The historical terminal-result counter is intentionally retained.

### DIFF
--- a/pkg/controller/sandboxupdateops/metrics.go
+++ b/pkg/controller/sandboxupdateops/metrics.go
@@ -1,0 +1,151 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sandboxupdateops
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
+)
+
+const (
+	updateOpsReplicaTypeTotal    = "total"
+	updateOpsReplicaTypeUpdated  = "updated"
+	updateOpsReplicaTypeUpdating = "updating"
+	updateOpsReplicaTypeFailed   = "failed"
+
+	updateOpsResultCompleted  = "completed"
+	updateOpsResultFailed     = "failed"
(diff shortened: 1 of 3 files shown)


HARSHRAJ2789 (commented):
Read this against master at `43765ba`, next to the sandboxclaim metrics it mirrors.

The shape matches what is already there. Metric names, the `namespace/name/phase` gauge, the
per-type replicas gauge and the terminal counter all line up with
`pkg/controller/sandboxclaim/metrics.go`, and clearing the other phases before setting the
current one is the same move `sandboxClaimStatusPhase` makes, so a phase gauge never reports
two live phases at once. The `sync.Map` guard keyed on UID is a good call: `updateStatus` runs
on repeated reconciles and a counter that increments per reconcile would be unusable.

The cleanup path is where the two controllers diverge, and I think this one has a gap.

`deleteSandboxUpdateOpsMetrics` is reachable only from `handleDeletion`, after the finalizer
update succeeds at `sandboxupdateops_controller.go:605-608`. Reconcile's own miss path returns
at `:106-108` with a bare `client.IgnoreNotFound(err)` and clears nothing. sandboxclaim covers
exactly that case: `sandboxclaim_controller.go:113-118` calls
`deleteSandboxClaimMetrics(req.Namespace, req.Name)` on `IsNotFound` before returning.

So any deletion the finalizer path does not observe leaves the phase and replica gauges published for a
SandboxUpdateOps that no longer exists, and leaves the `observedUpdateOpsResults` entry behind
for the process lifetime. A controller restart during deletion, or a finalizer removed by
hand, both land there.

The fix is not quite the sandboxclaim one-liner, which is worth saying because it is the part
that would bite. `deleteSandboxUpdateOpsMetrics` takes the whole object, and
`updateOpsResultKey` keys on `namespace/name/UID` whenever the UID is set. On the NotFound
path only `req.NamespacedName` is available, so a namespace-and-name cleanup would clear the
gauges and miss the `sync.Map` entry, which is the one that never expires. Splitting the gauge
cleanup, which needs only namespace and name, from the result-key cleanup, which needs the
UID, would let the NotFound path do the part it can.

Two smaller notes.

`recordSandboxUpdateOpsMetrics` runs immediately after the Get and before the concurrency
check, so a
SandboxUpdateOps that is about to be skipped still publishes its gauges. That reads
deliberate, since the object does exist and its status is real, and sandboxclaim records at
the same point. Worth confirming it is deliberate.

`recordSandboxUpdateOpsStatusMetrics` guards the phase gauge with `if status.Phase != ""` but
sets the replicas gauges unconditionally, so a SandboxUpdateOps observed before its first
status write publishes four zeros with no phase. Harmless on a dashboard, though it does mean
`sandbox_updateops_replicas` appears slightly before `sandbox_updateops_status_phase` for
every new object.

PR #829, opened 2026-08-17
feat(api): add headerManipulation action to security rules

delavet:

Add a non-terminal `HeaderManipulationAction` to SecurityRuleActions so a security rule can set or remove plaintext request headers on matching egress traffic. Header names are restricted to the same safe tchar subset as AuditHeader.Name and values are stored verbatim, so TokenTransformation with a CredentialRef remains the only path for credentials. Includes regenerated deepcopy and CRD manifests.


api/v1alpha1/securityprofile_types.go
```diff
@@ -520,6 +525,44 @@ type SecurityRuleActions struct {
...
+//
+// +kubebuilder:validation:XValidation:rule="has(self.set) || has(self.remove)",message="at least one of set or remove must be specified"
+type HeaderManipulationAction struct {
+	// Set adds or replaces headers. An existing header with the same
+	// name is replaced.
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=16
+	Set []HeaderValue `json:"set,omitempty"`
```
Kuromesi: Do we also need an add action?

api/v1alpha1/securityprofile_types.go
```diff
@@ -520,6 +525,44 @@ type SecurityRuleActions struct {
...
+	// +kubebuilder:validation:MaxItems=16
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=256
+	// +kubebuilder:validation:items:Pattern=`^[A-Za-z0-9!#$%&'*+\-.^_|~]+$`
+	Remove []string `json:"remove,omitempty"`
+}
+
+// HeaderValue is one plaintext header assignment.
+type HeaderValue struct {
+	// Name is the header name (case-insensitive). Restricted to a safe
```
furykerry: Name is documented as case-insensitive, but +listType=map +listMapKey=name gives the API server case-sensitive key uniqueness.

Fix: For a brand-new field, require lowercase. O(1) per item, no CEL cost, and it makes the map key match HTTP semantics exactly:

api/v1alpha1/securityprofile_types.go
```diff
@@ -511,6 +511,11 @@ type SecurityRuleActions struct {
 	// denied.
 	// +optional
 	MCPToolPolicy *MCPToolPolicySpec `json:"mcpToolPolicy,omitempty"`
+	// HeaderManipulation sets or removes plaintext request headers.
+	// Non-terminal. Values are stored verbatim — use TokenTransformation
+	// for credentials.
+	// +optional
+	HeaderManipulation *HeaderManipulationAction `json:"headerManipulation,omitempty"`
```
furykerry, line 521: Issue: The accepted proposal already reserves header injection as IdentityInjectionAction / IdentityHeader{Name, ValueFrom HeaderValueSource}, and docs/components/traffic-extension.md still lists IdentityInjection as pending. This PR lands a second, differently-shaped header-injection action without amending either. 

Recommendation: Amend the proposal with a short section explaining the split (headerManipulation = static verbatim values + removal; identityInjection = pod-metadata-sourced values), or fold static values into IdentityInjection as an alternative valueFrom. Two actions that both write request headers, with different precedence slots, is a lasting API-comprehension cost.

api/v1alpha1/securityprofile_types.go
```diff
@@ -520,6 +525,43 @@ type SecurityRuleActions struct {
...
+// SECURITY: values are stored verbatim in the enclosing object. Use
+// TokenTransformation with a CredentialRef for credentials.
+type HeaderManipulationAction struct {
+	// Set adds or replaces headers. An existing header with the same
+	// name is replaced.
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=16
+	Set []HeaderValue `json:"set,omitempty"`
```
Kuromesi, line 544: Do we also need an add action?

api/v1alpha1/securityprofile_types.go
delavet: HeaderValue.Name and the remove items now require lowercase (^[a-z0-9!#$%&'*+\-.^_|~]+$), so the list-map key matches HTTP's case-insensitive header semantics exactly.

api/v1alpha1/securityprofile_types.go
delavet: Addressed in 159e73f0 by amending the proposal rather than folding into IdentityInjection, for two reasons:

1. remove has no home in IdentityInjection — folding static values in as an alternative valueFrom would still leave removal semantics as a second shape.
2. The two actions sit on different trust boundaries: headerManipulation carries static verbatim values authored in the profile (or normalized from tenant-facing inputs), while identityInjection stays reserved for platform-derived values sourced from pod metadata, and never removes.

The proposal now documents this split, adds HeaderManipulation to the fixed non-terminal order (SecurityCheck → IdentityInjection → HeaderManipulation → TokenTransformation → …) so identity stamps can't be silently overwritten by later static rules while credential injection still sees the final header set, and mirrors the lowercase name contract. docs/components/traffic-extension.md is updated so the pending-actions note reflects where the action is enforced today.

api/v1alpha1/securityprofile_types.go
delavet: Not for the current use case — the tenant-facing semantics we're normalizing from only need set-or-replace plus remove. add (appending a  duplicate header line) can be introduced later as a fully compatible optional field if a concrete need shows up, so we'd rather keep the initial surface minimal.

furykerry (commented):
/lgtm

PR #835, opened 2026-08-18
fix(e2b): harden resource ownership and key storage validation

AiRanthem:

Hardens E2B API key authorization and key-storage configuration validation:

- Require an exact caller/owner match for Sandbox and Volume access. Ownership mismatches now return the same HTTP 404 response as missing resources, preventing authenticated callers from enumerating existing IDs.
- Preserve the authentication-disabled compatibility path: resources created without authentication remain owned by `AdminKeyID`, and the canonical admin key can still access them after authentication is enabled.
- Validate key-storage configuration in `NewKeyStorage`: both backends require a non-empty admin key, while MySQL also requires a DSN and a non-empty hash pepper. This protects direct factory callers and removes duplicate validation from `main`.
- Keep the existing Secret backend dependency checks and update the CLI help to document the MySQL pepper requirement.

NONE

```bash
GOFLAGS=-mod=readonly go test ./pkg/servers/e2b ./pkg/servers/e2b/keys \
  -run 'TestCheckApiKey_(SandboxOwnership|VolumeOwnership|AnonymousUserAdminKeyCompatibility)$|TestNewKeyStorage$' \
  -count=1
```

- The canonical admin key is not a general ownership bypass; admin and regular keys must both match the persisted owner.
- Ownership mismatches are intentionally indistinguishable from missing resources. Structured logs retain the owner and caller UUIDs for diagnostics without exposing them to the client.
- Authentication-disabled mode remains unchanged: it does not construct key storage, and `AnonymousUser.ID` remains `AdminKeyID` for one-way compatibility.
- This PR does not change `/debug`, `/metrics`, list endpoints, resource creation, or Volume route availability.

### DIFF
--- a/pkg/servers/e2b/routes.go
+++ b/pkg/servers/e2b/routes.go
@@ -103,7 +103,7 @@ var AnonymousUser = &models.CreatedTeamAPIKey{
 // CheckApiKey implements common ApiKey validation
 func (sc *Controller) CheckApiKey(ctx context.Context, r *http.Request) (context.Context, *web.ApiError) {
 	logger := klog.FromContext(ctx)
-	middleWareLog := logger.WithValues("middleware", "CheckApiKey").V(utils.DebugLogLevel)
+	middleWareLog := logger.WithValues("middleware", "CheckApiKey")
 	apiKey := r.Header.Get(models.HeaderApiKey)
 	var user *models.CreatedTeamAPIKey
 	var ok bool
@@ -115,7 +115,7 @@ func (sc *Controller) CheckApiKey(ctx context.Context, r *http.Request) (context
 		rawAPIKey := keys.ToStoredRawAPIKey(apiKey)
 		user, ok = sc.keys.LoadByKey(ctx, rawAPIKey)
 		if !ok {
-			middleWareLog.Info("failed to load key by API-KEY")
+			middleWareLog.V(utils.DebugLogLevel).Info("failed to load key by API-KEY")
 			return ctx, &web.ApiError{
 				Code:    http.StatusUnauthorized,
 				Message: "Invalid API Key",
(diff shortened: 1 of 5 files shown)


pkg/servers/e2b/routes.go
```diff
@@ -132,10 +132,15 @@ func (sc *Controller) CheckApiKey(ctx context.Context, r *http.Request) (context
 				Message: fmt.Sprintf("Sandbox route not found, maybe it is crashed or killed: %s", sandboxID),
 			}
 		}
-		if owner != AnonymousUser.ID.String() && owner != user.ID.String() {
+		// An ownership mismatch returns the same not-found response as a missing route so
+		// authenticated callers cannot probe which sandbox IDs exist. That makes this log
+		// the only signal separating a denial from a genuine miss, so it carries the ID
+		// itself rather than relying on the logger's accumulated values.
+		if owner != user.ID.String() {
+			middleWareLog.Info("sandbox owner mismatch", "sandboxID", sandboxID, "owner", owner, "user", user.ID.String())
```
furykerry: middleWareLog has already config with WithValues of sandboxID, sandboxID will be logged twice here

pkg/servers/e2b/routes.go
```diff
@@ -153,10 +158,13 @@ func (sc *Controller) CheckApiKey(ctx context.Context, r *http.Request) (context
 				Message: fmt.Sprintf("Volume not found: %s", volumeID),
 			}
 		}
-		if owner != AnonymousUser.ID.String() && owner != user.ID.String() {
+		// Same anti-enumeration rule as the sandbox check above: ownership mismatch is
+		// indistinguishable from a missing volume.
+		if owner != user.ID.String() {
+			middleWareLog.Info("volume owner mismatch", "volumeID", volumeID, "owner", owner, "user", user.ID.String())
```
furykerry: middleWareLog is configed with V(utils.DebugLogLevel), Info log will not be printed

HARSHRAJ2789:
For the record, now that the fix is public: I reported the sandbox ownership check privately to
kubernetes-security@service.aliyun.com on 13 August, 13:54 UTC, following SECURITY.md, which
asks that security problems not be filed as issues. This PR fixes it.

My report covered the sandbox conditional, `routes.go:135` as it stood on `2c640aa`. The volume
path at `:156` is a second instance I had missed, and returning 404 so that an ownership
mismatch cannot be used to probe which sandbox IDs exist goes beyond anything I described.

I held it while the report was unanswered and nothing had been published. With the fix merged,
the change is plain from the diff, so staying quiet no longer protects anyone. I am not adding
exploitation detail beyond what the diff already shows.

PR #836, opened 2026-08-18
ci: add govulncheck, zizmor and Scorecard scans; enable gosec

DahuK:

Rounds out the CI security tooling to match what mature Go projects in the CNCF
ecosystem run (etcd, containerd, Envoy, Kata Containers). All of it is
report-only — nothing gates a merge.

**New workflows**

- **`govulncheck.yml`** — the only Go tool that does symbol-level reachability,
  so a module CVE is reported only when the vulnerable function is actually
  called. Complements Trivy and Dependabot (which match by version) and CodeQL
  (which does not look at dependency CVEs at all).
- **`zizmor.yml`** — static analysis of the workflows themselves: script
  injection via template expansion, cache poisoning, credential persistence,
  unpinned actions. Same tool Kata Containers uses.
- **`scorecard.yml`** — OpenSSF Scorecard supply-chain posture grading. Runs on
  the default branch and on a schedule only, because the action does not support
  PR triggers.

**Changes to existing configuration**

- **`codeql.yml`**
  - Adds the **`actions`** language, which scans `.github/workflows` for
    CI-specific vulnerability classes (artifact/cache poisoning, script
    injection, untrusted checkout in a privileged context, secret exposure).
  - Switches to the `security-extended` query suite.
  - **Enables the `local` threat model for Go.** The Go library already models
    `os.Getenv`, `os.Args` and `os.ReadFile` as `environment` / `commandargs` /
    `file` sources, and `go/command-injection` / `go/path-injection` already take
    their sources from `ActiveThreatModelSource`. Those source kinds are grouped
    under `local` and are inert unless requested, so by default CodeQL sees none
    of them. For a sandbox runtime this matters a lot: untrusted values arrive
    through the container spec, the environment and mount points, not over HTTP.
  - Adds `merge_group` and a weekly schedule; excludes `test`/`examples`/
    `vendor`/`third_party`.
- **`dependabot.yaml`** — adds the **`gomod`** ecosystem. Without it Dependabot
  never raised a PR for a Go dependency, so CVEs in `go.mod` had to be fixed by
  hand. Grouped by `k8s.io/*`, `go.opentelemetry.io/*` and `golang.org/x/*` to
  keep PR volume manageable.
- **`.golangci.yml`** — enables **`gosec`**. It covers what CodeQL's Go queries
  do not: file and directory permissions, weak TLS configuration, integer
  overflow conversions, weak randomness.
- **`ci.yaml`** — sets `only-new-issues: true` on golangci-lint so the
  pre-existing findings do not block unrelated PRs, and lists
  `contents`/`pull-requests` in the `golangci-lint` job permissions block. Job
  level `permissions` *replaces* the workflow level `read-all` rather than
  merging with it, so those scopes were previously being reduced to `none`.

NONE

**1. Every workflow was verified green on a fork before opening this PR**

All 13 workflow runs completed successfully on `DahuK/agents`. New jobs and
their runtimes: CodeQL 3m59s (go + actions), govulncheck 1m26s, zizmor 27s.
Because they run in parallel and the existing E2E suite already takes 20–27
minutes, this adds nothing to the critical path.

**2. The scanners were verified to actually fire, using deliberately vulnerable fixtures**

A clean PR cannot demonstrate that a scanner works. On a throwaway fork branch I
added a fixture file split into two groups — one where taint enters from a
*remote* source (`*http.Request`) and one where it enters from a *local* source
(`os.Getenv`, `os.Args`) — plus an intentionally insecure workflow. Results:

| scanner | findings on the fixtures |
| --- | --- |
| CodeQL Go, default threat model | 12 — all in the remote group, **zero** in the local group |
| CodeQL Go, `threat-models: local` | 15 — the two local-taint cases now reported |
| CodeQL `actions` | 7 |
| zizmor | 25 |
| gosec | 17 across 15 rules, 6 HIGH |

The same `exec.Command` sink was reported when the value came from
`r.URL.Query()` and silent when it came from `os.Getenv()`. That is the whole
justification for the `local` threat model change, and for running gosec
alongside CodeQL.

**3. The noise cost of the `local` threat model was measured before enabling it**

Using CodeQL CLI 2.26.3, one database built from this tree, `go-security-extended`
run twice with the only difference being `--threat-model=local`:

| findings | default | local | delta |
| --- | --- | --- | --- |
| first-party code | 1 | 41 | **+40** |

72% of the delta is concentrated in two certificate-writing helpers
(`atomic_writer.go` 21, `writer/fs.go` 7) whose whole job is to write to
configured paths, so most of that cluster should triage as by-design. The
remainder is worth a look — in particular `os.Symlink` / `os.Stat` on a
caller-supplied target in `storage-cli/link/symlink.go`.

<h3 data-heading="Ⅳ. Special notes for reviews">Ⅳ. Special notes for reviews</h3>
<p><strong>A green check does not mean zero findings — please read the Security tab.</strong><br>
<code>govulncheck -format sarif</code> exits 0 even when vulnerabilities are found, and<br>
zizmor deliberately does not use its non-zero exit codes in SARIF mode. This is<br>
intentional: the repository has a backlog, and gating on it would red-flag every<br>
unrelated PR. Findings are published to code scanning instead. Nothing is<br>
suppressed; the intent is not to hide findings, only to avoid blocking work.</p>
<p><strong>Expect a one-time triage backlog in the Security tab, not PR spam.</strong> Code<br>
scanning only annotates a PR with alerts whose lines fall inside the diff. This<br>
was verified empirically: 45 pre-existing zizmor findings produced exactly <strong>one</strong><br>
PR comment. The ~40 new CodeQL alerts and the existing zizmor findings will<br>
appear in the Security tab and only resurface on a PR that touches those lines.</p>
<p><strong>The <code>local</code> threat model contradicts the current GitHub docs.</strong> The docs state<br>
threat models are "supported only by analysis for Java/Kotlin and C#" during<br>
public preview. The Go library CHANGELOG has said otherwise since lib 1.1.2, and<br>
the fixture run confirms Go works. The docs appear stale — flagging it so a<br>
reviewer is not surprised by the discrepancy.</p>
<p><strong>Scorecard will not work on forks.</strong> The action's README states this explicitly.<br>
It also only supports the default-branch <code>push</code> and <code>schedule</code> triggers, so it<br>
will not run on this PR. Set <code>publish_results: false</code> if publishing the score to<br>
the OpenSSF API is not wanted.</p>

<p><strong>One caveat worth knowing about <code>#nosec</code>.</strong> It does not take effect when the<br>
flagged call sits in an <code>if err := f(); err != nil</code> init statement — verified<br>
with a minimal reproducer. <code>storage-cli/link/symlink.go:55</code> carries such an<br>
annotation and is still reported. There are 58 <code>#nosec</code> annotations in the tree,<br>
so some existing suppressions may be silently ineffective. Not addressed here.</p>
<p><strong>Not included in this PR:</strong> the vulnerable fixtures used for verification.<br>
They live on a separate throwaway branch and must never reach a default branch —<br>
the fixture workflow uses <code>pull_request_target</code>, which is inert on a topic<br>
branch but would become a real pwn-request vector if merged.</p>

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -6,6 +6,32 @@
 # Configuration options: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 version: 2
 updates:
+# Go module updates. Without this entry Dependabot raises no PRs for Go
+# dependencies at all, which means CVEs in go.mod are never auto-remediated
+# and govulncheck/Trivy findings have to be fixed by hand.
+- package-ecosystem: "gomod"
+  directory: "/"
+  open-pull-requests-limit: 10
+  schedule:
+    interval: "weekly"
+  groups:
+    # Keep the PR volume manageable: k8s and otel release many modules in
+    # lockstep and must be bumped together anyway.
+    k8s:
+      patterns:
+        - "k8s.io/*"
+        - "sigs.k8s.io/*"
+    otel:
+      patterns:
+        - "go.opentelemetry.io/*"
+    golang-x:
+      patterns:
+        - "golang.org/x/*"
+  # NOTE: intentionally no `ignore:` rules here. Dependabot ignore conditions
+  # apply to security updates as well as version updates, so silencing a
+  # dependency also silences its CVE PRs. If the PR volume is too high, prefer
+  # widening the `groups` above or lowering open-pull-requests-limit.
+
 - package-ecosystem: "github-actions"
   directory: "/"
   # Allow up to 10 open pull requests for update github-actions
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,7 +38,14 @@ jobs:
   golangci-lint:
     runs-on: ubuntu-24.04
     permissions:
+      # Job-level permissions REPLACE the workflow-level `read-all`, they do not
+      # merge with it. Any scope omitted here becomes `none`, so contents and
+      # pull-requests must be listed explicitly.
+      contents: read
       security-events: write
+      # Required by golangci-lint-action's only-new-issues, which fetches the
+      # PR diff from the API.
+      pull-requests: read
     steps:
       - name: Checkout Code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -64,6 +71,10 @@ jobs:
           args: --verbose
           skip-pkg-cache: true
           mod: readonly
+          # Report only issues introduced by this PR. This is what makes it safe
+          # to enable gosec without the pre-existing findings blocking every
+          # unrelated pull request.
+          only-new-issues: true
       - name: Run Trivy vulnerability scanner in repo mode
         uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # master
         with:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -38,11 +38,27 @@ jobs:
           # Go does NOT support build-mode 'none'; it needs autobuild or manual.
           - language: go
             build-mode: autobuild
+            # Extend the threat model to LOCAL sources of untrusted data. The Go
+            # library already models os.Getenv, os.Args, os.ReadFile and friends
+            # as `environment` / `commandargs` / `file` sources, and those kinds
+            # are grouped under `local` in
+            # shared/threat-models/ext/threat-model-grouping.model.yml. They are
+            # inert by default because only the `default` group (== `remote`) is
+            # enabled.
+            #
+            # This matters for a sandbox runtime: untrusted values arrive through
+            # the container spec, environment and mount points, not over HTTP.
+            # Without this, `os.Getenv() -> exec.Command()` is not reported at
+            # all, which we verified empirically against hack/vulnfixtures.
+            #
+            # Extending is additive - the default (remote) model still applies.
+            extra-config: "threat-models: local"
           # Scans .github/workflows/*.yml and **/action.yml for CI-specific
           # vulnerabilities: artifact/cache poisoning, script injection,
           # untrusted checkout in privileged context, secret exposure, TOCTOU.
           - language: actions
             build-mode: none
+            extra-config: ""
 
     steps:
       - name: Checkout repository
@@ -75,6 +91,7 @@ jobs:
               - examples
               - vendor
               - third_party
+            ${{ matrix.extra-config }}
 
       - name: Autobuild
         if: matrix.build-mode == 'autobuild'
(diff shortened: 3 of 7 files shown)


HARSHRAJ2789 (commented):
I checked this against a local checkout, because most of the claims here are the kind that are
easy to state and hard to verify. They hold up. Every action in the three
new workflows is SHA-pinned with a version comment, job-level `permissions` list `contents: read`
explicitly, which the workflow default would not have given them, checkout sets `persist-credentials:
false`, and `setup-go` disables the cache with the cache-poisoning rationale written next to it.

The fork handling is better than I expected to find. `govulncheck.yml` guards the SARIF upload on
`head.repo.full_name == github.repository` and falls back to `$GITHUB_STEP_SUMMARY`, and
`zizmor.yml` goes further by flipping `advanced-security` and `annotations` in opposite
directions so a fork pull request still gets inline output. I went looking for the usual
fork-PR-cannot-write-security-events failure and did not find it.

I also reproduced your `#nosec` finding exactly. `storage-cli/link/symlink.go:55` carries
`// #nosec G301` inside an `if err := os.MkdirAll(...)` init statement and gosec still reports
`symlink.go:55:12: G301`. The tree has exactly 58 `#nosec` annotations, so that caveat is worth
the space you gave it.

Two things to correct or reconsider.

The `.golangci.yml` comment says the current volume is 3 findings repo-wide. I measure 9. That is
`golangci-lint run ./...` with this branch's config on v2.9.0, the version `GOLANGCI_VERSION`
pins, and v2.7.2 gives the same 9. The split is G115 in `cmd/sandbox-manager/main.go`, G301 in
`storage-cli/link/symlink.go`, two G101 and one G302 in
`sandbox-gateway/runtimecredentials/credentials.go`, G402 in `traffic-extension/server/runserver.go`,
and three more under `test/e2b/assets/`. For what it is worth, 3 is exactly the `test/` subset,
which may be how the number was taken.

The conclusion drawn from it still stands, and the detail argues for it: the two G101 hits are
`envSecretNamespace` and `envSecretName`, environment variable names that contain the word
SECRET, so they are plainly false positives. At 9 findings with 2 of them noise there is still no
case for filtering by severity or confidence. Only the number needs fixing.

The other is the G104 exclusion. The reason given is that it overlaps with errcheck, but errcheck
is not enabled: `linters.default` is `none` and the enable list is copyloopvar, dupl,
ginkgolinter, goconst, gocyclo, gosec, govet, ineffassign, misspell, unconvert, unused. The only
occurrence of the string errcheck in the file is that comment. So excluding G104 leaves unhandled
errors checked by nothing at all.

That may well still be the right call, since G104 is very noisy and turning it on across this
tree would not be a small change. But the comment as written would lead a later maintainer
to believe the coverage exists somewhere else, and it does not.

One adjacent thing while you are in this file, not caused by this PR: `settings.revive` is
configured at line 34 with `comment-spacings` and `import-shadowing`, and revive is not in the
enable list either, so those rules never run.

Nothing here changes my read that this should go in. The verification section is the most
thorough I have seen on this repository, and the fixture methodology for proving the scanners
actually fire is the part I would keep.

PR #838, opened 2026-08-19
E2B inline security rules / network L7 rules support

delavet:

Builds the Sandbox-Manager half of the per-sandbox inline L7 egress rules loop (data-plane half: openkruise/agentio#18, merged, on top of the API from #829). Tenants author egress header rules through two mutually exclusive E2B inputs, which the server validates, normalizes, and persists as the `agents.kruise.io/security-rules` Sandbox annotation for the data plane to enforce under verified workload identity:

1. **Native `network.rules[domain].transform.headers`** — translated into one `headerManipulation` rule per domain (sorted domains, lowercase sorted header names, deterministic output).
2. **Reserved metadata key `e2b.agents.kruise.io/security-rules`** — a strict JSON array of `SecurityRule` (unknown fields rejected, exactly one JSON value), admitting only `block` and `headerManipulation` actions.

`PUT /sandboxes/{id}/network` replaces the rule chain with full-replacement semantics: an absent `rules` field keeps the existing chain, an explicit `{}` clears it, and a non-empty map replaces it after the same validation as creation. The annotation write (conflict-retried, no-op when unchanged) lands before the L4 TrafficPolicy widens.

NONE

Unit tests:

```bash
go test ./pkg/servers/e2b/... ./api/... ./pkg/sandbox-manager/... ./pkg/controller/sandbox/...
```

Table-driven coverage includes both input paths, PUT semantics, every rejection above, the recycle wipe (`TestResetForPool`), and annotation write/no-op/delete round trips (`TestUpdateSecurityRules`).

End to end (verified on a real cluster against the merged agentio EPE):

1. Create a sandbox with `network.rules` header transforms → the annotation carries the normalized chain; egress requests to the matched domain arrive with the header injected, and a client-supplied header of the same name is replaced.
2. Create a sandbox with the metadata key carrying `set` + `remove` → injected and stripped respectively; `block: {}` returns HTTP 403 via the defaulted statusCode.
3. `PUT /sandboxes/{id}/network`: replace → new value takes effect; absent `rules` → chain kept; `"rules": {}` → annotation removed and mutation stops; narrowing `allowOut` while keeping the chain → 400.
4. Negative requests (dual input, top-level `egressProxy`/`maskRequestHost`, whitelist violation, `Host`, uppercase names, control characters, `bypass`, out-of-range values) → all 400 with nothing written.
5. Pre-annotate a pooled Sandbox CR with a stale rule chain, then claim it without rules → the annotation is removed at claim and traffic is unaffected.

- Both data planes (annotation → EPE, TrafficPolicy → xDS) converge eventually and independently, matching the existing L4 behavior; the write ordering (rules before allowOut widening) bounds the exposure on updates, but neither path today offers a convergence signal to the caller. We need to open a follow-up issue to track convergence/observability improvements for both the L4 TrafficPolicy and the L7 rule chain together (e.g. a readiness signal or status condition), rather than solving it piecemeal in this PR.
- The whitelist-contract relaxation for CIDR/IP `allowOut` entries is a deliberate deviation from a strict literal check: a static check cannot correlate an address-based entry with a domain, so its presence disables the contract rather than producing false 400s.

### DIFF
--- a/pkg/servers/e2b/models/extensions.go
+++ b/pkg/servers/e2b/models/extensions.go
@@ -68,6 +68,9 @@ const (
 	MetadataKeySandboxResource                = v1alpha1.E2BPrefix + "sandbox-resource"
 	ExtensionKeySandboxName                   = v1alpha1.E2BPrefix + "sandbox-name"
 	ExtensionKeySandboxGenerateName           = v1alpha1.E2BPrefix + "sandbox-generate-name"
+	// ExtensionKeySecurityRules is the reserved metadata key carrying inline
+	// egress security rules as a JSON array of SecurityRule.
+	ExtensionKeySecurityRules = v1alpha1.MetadataKeySecurityRules
 )
 
 const (
@@ -102,6 +105,14 @@ func (r *NewSandboxRequest) ParseExtensions() error {
 	// parse csi mount config
 	if err := r.parseExtensionCSIMount(); err != nil {
 		return err
+	}
+	// Capture the reserved inline security-rules key. The raw value is only
+	// moved into Extensions; validation and normalization happen in the e2b
+	// server layer before the sandbox is created.
+	if raw, ok := r.Metadata[ExtensionKeySecurityRules]; ok {
+		r.Extensions.SecurityRulesRaw = raw
+		r.Extensions.SecurityRulesPresent = true
+		delete(r.Metadata, ExtensionKeySecurityRules)
 	}
 	return nil
 }
--- a/pkg/servers/e2b/models/extensions.go
+++ b/pkg/servers/e2b/models/extensions.go
@@ -19,6 +19,7 @@ package models
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 	"strings"
@@ -106,15 +107,40 @@ func (r *NewSandboxRequest) ParseExtensions() error {
 	if err := r.parseExtensionCSIMount(); err != nil {
 		return err
 	}
-	// Capture the reserved inline security-rules key. The raw value is only
-	// moved into Extensions; validation and normalization happen in the e2b
-	// server layer before the sandbox is created.
+	// Parse the reserved inline security-rules key into structured rules.
+	// Validation and normalization happen in the e2b server layer before
+	// the sandbox is created.
 	if raw, ok := r.Metadata[ExtensionKeySecurityRules]; ok {
-		r.Extensions.SecurityRulesRaw = raw
-		r.Extensions.SecurityRulesPresent = true
+		rules, err := parseInlineSecurityRules(raw)
+		if err != nil {
+			return err
+		}
+		r.Extensions.SecurityRules = rules
 		delete(r.Metadata, ExtensionKeySecurityRules)
 	}
 	return nil
+}
+
+// parseInlineSecurityRules decodes the reserved metadata value strictly:
+// unknown fields fail the request instead of being silently dropped, and the
+// value must be exactly one JSON array with nothing but whitespace after it.
+func parseInlineSecurityRules(raw string) ([]v1alpha1.SecurityRule, error) {
+	if strings.TrimSpace(raw) == "" {
+		return nil, fmt.Errorf("metadata key %q must not be empty: provide a security-rules JSON array or omit the key", ExtensionKeySecurityRules)
+	}
+	dec := json.NewDecoder(strings.NewReader(raw))
+	dec.DisallowUnknownFields()
+	var rules []v1alpha1.SecurityRule
+	if err := dec.Decode(&rules); err != nil {
+		return nil, fmt.Errorf("metadata %q is not a valid security-rules JSON array: %v", ExtensionKeySecurityRules, err)
+	}
+	if _, err := dec.Token(); err != io.EOF {
+		return nil, fmt.Errorf("metadata %q must contain exactly one security-rules JSON array value", ExtensionKeySecurityRules)
+	}
+	if len(rules) == 0 {
+		return nil, fmt.Errorf("metadata %q must contain at least one security rule", ExtensionKeySecurityRules)
+	}
+	return rules, nil
 }
 
 func (r *NewSandboxRequest) parseCommonExtensions() error {
--- a/pkg/servers/e2b/network.go
+++ b/pkg/servers/e2b/network.go
@@ -24,7 +24,6 @@ import (
 
 	"k8s.io/klog/v2"
 
-	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
 	"github.com/openkruise/agents/pkg/sandbox-manager/infra"
 	"github.com/openkruise/agents/pkg/servers/e2b/models"
 	"github.com/openkruise/agents/pkg/servers/web"
@@ -112,18 +111,8 @@ func (sc *Controller) UpdateSandboxNetwork(r *http.Request) (web.ApiResponse[str
 		}
 	}
 
-	// Resolve the L7 rule replacement before touching any resource so a
-	// validation failure returns 400 with nothing written. Nil rules keep
-	// the existing chain; an explicit empty object clears it.
-	rulesJSON, rulesPresent, err := resolveSecurityRulesUpdate(&req)
-	if err != nil {
-		return web.ApiResponse[struct{}]{}, &web.ApiError{
-			Code:    http.StatusBadRequest,
-			Message: err.Error(),
-		}
-	}
-
-	// Validate and build the network config in one step.
+	// Validate and build the network config in one step, before resolving
+	// the L7 replacement, so the L4 lists are checked first.
 	netConfig, err := validateAndBuildNetworkConfig(&models.SandboxNetworkConfig{
 		AllowOut: req.AllowOut,
 		DenyOut:  req.DenyOut,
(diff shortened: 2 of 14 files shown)


pkg/servers/e2b/models/extensions.go
```diff
@@ -103,6 +106,14 @@ func (r *NewSandboxRequest) ParseExtensions() error {
 	if err := r.parseExtensionCSIMount(); err != nil {
 		return err
 	}
+	// Capture the reserved inline security-rules key. The raw value is only
+	// moved into Extensions; validation and normalization happen in the e2b
+	// server layer before the sandbox is created.
+	if raw, ok := r.Metadata[ExtensionKeySecurityRules]; ok {
```
furykerry, line 113: consider parse security rule directly by calling parseInlineSecurityRules without save the raw string here

pkg/servers/e2b/securityrules.go
```diff
@@ -0,0 +1,579 @@
...
+// size limits that apply to both input entries, and fills the block
+// statusCode default — the annotation bypasses the apiserver, so kubebuilder
+// defaults never run for it. Errors name the rule index so the HTTP response
+// points at the offending entry.
+func validateInlineSecurityRules(rules []agentsv1alpha1.SecurityRule) error {
+	if len(rules) > maxSecurityRules {
+		return fmt.Errorf("security-rules: %d rules exceed the maximum of %d", len(rules), maxSecurityRules)
+	}
+	for i := range rules {
+		rule := &rules[i]
```
furykerry, line 283: consider validate the uniqueness of the rule.name

pkg/servers/e2b/securityrules.go
```diff
@@ -0,0 +1,579 @@
...
+		}
+	}
+	for k, scheme := range match.Schemes {
+		if len(scheme) > 32 || !schemePattern.MatchString(scheme) {
+			return fmt.Errorf("%s.schemes[%d]: %q is not a valid scheme", path, k, scheme)
+		}
+	}
+	for k := range match.Paths {
+		p := &match.Paths[k]
+		if p.Type == agentsv1alpha1.PathMatchTypeRegex {
```
furykerry, line 353: consider validate the type is with allowed per `// +kubebuilder:validation:Enum=Prefix;Exact;Regex`

pkg/servers/e2b/securityrules.go
```diff
@@ -0,0 +1,579 @@
...
+		return nil
+	}
+	contract := newAllowOutContract(allowOut)
+	for i := range rules {
+		if !strings.HasPrefix(rules[i].Name, "e2b-rules-") || rules[i].Actions.HeaderManipulation == nil {
+			continue
+		}
+		for _, match := range rules[i].Match {
+			for _, domain := range match.Domains {
+				if !contract.permits(domain) {
```
furykerry: is it possible the global traffic policy has entries to allow out coming traffic?

pkg/servers/e2b/network.go
```diff
@@ -99,6 +112,17 @@ func (sc *Controller) UpdateSandboxNetwork(r *http.Request) (web.ApiResponse[str
 		}
 	}
 
+	// Resolve the L7 rule replacement before touching any resource so a
+	// validation failure returns 400 with nothing written. Nil rules keep
+	// the existing chain; an explicit empty object clears it.
+	rulesJSON, rulesPresent, err := resolveSecurityRulesUpdate(&req)
```
furykerry: Issue: resolveSecurityRulesUpdate runs before validateAndBuildNetworkConfig, so newAllowOutContract(req.AllowOut) sees entries that have not passed validateAllowOut.

Recommendation: move the validateAndBuildNetworkConfig call above resolveSecurityRulesUpdate.

pkg/servers/e2b/securityrules.go
```diff
@@ -0,0 +1,579 @@
...
+	if rulesJSON == "" || len(allowOut) == 0 {
+		return nil
+	}
+	var rules []agentsv1alpha1.SecurityRule
+	if err := json.Unmarshal([]byte(rulesJSON), &rules); err != nil {
+		return nil
+	}
+	contract := newAllowOutContract(allowOut)
+	for i := range rules {
+		if !strings.HasPrefix(rules[i].Name, "e2b-rules-") || rules[i].Actions.HeaderManipulation == nil {
```
furykerry: plz extract const securityRuleNamePrefix = "e2b-rules-"

pkg/servers/e2b/securityrules.go
```diff
@@ -0,0 +1,579 @@
...
+	for _, hv := range merged {
+		set = append(set, hv)
+	}
+	sort.Slice(set, func(i, j int) bool { return set[i].Name < set[j].Name })
+	return set, nil
+}
+
+// securityRuleNameForDomain derives a deterministic rule name from a domain.
+// Characters outside the CRD name alphabet are folded to '-' so wildcard and
+// dotted domains always yield a valid name.
```
furykerry: Issue: validateAllowOut hard-rejects any allowOut entry containing *, while RuleMatch.Domains explicitly supports * and *.example.com (and securityRuleNameForDomain's comment says it handles wildcards). So with any non-empty FQDN-only allowOut, a wildcard rule domain is a guaranteed 400. Reproduced:
```
allowOut: ["api.example.com"], rules: {"*.example.com": {headers...}}
→ network.rules["*.example.com"]: a domain with transform.headers must also be listed in network.allowOut
```

Recommendation: consider make validateAllowOut consistent with securityRuleNameForDomain logic

pkg/servers/e2b/securityrules.go
```diff
@@ -0,0 +1,579 @@
...
+	for k := range match.Paths {
+		p := &match.Paths[k]
+		if p.Type == agentsv1alpha1.PathMatchTypeRegex {
+			if _, err := regexp.Compile(p.Value); err != nil {
+				return fmt.Errorf("%s.paths[%d] regex %q does not compile: %v", path, k, p.Value, err)
+			}
+		}
+	}
+	for k := range match.Headers {
+		h := &match.Headers[k]
```
furykerry, line 360: L390-407 is not covered in UT

pkg/servers/e2b/securityrules.go
delavet: Yes. The cluster-level agentio egress policy (configured by the administrator) can allow traffic to domains that are not listed in the sandbox's `allowOut`. That makes the previous contract — "every `network.rules` domain must also appear in `allowOut`" — incorrect. 
The official API reference describes `network.rules` as: "A domain listed here is not automatically allowed - use allowOut to permit the traffic." — i.e. rules and allowOut are orthogonal by design: a rule never grants egress, and E2B does not require (or validate) that a rules domain appears in `allowOut`. Rejecting such requests with 400 would actually diverge from the upstream contract. Therefore, I removed it entirely.

pkg/servers/e2b/network.go
delavet: Fixed by dropping the allowOut containment check. Same as https://github.com/openkruise/agents/pull/838#discussion_r3840801429

pkg/servers/e2b/securityrules.go
```diff
@@ -0,0 +1,518 @@
...
+		set = append(set, hv)
+	}
+	sort.Slice(set, func(i, j int) bool { return set[i].Name < set[j].Name })
+	return set, nil
+}
+
+// securityRuleNameForDomain derives a deterministic rule name from a domain.
+// Characters outside the CRD name alphabet are folded to '-' so wildcard and
+// dotted domains always yield a valid name.
+func securityRuleNameForDomain(domain string) string {
```
furykerry, line 252: Issue: The duplicate check fixed the silent corruption, but the generator is still collision-prone. Probe-verified: a request transforming both a_b.com and a-b.com — two distinct, legitimate domains — now fails with security-rules[1].name "e2b-rules-a-b.com" duplicates security-rules[0].name. Same for any character outside [a-z0-9.-] (folded to -) and for the 253-char truncation.

Why: Fail-closed is acceptable as a security posture, but tenants get a 400 for a valid request with an error that points at rule names they never authored.

Recommendation: make generation collision-free, e.g.
```
sum := sha256.Sum256([]byte(strings.ToLower(domain)))
suffix := "-" + hex.EncodeToString(sum[:4])
name := securityRuleNamePrefix + strings.Trim(b.String(), "-")
if len(name)+len(suffix) > 253 {
    name = name[:253-len(suffix)]
}
return name + suffix
```

pkg/servers/e2b/securityrules.go
```diff
@@ -0,0 +1,518 @@
...
+	}
+	for k, scheme := range match.Schemes {
+		if len(scheme) > 32 || !schemePattern.MatchString(scheme) {
+			return fmt.Errorf("%s.schemes[%d]: %q is not a valid scheme", path, k, scheme)
+		}
+	}
+	for k := range match.Paths {
+		p := &match.Paths[k]
+		switch p.Type {
+		// An empty type is valid: EPE applies the CRD default (Prefix) at
```
furykerry, line 347: what does EPE means ?

pkg/servers/e2b/securityrules.go
delavet: done

pkg/servers/e2b/securityrules.go
delavet: That is the official name after the Traffic Extension was migrated into AgentIO：https://github.com/openkruise/agentio/tree/release-0.1/extensions/epe 😭

pkg/servers/e2b/securityrules.go
```diff
@@ -0,0 +1,525 @@
...
+// L7 rule simply never fires on traffic the L4 layer does not let out.
+func translateNetworkRules(net *models.SandboxNetworkConfig) ([]agentsv1alpha1.SecurityRule, error) {
+	domains := make([]string, 0, len(net.Rules))
+	for domain := range net.Rules {
+		domains = append(domains, domain)
+	}
+	sort.Strings(domains)
+
+	rules := make([]agentsv1alpha1.SecurityRule, 0, len(domains))
+	for _, domain := range domains {
```
furykerry, line 177: [nit] Case-variant rule domains produce a confusing generated-name error

Domains that differ only by case (e.g. `Example.com` and `example.com`) are distinct keys in the `network.rules` map, but they fold to the same lowercased input in `securityRuleNameForDomain`, so both generate the identical rule name and trip the rule-name uniqueness check added in `validateInlineSecurityRules`. Verified at `9e41e53f7`:

```json
{
  "network": {
    "rules": {
      "Example.com": [{"transform": {"headers": {"x-one": "1"}}}],
      "example.com": [{"transform": {"headers": {"x-two": "2"}}}]
    }
  }
}
```

```
400: security-rules[1].name "e2b-rules-example.com-a379a6f6" duplicates security-rules[0].name
```

This fails closed, which is fine — but the error blames a server-generated artifact instead of pointing at the actual input problem. Since DNS names are case-insensitive, these two keys are the same domain, and the hash suffix makes the collision deterministic rather than accidental.

Suggestion: detect case-variant domain keys up front and reject them with a domain-level message, mirroring the existing case-variant header check in `translateDomainTransforms`:

```go
rules := make([]agentsv1alpha1.SecurityRule, 0, len(domains))
seenDomains := make(map[string]string, len(domains))
for _, domain := range domains {
	lower := strings.ToLower(domain)
	if prev, dup := seenDomains[lower]; dup {
		return nil, fmt.Errorf("network.rules: %q and %q are the same domain (domains are case-insensitive)",
			prev, domain)
	}
	seenDomains[lower] = domain
	// ... existing checks
}
```

delavet: Fixed. `translateNetworkRules` now detects case-variant domain keys up front.

PR #839, opened 2026-08-19
feat(api): support PauseStrategy on SandboxSet

zmberg:

`SandboxSet` had no way to configure `PauseStrategy`, even though `Sandbox`
itself already supports it. Sandboxes created from a `SandboxSet` always got
a nil `PauseStrategy` regardless of what the pool wanted.

This PR:

- Adds `SandboxSetSpec.PauseStrategy (*PauseStrategy)`.
- Copies the field verbatim onto every `Sandbox` created via
  `NewSandboxFromSandboxSet`.
- Regenerates deepcopy and CRD manifests via `make generate manifests`.
- Extends the existing table-driven `TestNewSandboxFromSandboxSet` with a
  `pauseStrategy` case and assertion.

NONE

1. `go test ./pkg/controller/sandboxset/... -run TestNewSandboxFromSandboxSet -v`
2. Create a `SandboxSet` with `spec.pauseStrategy` set, and verify the
   resulting `Sandbox.spec.pauseStrategy` matches.

- `PauseStrategy` is intentionally excluded from `computeRevisionHash`: it
  is a runtime pause-behavior setting, not part of the pod template, so
  changing it should not trigger a `SandboxSet` update revision.
- Scope is limited to `SandboxSet`; `SandboxTemplate` and `SandboxClaim` are
  not covered by this change.


pkg/controller/sandboxset/revision.go
```diff
@@ -52,6 +52,7 @@ func (r *Reconciler) buildSandboxTemplateSpec(ctx context.Context, sbs *agentsv1
 			VolumeClaimTemplates: sbs.Spec.VolumeClaimTemplates,
 			PersistentContents:   sbs.Spec.PersistentContents,
 			Runtimes:             sbs.Spec.Runtimes,
+			PauseStrategy:        sbs.Spec.PauseStrategy,
```
HARSHRAJ2789, line 55: The "Special notes" bullet reads backwards against the code and against this PR's own godoc.

The note says `PauseStrategy` is intentionally excluded from `computeRevisionHash`, so changing
it should not trigger an update revision. The chain does the opposite. This line and `:68` put
it into the spec that `ensureTemplateRevision` hands to `computeRevisionHash`
(`template_materialize.go:42-46`), and that function marshals the whole `SandboxTemplateSpec`,
not `spec.template` alone. `initNewStatus` then assigns the result to
`newStatus.UpdateRevision`.

Both godoc comments added here describe that correctly. `SandboxSetSpec.PauseStrategy` says it
"is part of the update revision hash, so changing it triggers a rolling update of
already-created pool sandboxes", and `SandboxSetStatus.UpdateRevision` is updated to list
`spec.pauseStrategy` among the hashed fields. `computeLegacyRevisionHash` is the one that
excludes it, and that is the comment corrected at line 97.

So the behaviour looks deliberate and documented, and only the review note contradicts it.
Worth correcting before merge, since it points a reviewer at the opposite of what happens.

Two details that hold either way. Both fields are `*PauseStrategy` with `omitempty`, so nil
marshals away and no existing SandboxSet changes hash on upgrade. And because
`isRevisionUpdated` (`rolling_update.go:104`, untouched here) also accepts the legacy hash, a
sandbox still carrying that label stays up to date across a `pauseStrategy` change; the
replacement applies to sandboxes labelled with the current revision.

PR #840, opened 2026-08-19
Fix invalid SandboxClaim retry loop

googs1025:

This PR stops invalid `SandboxClaim` specs from retrying indefinitely when they set reserved sandbox identity metadata.

`validateClaimReservedIdentityKeys` now wraps reserved `sandbox-id` label/annotation violations with a typed `ErrInvalidClaimSpec`. When claiming sees this non-recoverable error, the controller transitions the claim to `Completed` with reason `InvalidClaimSpec`, records a warning event, and returns without requeueing.

NONE

```bash
GOCACHE=/private/tmp/agents-go-build-cache go test ./pkg/controller/sandboxclaim -count=1
GOCACHE=/private/tmp/agents-go-build-cache go test ./pkg/controller/sandboxclaim/core -count=1
```

This PR only adds the controller-side terminal handling. It intentionally does not add a `SandboxClaim` admission webhook yet, so existing invalid claims and clusters without the webhook enabled still stop retrying. Admission rejection for new invalid specs can be considered as a follow-up.

### DIFF
--- a/pkg/controller/sandboxclaim/core/common_control.go
+++ b/pkg/controller/sandboxclaim/core/common_control.go
@@ -155,7 +155,12 @@ func (c *commonControl) EnsureClaimClaiming(ctx context.Context, args ClaimArgs)
 	claimed, err := c.claimSandboxes(ctx, claim, sandboxSet, batchSize)
 	if err != nil {
 		if errors.Is(err, ErrInvalidClaimSpec) {
+			// Strip the generic "failed to build claim options" wrapper so users
+			// see the actual validation reason in status and events.
 			msg := err.Error()
+			if inner := errors.Unwrap(err); inner != nil {
+				msg = inner.Error()
+			}
 			log.Info("Invalid SandboxClaim spec, completing claim without retry", "err", err)
 			c.recorder.Event(claim, "Warning", "InvalidClaimSpec", msg)
 			TransitionToCompleted(args.NewStatus, "InvalidClaimSpec", msg)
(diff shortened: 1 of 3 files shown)


HARSHRAJ2789:
Ran both packages from the description on the branch, `sandboxclaim` and `sandboxclaim/core` pass.
The terminal handling reads correctly to me: `buildClaimOptions` returns the sentinel, `:248`
re-wraps it with `%w`, so `errors.Is` at `:157` still matches through that layer, and
`claimedCount` is zero on that path, so returning before Step 9 loses nothing.

On the admission follow-up you mention in the notes, one thing about how far the validator
currently reaches.

`validateClaimReservedIdentityKeys` rejects two keys, and the propagation it guards copies all of
`claim.Spec.Labels` onto the Sandbox with no filter at `:337`, and again onto the pod template at
`:348`. The annotations at `:357` do go through `annotationutils.FilterBlackListed`. Labels have no
equivalent, so every other `agents.kruise.io/` label reaches the sandbox.

`LabelSandboxIsClaimed` comes out intact, but the ordering is what saves it, and the validator
plays no part.
`modifyPickedSandbox` runs `opts.Modifier` at `claim.go:706` and re-asserts the label afterwards at
`:725`. `LabelSandboxPool` has no such re-assertion. Feeding the same unfiltered copy through
`modifyPickedSandbox` with both labels set gives:

```
PROBE pool="victim-pool"  claimed="true"
```

`validateRecyclePreconditions` resolves the SandboxSet from that label, `recycle.go:226` into the
`client.Get` at `:231`, so a claim-supplied pool label points the recycle at a different SandboxSet
or fails it with `SandboxSet %s not found`.

None of this arrives with your change: the loop is identical on master at `:327`. It is worth
knowing here because this PR is what makes the reserved-key path terminal, which puts the distance
between the two keys that are checked and the rest that are not on the same code path.

googs1025:
@furykerry /PTAL

PR #704, opened 2026-07-28
fix(proxy): add nil checks for HttpHeaders in ext_proc header helpers

vishalmore90:
### Ⅰ. Describe what this PR does

This PR fixes a potential runtime nil pointer dereference panic in `pkg/proxy/ext_proc.go`. 

Specifically:
- Added `nil` checks for `httpHeaders == nil || httpHeaders.Headers == nil` in `extProcHeadersToMap` and `headerModifiers`.
- Prevents the proxy server stream handler from panicking when processing gRPC requests sent by Envoy containing empty or `nil` header maps.
- Added unit tests `TestExtProcHeadersToMap_NilHandling` and `TestHeaderModifiers_NilHandling` in `ext_proc_test.go`.

### Ⅱ. Does this pull request fix one issue?

fixes #703

### Ⅲ. Describe how to verify it

Run the unit tests in `pkg/proxy`:
```bash
go test -v ./pkg/proxy/...

PR #844, opened 2026-08-19
fix(proxy): answer each ext_proc phase with its matching response

### LINKED ISSUE #834: [BUG] ext_proc answers every processing phase with a `RequestHeaders` response — works on envoy 1.33, fails closed on envoy ≥ 1.35
**What happened**:

Running the E2B Python SDK against sandbox-manager through its envoy sidecar (envoy ≥ 1.35), every request failed with
a locally generated 500 and an empty body, so `Sandbox.create()` died on
`json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`. envoy logged one warning per request:

```
[warning][ext_proc] Spurious response message 1 received on gRPC stream
```

`Server.Process` (`pkg/proxy/ext_proc.go`) pre-builds a single `RequestHeaders` response and sends it
for every `ProcessingRequest`, whatever phase envoy is in — the `default` branch only logs and still sends it:

```go
resp := &extProcPb.ProcessingResponse{
    Response: &extProcPb.ProcessingResponse_RequestHeaders{...},
}
switch v := req.Request.(type) {
case *extProcPb.ProcessingRequest_RequestHeaders:
    resp = s.handleRequestHeaders(...)
default:
    log.Info("Unknown Request type", "type", v)
}
srv.Send(resp)
```

The response-header phase is explicitly enabled by our own config
(`config/sandbox-manager/envoy-config.yaml`: `response_header_mode: SEND`), so envoy always reaches a
phase that gets the wrong response type.

Rolling the envoy sidecar back to `v1.33.2` makes everything pass with no code change, and that is what
we run now. The reason is a policy change in envoy: up to 1.34 an out-of-order message is
unconditionally ignored (`closeStream()` + `clearAsyncState()`, which resumes the paused response), so
the request still succeeds; from 1.35 that branch is gated by
`envoy.reloadable_features.ext_proc_fail_close_spurious_resp`, declared `RUNTIME_GUARD` — i.e. **on by
default** — and takes `handleErrorResponse()`, returning 500.

So this is not urgent on the pinned envoy, but it looks worth improving: the handler violates the
ext_proc contract, the mismatch is invisible until the sidecar is upgraded.

Each phase envoy sends is answered with the matching `ProcessingResponse` type, so behavior does not
depend on the envoy version or on which phases `processing_mode` enables.

Deploy sandbox-manager with the repository's own envoy ConfigMap but an envoy image ≥ 1.35 (I used
`envoyproxy/envoy:v1.39.0`), apply `test/e2b/assets/sandboxset-code-interpreter.yaml`, then talk to the
envoy port (7788) with the SDK plus the official patch:

```python
os.environ["E2B_DOMAIN"] = "localhost:18080"
from kruise_agents.patch_e2b import patch_e2b
patch_e2b(https=False)
from e2b import Sandbox
Sandbox.create(template="code-interpreter", timeout=300)   # 500, empty body
```

`curl` on the same port reproduces it too. Swap the sidecar to `v1.33.2` and the same call succeeds.

Measured in one kind cluster, same script, envoy ConfigMap untouched (`ext_proc.failure_mode_allowed`
stayed 0, so no fail-open was configured):

- unpatched manager + envoy 1.33.2 → all requests succeed, `ext_proc.spurious_msgs_received: 7`
  (one per request, silently)
- unpatched manager + envoy 1.39.0 → 500 with `bytes_sent:0`, `spurious_msgs_received: 1`
- patched manager + envoy 1.39.0 → all requests succeed, `spurious_msgs_received: 0`

The counter on 1.33 is the useful part: the contract is already being violated on the pinned version,
it just does not surface. `failure_mode_allow: true` is the other non-code workaround, but it also makes
the filter fail open for genuine errors.

**Environment**:
- OpenKruiseAgent version: release v0.3.0
- Kubernetes version: v1.33.1 (kind)
- Install details: sandbox-manager + envoy sidecar, envoy ConfigMap unmodified, Service →
  `targetPort: 7788`; envoy swapped between `v1.39.0` (fails) and `v1.33.2` (passes)
- Others: sandbox image `registry-ap-southeast-1.ack.aliyuncs.com/acs/code-interpreter:v1.6`; client
  `e2b==2.10.2` with `sdk/customized_e2b`

nishantbkl3345-ship-it:

Fixes #834.

**Problem**

`Server.Process` (`pkg/proxy/ext_proc.go`) pre-built a single `RequestHeaders` response and sent it for every `ProcessingRequest`, whatever phase envoy was in. The `default` branch logged the unhandled type and then sent that same `RequestHeaders` response anyway:

```go
resp := &extProcPb.ProcessingResponse{
    Response: &extProcPb.ProcessingResponse_RequestHeaders{...},
}
switch v := req.Request.(type) {
case *extProcPb.ProcessingRequest_RequestHeaders:
    resp = s.handleRequestHeaders(...)
default:
    log.Info("Unknown Request type", "type", v)
}
srv.Send(resp)
```

ext_proc requires the response variant to match the request variant. The mismatched phase is reached in the shipped configuration — `config/sandbox-manager/envoy-config.yaml` sets `response_header_mode: SEND`, so envoy always reaches the response-header phase and receives a `RequestHeaders` reply for it.

**Impact**

envoy up to 1.34 unconditionally ignores an out-of-order message and resumes the paused stream, so the request still succeeds and the only symptom is one warning per request:

```
[warning][ext_proc] Spurious response message 1 received on gRPC stream
```

From envoy 1.35 that branch is gated by `envoy.reloadable_features.ext_proc_fail_close_spurious_resp`, declared `RUNTIME_GUARD` and therefore on by default, and the stream instead fails closed. Every request through the sidecar then returns a locally generated 500 with an empty body, which surfaces in the E2B SDK as `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`.

Scoping this honestly: **this is latent, not a live outage.** `config/sandbox-manager/deployment.yaml` pins the sidecar to `envoyproxy/envoy:v1.33-latest`, where the mismatch is tolerated. The value of the fix is that the handler stops depending on that pin — the sandbox-gateway image already moved to envoy 1.37.3 in #509, so a sidecar bump is a realistic change that would otherwise take the data path down.

**Root cause**

The phase was never used to select the response variant. A single `RequestHeaders` response was built up front and only overwritten for the `RequestHeaders` case, so every other phase — including the one the shipped config enables — received a variant that does not correspond to the request.

**Fix**

Select the response variant from the phase, covering all six processing phases. Phases the proxy does not act on are answered with an empty continue response (`CommonResponse` defaults to `CONTINUE`), which leaves existing behavior for the request-header phase untouched.

A phase the server does not recognize is now skipped instead of being answered with a guessed variant, since sending the wrong variant is precisely what envoy rejects.

The change is confined to the phase dispatch in `Process`; `handleRequestHeaders` and the routing logic are unchanged.

fixes #834

The regression coverage extends the existing `TestServer_Process` table in `pkg/proxy/ext_proc_test.go`, which drives the real `Process` loop over a mock `ExternalProcessor_ProcessServer` — the same entry point envoy talks to.

Two parts:

1. A response-variant assertion in the verification loop, so a phase/response mismatch fails the test instead of being ignored:

   ```go
   assert.IsType(t, expected.Response, actual.Response)
   ```

2. New cases for the `response headers`, `request body` and `response trailers` phases, each expecting its matching response variant.

One pre-existing case needed correcting rather than extending. The case named `unknown request type` sent a **`ResponseHeaders`** request and asserted a **`RequestHeaders`** reply — it had codified the mismatch, and treated a phase the shipped config explicitly enables as "unknown". It now exercises the `default` branch honestly with an unset request type and expects no response to be sent.

```bash
go test ./pkg/proxy/ -run 'TestServer_Process$' -count=1
```

**Before the fix** (final test file, previous `ext_proc.go`):

```
--- FAIL: TestServer_Process
    --- FAIL: TestServer_Process/unset_request_type
        ext_proc_test.go:615: expect 0 responses, got 1
    --- FAIL: TestServer_Process/response_headers_phase
        Error: Object expected to be of type *ext_procv3.ProcessingResponse_ResponseHeaders,
               but was *ext_procv3.ProcessingResponse_RequestHeaders
    --- FAIL: TestServer_Process/request_body_phase
        Error: Object expected to be of type *ext_procv3.ProcessingResponse_RequestBody,
               but was *ext_procv3.ProcessingResponse_RequestHeaders
    --- FAIL: TestServer_Process/response_trailers_phase
        Error: Object expected to be of type *ext_procv3.ProcessingResponse_ResponseTrailers,
               but was *ext_procv3.ProcessingResponse_RequestHeaders
FAIL
```

**After the fix** — all 13 subtests pass, including the 8 pre-existing ones that are unaffected by the change:

```
--- PASS: TestServer_Process
ok      github.com/openkruise/agents/pkg/proxy  0.406s
```

Validation performed:

| Check | Result |
| --- | --- |
| `gofmt -l` on both changed files | clean |
| `go build ./...` | pass |
| `go vet ./pkg/proxy/...` | pass |
| `golangci-lint run ./pkg/proxy/...` | 0 issues |
| `go test ./pkg/proxy/... -count=1` | pass |
| `go test ./pkg/proxy/... -count=3 -race` | clean |
| `go test ./pkg/... -count=1` | zero failures |

No API types changed, so `make generate manifests` does not apply.

The empty responses are deliberate: `CommonResponse.Status` defaults to `CONTINUE`, and `TrailersResponse` carries only an optional header mutation, so an empty value for each is the "continue, change nothing" answer for that phase. The proxy's behavior on the request-header phase — the only phase it acts on — is byte-for-byte unchanged.

`processing_mode` in the shipped config currently enables only `request_header_mode: SEND` and `response_header_mode: SEND`, so the body and trailer cases are not reachable with today's ConfigMap. They are handled anyway so that widening `processing_mode` later does not reintroduce the same class of failure.

The `default` branch now skips the turn instead of sending a response. All six phases are enumerated in the ext_proc proto, so this is reachable only for an unset or newly added request variant; for those, sending nothing is strictly safer than sending a variant envoy will reject.

This touches `pkg/proxy/ext_proc.go`, which #704 and #818 also modify. There is no functional overlap — #704 adds nil guards to the header helpers and #818 filters reserved keys in the header modifier, while this PR only changes how the phase selects the response variant. Whichever merges first, the others should rebase cleanly or need only a trivial resolution in `Process`.

### DIFF
--- a/pkg/proxy/ext_proc.go
+++ b/pkg/proxy/ext_proc.go
@@ -65,22 +65,55 @@ func (s *Server) Process(srv extProcPb.ExternalProcessor_ProcessServer) error {
 			return status.Errorf(codes.Unknown, "cannot receive stream request: %v", err)
 		}
 
-		// build response based on request type
-		resp := &extProcPb.ProcessingResponse{
-			Response: &extProcPb.ProcessingResponse_RequestHeaders{
-				RequestHeaders: &extProcPb.HeadersResponse{
-					Response: &extProcPb.CommonResponse{},
-				},
-			},
-		}
+		// Build the response for the phase envoy is in. ext_proc requires the
+		// response variant to match the request variant: envoy up to 1.34 ignores
+		// a mismatch, but from 1.35 it fails the stream closed and returns 500.
+		// Phases we do not act on are answered with an empty continue response.
+		var resp *extProcPb.ProcessingResponse
 		switch v := req.Request.(type) {
 		case *extProcPb.ProcessingRequest_RequestHeaders:
-			h := req.Request.(*extProcPb.ProcessingRequest_RequestHeaders)
-			resp = s.handleRequestHeaders(h, log)
-
+			resp = s.handleRequestHeaders(v, log)
+		case *extProcPb.ProcessingRequest_ResponseHeaders:
+			resp = &extProcPb.ProcessingResponse{
+				Response: &extProcPb.ProcessingResponse_ResponseHeaders{
+					ResponseHeaders: &extProcPb.HeadersResponse{
+						Response: &extProcPb.CommonResponse{},
+					},
+				},
+			}
+		case *extProcPb.ProcessingRequest_RequestBody:
+			resp = &extProcPb.ProcessingResponse{
+				Response: &extProcPb.ProcessingResponse_RequestBody{
+					RequestBody: &extProcPb.BodyResponse{
+						Response: &extProcPb.CommonResponse{},
+					},
+				},
+			}
+		case *extProcPb.ProcessingRequest_ResponseBody:
+			resp = &extProcPb.ProcessingResponse{
+				Response: &extProcPb.ProcessingResponse_ResponseBody{
+					ResponseBody: &extProcPb.BodyResponse{
+						Response: &extProcPb.CommonResponse{},
+					},
+				},
+			}
+		case *extProcPb.ProcessingRequest_RequestTrailers:
+			resp = &extProcPb.ProcessingResponse{
+				Response: &extProcPb.ProcessingResponse_RequestTrailers{
+					RequestTrailers: &extProcPb.TrailersResponse{},
+				},
+			}
+		case *extProcPb.ProcessingRequest_ResponseTrailers:
+			resp = &extProcPb.ProcessingResponse{
+				Response: &extProcPb.ProcessingResponse_ResponseTrailers{
+					ResponseTrailers: &extProcPb.TrailersResponse{},
+				},
+			}
 		default:
+			// Answering an unrecognized phase with a guessed variant is what envoy
+			// rejects, so skip the turn instead of sending a mismatched response.
 			log.Info("Unknown Request type", "type", v)
-
+			continue
 		}
 
 		if err = srv.Send(resp); err != nil {
--- a/pkg/proxy/ext_proc_test.go
+++ b/pkg/proxy/ext_proc_test.go
@@ -421,28 +421,18 @@ func TestServer_Process(t *testing.T) {
 			expectError: true,
 		},
 		{
-			name:        "unknown request type",
+			// A phase the server does not recognize is skipped rather than
+			// answered with a guessed response variant.
+			name:        "unset request type",
 			setupRoutes: []sandboxroute.Route{},
 			adapter: &testRequestAdapter{
 				entry: "127.0.0.1:8080",
 			},
 			requests: []*extProcPb.ProcessingRequest{
-				{
-					Request: &extProcPb.ProcessingRequest_ResponseHeaders{
-						ResponseHeaders: &extProcPb.HttpHeaders{},
-					},
-				},
+				{Request: nil},
 			},
 			expectError: false,
-			expectResp: []*extProcPb.ProcessingResponse{
-				{
-					Response: &extProcPb.ProcessingResponse_RequestHeaders{
-						RequestHeaders: &extProcPb.HeadersResponse{
-							Response: &extProcPb.CommonResponse{},
-						},
-					},
-				},
-			},
+			expectResp:  []*extProcPb.ProcessingResponse{},
 		},
 		{
 			name: "extra headers",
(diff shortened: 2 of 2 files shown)


HARSHRAJ2789 (commented):
I went through this against a local checkout of the branch. The diagnosis holds up: I confirmed
`response_header_mode: SEND` at `config/sandbox-manager/envoy-config.yaml:70`, the sidecar pin at
`deployment.yaml:103`, and that #509 moved the gateway base image to v1.37.3. The scoping as
latent is right for the reason you give and one more: `proxy.Server` is only constructed at
`pkg/sandbox-manager/core.go:81`, and the gateway runs a native `.so` filter, not
ext_proc, so the v1.37.3 image never talks to this server.

I also reran your mutation check, master's `ext_proc.go` against your test file. It is stronger
than the PR body shows. Six phase cases fail plus the unset one, where the "Before the fix"
block lists three phase failures plus unset. That block also cites `ext_proc_test.go:615`, and
in the final file that assertion sits at `:655`, so it looks pasted before the `response_body`
and `request_trailers` cases were added. The prose above it names three new phases where the
diff adds five. Worth refreshing so the coverage does not read as thinner than it is.

Three things I would add or ask about.

`config/sandbox-manager/image_patch.yaml:9` carries the same `envoyproxy/envoy:v1.33-latest` pin.
The body names only `deployment.yaml`, so the "realistic sidecar bump" you describe is a
two-file change.

`go.mod:12` already requires `github.com/envoyproxy/envoy v1.37.3`, so the tree compiles against
1.37 protos while running a 1.33 sidecar. That strengthens the argument for fixing this now.

The one I would push back on is `default: continue` being "strictly safer". The filter config
sets neither `message_timeout` nor `failure_mode_allow`, so envoy's defaults apply, and by my
reading of the docs that is a 200ms deadline with failure closed. If so, then on the version
actually deployed the two behaviours are not ordered the way the comment says: envoy 1.33
ignores a spurious response and resumes the stream, so the old code let the request through,
while sending nothing stalls for the timeout and then fails. `continue` would be safer on 1.35+
and worse on the pinned 1.33.

This is close to unreachable today, since all six variants are enumerated and only an unset
oneof reaches `default`, so I am not asking for a change. But the comment states a general rule
that inverts on the deployed version, and the reasoning would matter to whoever widens
`processing_mode` later. Either narrowing the comment to 1.35+, or setting an explicit
`failure_mode_allow`, would make it hold in both directions.

One small thing on the invariant as stated: `handleRequestHeaders` can return an
`ImmediateResponse` on the error paths, which is legal for any phase. A reader checking the code
against "the response variant must match the request variant" may briefly think that path
violates it, so it may be worth a word.

nishantbkl3345-ship-it:
Hey @AiRanthem @zmberg, just following up on this PR. When you get some time, could you please take a look . I’m happy to make any changes needed. Thanks!

PR #849, opened 2026-08-21
fix(controller): surface missing template refs

googs1025:

Surface missing SandboxTemplate references to users instead of leaving the CR status without the root cause.

- Set `TemplateResolved=False` when SandboxSet `templateRef` resolution fails during status initialization.
- Set Sandbox `Ready=False` when Pod generation fails before Pod creation, including missing `templateRef`.
- Add regression tests for visible condition feedback on missing template references.

NONE

- `GOCACHE=/private/tmp/agents-go-build-cache go test ./pkg/controller/sandbox/core -count=1`
- `GOCACHE=/private/tmp/agents-go-build-cache go test ./pkg/controller/sandboxset -count=1`
- `GOCACHE=/private/tmp/agents-go-build-cache go test ./pkg/controller/sandbox/... -count=1`

The local git transport to GitHub was unavailable in this environment, so the branch commit was created on the fork through GitHub API with the same file changes and Signed-off-by trailer.

### DIFF
--- a/pkg/controller/sandboxset/utils.go
+++ b/pkg/controller/sandboxset/utils.go
@@ -18,6 +18,7 @@ package sandboxset
 
 import (
 	"context"
+	"errors"
 	"maps"
 	"strings"
 	"time"
@@ -59,13 +60,31 @@ type initStatusResult struct {
 
 func (r *Reconciler) initNewStatus(ctx context.Context, ss *agentsv1alpha1.SandboxSet) (*initStatusResult, error) {
 	newStatus := ss.Status.DeepCopy()
+	// Record the generation even when template resolution fails.
+	newStatus.ObservedGeneration = ss.Generation
 	hash, name, err := r.ensureTemplateRevision(ctx, ss)
 	if err != nil {
+		utils.SetCondition(&newStatus.Conditions, metav1.Condition{
+			Type:               string(agentsv1alpha1.SandboxSetConditionTemplateResolved),
+			Status:             metav1.ConditionFalse,
+			LastTransitionTime: metav1.Now(),
+			Reason:             agentsv1alpha1.SandboxSetReasonTemplateResolveFail,
+			Message:            utils.TruncateConditionMessage(err.Error()),
+		})
+		if updateErr := r.updateSandboxSetStatus(ctx, *newStatus, ss); updateErr != nil {
+			return nil, errors.Join(err, updateErr)
+		}
 		return nil, err
 	}
 	newStatus.UpdateRevision = hash
-	newStatus.ObservedGeneration = ss.Generation
 	newStatus.CurrentRevision = name
+	utils.SetCondition(&newStatus.Conditions, metav1.Condition{
+		Type:               string(agentsv1alpha1.SandboxSetConditionTemplateResolved),
+		Status:             metav1.ConditionTrue,
+		LastTransitionTime: metav1.Now(),
+		Reason:             agentsv1alpha1.SandboxSetReasonTemplateResolved,
+		Message:            "Sandbox template resolved",
+	})
 
 	// Compute legacy hash for backward compatibility with sandboxes created
 	// before the hash algorithm was changed. Errors are non-fatal: if legacy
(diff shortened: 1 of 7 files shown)


HARSHRAJ2789 (commented):
Checked this out locally and ran it. Both new tests are load-bearing, which I verified by
reverting the production change while keeping the tests:
`TestCreatePod_TemplateRefResolutionFailureSetsCondition` fails on "expected Ready condition to
be set", and `TestReconcile_MissingTemplateRefSetsCondition` fails on "expected TemplateResolved
condition to be set". The sandboxset half needs a surgical revert to show this, since dropping
the whole file takes the helpers the test imports with it.

A few other things I checked that are fine: `Conditions` already exists on `SandboxSetStatus`
(`sandboxset_types.go:171`) and in the CRD, so no regeneration is needed. A repeated failure also settles quietly, because `setSandboxSetCondition` returns early when
status, reason and message all match, so `LastTransitionTime: metav1.Now()` never lands, and
`updateSandboxSetStatus` guards on `reflect.DeepEqual` before writing.

Two things worth changing.

The description says both paths "emit a Warning event", and neither does so as a result of this
PR. On the SandboxSet side the event already exists on master:
`sandboxset_controller.go:376` calls `r.Recorder.Eventf(... EventCreateSandboxFailed, "Failed to
resolve sandbox template: %s", err)`. This PR adds the condition next to it, not the event.

On the Sandbox side there is no event at all. `pod_control.go` returns from the `generatePod`
failure without touching the recorder both before and after this change, and the PR's own test
pins that:

```go
select {
case event := <-recorder.Events:
    t.Fatalf("did not expect event for templateRef resolution failure: %s", event)
default:
}
```

The test is right and the description is not. Since a reviewer skimming the summary would expect
an event on the Sandbox path, it seems worth correcting the description, or adding the event and
flipping that assertion.

The second is reuse. `setSandboxSetCondition` and `getSandboxSetCondition` are line-for-line the
same as `utils.SetSandboxCondition` and `utils.GetSandboxCondition` in `pkg/utils/utils.go:49`
and `:66`, down to the `else if currentCond == nil` shape and the `LastTransitionTime.IsZero()`
check. The only difference is `*SandboxSetStatus` where the existing pair takes `*SandboxStatus`.
This PR already calls `utils.SetSandboxCondition` on the Sandbox side, so it ends up using the
helper in one file and a copy of it in another.

Both types hold a plain `[]metav1.Condition`, so a helper taking the slice, or a small generic
over the two status types, would let both sides share one implementation. Worth doing before a third copy
appears, since a future fix to the transition-time logic would otherwise need to be applied in
more than one place.

The change itself reads well and the two tests are the right two tests.

googs1025:
Updated this in the PR branch.

- PR description now only claims condition feedback, matching the current behavior and tests.
- Extracted shared `utils.SetCondition` / `utils.GetCondition` helpers over `[]metav1.Condition`. `SetSandboxCondition` now delegates to the shared helper, and the SandboxSet path uses the same helper instead of its local copy.

Verified locally:
- `GOCACHE=/private/tmp/agents-go-build-cache go test ./pkg/utils -count=1`
- `GOCACHE=/private/tmp/agents-go-build-cache go test ./pkg/controller/sandbox/core -count=1`
- `GOCACHE=/private/tmp/agents-go-build-cache go test ./pkg/controller/sandboxset -count=1`

googs1025:
@furykerry /PTAL

pkg/controller/sandboxset/utils.go
```diff
@@ -38,6 +39,12 @@ var (
 	retryAfterForceDeleteExpectation = 3 * time.Second
 )
 
+const (
+	SandboxSetConditionTemplateResolved = "TemplateResolved"
```
furykerry: plz move sandboxset condition type and reason definition to api/v1alpha1/sandboxset_types.go

googs1025: Fixed in 07b77de

googs1025:
@furykerry can you take a look this? 😄

PR #853, opened 2026-08-24
feat: improve CSI mount error observability in SandboxClaim

AlbeeSo:
- Include stderr content in CSIMount error messages
- Add last error to SandboxClaim status.Message and Warning events
- Add truncateError helper for event message length limits
- Keep the last error on the clone path so a CloneTimeout no longer discards the CSI mount cause, mirroring the claim path
- Stop duplicating the whole message on both paths when the retry loop already promoted the last error into the returned error

### DIFF
--- a/pkg/controller/sandboxclaim/core/common_control.go
+++ b/pkg/controller/sandboxclaim/core/common_control.go
@@ -159,6 +159,9 @@ func (c *commonControl) EnsureClaimClaiming(ctx context.Context, args ClaimArgs)
 	finalCount := currentCount + int32(claimed) // #nosec G115 -- K8s object count
 	args.NewStatus.ClaimedReplicas = finalCount
 	args.NewStatus.Message = fmt.Sprintf("Claiming sandboxes: %d/%d claimed", finalCount, desiredReplicas)
+	if err != nil {
+		args.NewStatus.Message = fmt.Sprintf("%s, last error: %s", args.NewStatus.Message, utils.TruncateConditionMessage(err.Error()))
+	}
 
 	// Step 10: Record results and determine requeue strategy
 	if claimed > 0 {
@@ -176,8 +179,11 @@ func (c *commonControl) EnsureClaimClaiming(ctx context.Context, args ClaimArgs)
 	// No progress - no available sandboxes
 	log.Info("No available sandboxes, will retry",
 		"retryInterval", ClaimRetryInterval)
-	c.recorder.Event(claim, "Warning", "NoAvailableSandboxes",
-		fmt.Sprintf("No available sandboxes in pool %s", sandboxSet.Name))
+	eventMsg := fmt.Sprintf("No available sandboxes in pool %s", sandboxSet.Name)
+	if err != nil {
+		eventMsg = fmt.Sprintf("%s, last error: %s", eventMsg, utils.TruncateConditionMessage(err.Error()))
+	}
+	c.recorder.Event(claim, "Warning", "NoAvailableSandboxes", eventMsg)
 	sandboxSetClaimsTotal.WithLabelValues(claim.Namespace, "failed").Inc()
 	// Retry after interval to avoid busy loop
 	return RequeueAfter(ClaimRetryInterval), nil
--- a/pkg/sandbox-manager/infra/sandboxcr/infra.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/infra.go
@@ -211,8 +211,13 @@ func buildClaimError(err error, lastError error, failures []infra.PickSandboxFai
 	if errors.As(err, &mErr) {
 		return mErr
 	}
-	// Retry exhausted or interrupted: wrap as Internal
-	base := fmt.Sprintf("%v, last error: %v", err, lastError)
+	// Retry exhausted or interrupted: wrap as Internal. On retry exhaustion the
+	// caller already promoted lastError into err, so appending it again would
+	// duplicate the whole message (stderr included).
+	base := fmt.Sprintf("%v", err)
+	if lastError != nil && lastError.Error() != err.Error() {
+		base = fmt.Sprintf("%s, last error: %v", base, lastError)
+	}
 	if len(failures) > 0 {
 		raw, marshalErr := json.Marshal(failures)
 		if marshalErr == nil {
--- a/pkg/sandbox-manager/infra/sandboxcr/infra_test.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/infra_test.go
@@ -1469,6 +1469,7 @@ func TestBuildClaimError_PreservesTerminalError(t *testing.T) {
 		failures        []infra.PickSandboxFailure
 		expectErrorCode managererrors.ErrorCode
 		expectCause     bool
+		expectError     string
 	}{
 		{
 			name:            "nil error returns nil",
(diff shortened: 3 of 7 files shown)


HARSHRAJ2789:
Ran the touched packages on the branch: `sandboxclaim`, `sandboxclaim/core`, `sandboxcr` and
`utils/runtime` all pass. One thing in the new helper.

`truncateError` slices `msg[:maxLen-len(ellipsis)]` on bytes, so a cut landing inside a multi-byte
character emits invalid UTF-8. A volume path with non-ASCII components is enough to reach it, and
stderr from a mount helper is exactly where such text arrives. Running the helper over
`failed to perform csi mount: stderr: ossfs: 挂载失败，权限不足，路径 /数据/卷`:

```
maxLen=47  validUTF8=true
maxLen=48  validUTF8=false   json -> "...ossfs: �..."
maxLen=49  validUTF8=false   json -> "...ossfs: ��..."
```

`encoding/json` accepts it and substitutes U+FFFD, so the effect is a replacement character
in `status.Message` and in the Warning event. That lands on the field whose whole purpose in this PR
is making a CSI failure readable.

Cutting back to a rune boundary covers it:

```go
cut := maxLen - len(ellipsis)
for cut > 0 && !utf8.RuneStart(msg[cut]) {
	cut--
}
return msg[:cut] + ellipsis
```

`truncateError` has no test in the diff, and the branch that truncates is the one left unexercised,
so a case with a multi-byte tail would cover the helper and this at the same time.

Smaller thing at `csi.go:78`: `stderr: %s` is formatted unconditionally, and on the transport-error
path `result.Stderr` is usually empty, so the message ends with a bare `stderr: `. Gating on
`len(result.Stderr) > 0` keeps that one readable.

AlbeeSo:
Thanks — both were real, and both are fixed in ae73909.

**The truncation.** I reproduced it before changing anything, and the boundary you found is exact:

```
maxLen=47  validUTF8=true   json -> "...r: ossfs: ..."
maxLen=48  validUTF8=false  json -> "...fs: \ufffd..."
maxLen=49  validUTF8=false  json -> "...fffd\ufffd..."
maxLen=50  validUTF8=true   json -> "...ossfs: 挂..."
```

Cutting back to a rune boundary as you suggested, and the test walks every limit the message can be cut at rather than a single case, asserting three things per limit: the result is valid UTF-8, it stays within the limit, and it survives `json.Marshal` without a replacement character.

I checked the test actually holds the fix down by reverting to the single slice — it fails and names `maxLen=48` and `maxLen=49`, the two values you reported. That also took `truncateError` from 42.9% to 100%, which covers the four lines codecov was pointing at.

**The dangling label.** Right, and it is worse than cosmetic on that path: the transport error is the case where the command never ran, so there is nothing to show and the message ended on `stderr: ` with nothing after it. Both call sites now go through a small helper that adds the label only when there is output, with cases for none, empty strings, one line and several.

One related thing I want to flag rather than fix here. What reaches `status.Message` is wrapped by every layer it crossed, and the wrapping is longer than the reason:

```
Internal: failed to perform csi mount: command failed: [1] stderr: Mount failed (costMs=1832):
mount failed for driver ossplugin.csi.alibabacloud.com: NodePublishVolume failed for driver
"ossplugin.csi.alibabacloud.com": rpc error: code = Internal desc = failed to mount: ossfs exited:
exit status 1, with stderr: ossfs: identity provider returned 403(endpoint=..., requestId=...)
```

Roughly 280 characters of wrapping around 110 characters of reason, with the driver name twice and five variations on "failed to mount". At a 256 byte limit the truncation keeps the wrapping and drops the reason, which defeats the point of the field.

I had a version that pulled the reason out by looking for the last `stderr: `, and dropped it: it pins this repo to a format string owned by the CSI driver. The wrapping is better reduced where it is added, so I would rather leave that to those layers than pattern-match here.

HARSHRAJ2789:
Confirmed both on `ae73909`. Running the original message through `truncateError` at every limit around the boundary, `maxLen` 48 and 49 now back off to 47 bytes and every case is valid UTF-8 with no replacement character in the marshalled output. `stderrSuffix` covers the transport path at both call sites.

On the wrapping problem you flagged, there is an option that does not need the format string. The reason sits at the end of the message and the wrapping accumulates at the front, so keeping a head and a tail preserves both without knowing what either contains. Taking the message you pasted and joining it to one line gives 410 bytes; at a 256 byte limit:

```
HEAD-ONLY (today)
Internal: failed to perform csi mount: command failed: [1] stderr: Mount failed (costMs=1832):
mount failed for driver ossplugin.csi.alibabacloud.com: NodePublishVolume failed for driver
"ossplugin.csi.alibabacloud.com": rpc error: code = Internal desc ...

HEAD + TAIL
Internal: failed to perform csi mount: command failed: [1] stderr: Mount failed (cos...ternal
desc = failed to mount: ossfs exited: exit status 1, with stderr: ossfs: identity provider
returned 403(endpoint=oss-cn-hangzhou.aliyuncs.com, requestId=6A1B2C3D)
```

Both are 256 bytes and both are valid UTF-8. The second keeps the outermost frame, which says where in the stack this came from, and the 403 that says why.

The rune backoff you just added applies to both cut points:

```go
budget := maxLen - len(ellipsis)
head := budget / 3
tail := budget - head
for head > 0 && !utf8.RuneStart(msg[head]) {
    head--
}
start := len(msg) - tail
for start < len(msg) && !utf8.RuneStart(msg[start]) {
    start++
}
return msg[:head] + ellipsis + msg[start:]
```

The split is arbitrary and a third to the head is only what I tried. Reducing the wrapping where it is added is still the better fix and this does not replace it, but it stops the field losing the reason while that happens.

AlbeeSo:
Heads up that the approach changed since your review, and the branch is now a single squashed commit (`32d023b9`), so the comments above point at code that no longer exists. The fix you asked for is still there — it moved.

**Where the rune boundary went.** Into `utils.TruncateConditionMessage`. `truncateError` is gone: it was a second answer to a question `pkg/utils` had already answered, and five conditions in the sandbox controller are held to that helper. One of them reports `err.Error()` from a mount path verbatim, so they had the same bug and now get the fix too. The test walks the character across every byte offset it can straddle the cut at, and removing the backoff makes it fail with `"aaa\xe4\xb8..."` — half a character — rather than passing quietly.

**Why the limit also had to go.** My helper picked 256 bytes. The project uses 1024, configurable through `MAX_CONDITION_MESSAGE_LEN`. That difference turned out to decide whether the message says anything at all.

I ran this against a real failure — an identity provider refusing a credential. The reason is 145 bytes; the whole message is 529, because it passes through four layers that each name themselves on the way out:

| | bytes |
|---|---|
| this repo (`failed to perform csi mount:` / `process error: exit status 1, stderr:`) | 67 |
| storage CLI (log header + two wrappers) | 131 |
| gRPC (`rpc error: code = Internal desc =`) | 34 |
| CSI driver + fuse client log framing | 152 |
| **the actual reason** | **145** |

At 256 the message ended inside the framing:

```
... rpc error: code = Internal desc = failed to mount: ossf...
```

At 1024 it arrives whole, `identity provider returned 500(endpoint=..., requestId=...)` included — with no change to the three repos upstream of it. Reducing that framing is worth doing on its own, but it is not what stands between a user and the reason.

One thing worth flagging: `ProcessCSIMounts` joins per-volume errors, so two failing volumes is ~1059 bytes and 1024 clips the tail. `MAX_CONDITION_MESSAGE_LEN` covers that case without a code change, which is part of why reusing the shared helper is better than a constant of my own.

PR #854, opened 2026-08-24
fix(e2b): persist configured admin key and skip unreadable Secret entries

AiRanthem:
## What changed

- Require `--e2b-admin-key` when `--e2b-enable-auth` is true. Empty admin keys are no longer auto-generated.
- Secret storage now reconciles the well-known admin key on Init: missing, rotated, or non-admin-team entries are rewritten to the configured key on the admin team. Unreadable `Secret.Data` entries are treated as absent on admin load, refresh, team lookup, and `ListLimited`.
- MySQL `ensureAdminKey` skips the update when the persisted row already matches.
- Conflict retries preserve both the original error and context cancellation via `errors.Join`. Exhausted Secret conflicts re-read the API server and succeed only if the desired admin key is already there.

## Why

Init previously only created the admin key when the canonical map slot was missing, so a configured rotation never overwrote a stale value. Unreadable Secret blobs also failed the whole `ListLimited` call and blocked quota anti-drift for every other key. Auto-generated admin keys made auth-enabled deployments start without an operator-known credential.

## Impact

- **Breaking:** deployments with E2B auth enabled must set `--e2b-admin-key`. In-tree `config/sandbox-manager` already does.
- Admin key rotation is config-authoritative. During a rolling update, a surviving old replica that restarts Init can write the old key back; replace all replicas before considering rotation complete.
- The well-known admin key stays on the admin team and is not given a quota.
- A corrupt Secret entry no longer fails listing limited keys; it is omitted as absent.

## Validation

```text
go test ./pkg/servers/e2b/keys/ -count=1
go vet ./pkg/servers/e2b/keys/
```

### DIFF
--- a/pkg/servers/e2b/keys/secret.go
+++ b/pkg/servers/e2b/keys/secret.go
@@ -116,6 +116,11 @@ func init() {
 
 // secretKeyStorage is a simple implement for api-key storage using k8s secret as storage backend.
 // It is only for demo purpose.
+//
+// Unreadable or otherwise invalid Secret.Data entries are treated as absent.
+// Admin load, refresh, team lookup, and list paths skip them and continue; they
+// must not fail the surrounding operation. Kubernetes API errors (missing
+// Secret, conflict, patch failure) remain real failures.
 type secretKeyStorage struct {
 	Namespace string
 	AdminKey  string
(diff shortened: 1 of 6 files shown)


pkg/servers/e2b/keys/secret.go
```diff
@@ -365,14 +404,17 @@ func (k *secretKeyStorage) retryPatchSecretKey(
 		patchedKey = apiKey
 		return nil
 	})
+	if ctxErr := ctx.Err(); ctxErr != nil {
```
furykerry, line 408: if err is nil, shall we skip the error reporting ? or just log a warning

HARSHRAJ2789: On the `err is nil` case: skipping the error looks right, because reporting it loses a key that
was already written.

`retryPatchSecretKey` returns `nil, ctxErr` whenever the retry loop succeeded and the
cancellation lands between the last successful patch and the check at `:408`. `CreateKey` passes
that straight out at `:551`. On `ff643b3`, letting the patch through and cancelling immediately
after it:

```
CreateKey returned=context canceled  created=false  entries persisted in Secret=1
```

The caller is told the key was not created while its entry sits in the Secret. `DeleteKey` at
`:563` has the same shape, where the cost is a delete reported as failed after it landed.

The window is narrow, since the cancellation has to arrive after the patch returns and before the
check. Returning `patchedKey, nil` when `err == nil`, and keeping the join for the `err != nil`
case, closes it.

PR #857, opened 2026-08-25
feat(sandbox-manager): load secret config at startup and add startup hook

AiRanthem:
## Summary
- Add `--secret-config` so sandbox-manager can read `E2B_ADMIN_KEY`, `E2B_KEY_STORAGE_DSN`, `E2B_KEY_HASH_PEPPER`, `QUOTA_REDIS_USERNAME`, and `QUOTA_REDIS_PASSWORD` from a Kubernetes Secret once at startup, overlaying flags and env vars. Missing keys fail fast; secret values are never written into error messages.
- Introduce a community no-op `startupHook` after the rest config and optional secret overlay, so internal builds can inject bootstrap work without changing the open-source main path.
- Share one controller-runtime client for the secret-config and runtime-client TLS Secret reads, and validate `--e2b-admin-key` only after the overlay so a Secret-provided admin key satisfies `--e2b-enable-auth`.

## Test plan
- [ ] Unit tests in `cmd/sandbox-manager` (`secretconfig_test.go`, `hook_test.go`) cover parse/load/validation, missing keys, ref forms, a single precise Get, and that errors omit secret values.
- [ ] Start sandbox-manager without `--secret-config` and confirm flags/env still apply.
- [ ] Start with `--secret-config=ns/name` pointing at a Secret that has all five keys and confirm those values override flags/env (changes require restart).
- [ ] Confirm a Secret missing any of the five keys fails startup without leaking values.
- [ ] With `--e2b-enable-auth`, confirm an empty flag/env admin key is accepted when the Secret supplies `E2B_ADMIN_KEY`, and still fails if the overlay leaves it empty.

### DIFF
--- a/cmd/sandbox-manager/main.go
+++ b/cmd/sandbox-manager/main.go
@@ -51,9 +51,9 @@ import (
 // and the data keys of the Secret referenced by --secret-config. The two sources
 // are intentionally spelled the same so operators can move a value between them.
 const (
-	E2BAdminKeySecretKey        = "E2B_ADMIN_KEY"
-	E2BKeyStorageDSNSecretKey   = "E2B_KEY_STORAGE_DSN"
-	E2BKeyHashPepperSecretKey   = "E2B_KEY_HASH_PEPPER"
+	E2BAdminKeySecretKey        = "E2B_ADMIN_KEY"       // #nosec G101 -- env-var/Secret data key name, not a credential
+	E2BKeyStorageDSNSecretKey   = "E2B_KEY_STORAGE_DSN" // #nosec G101 -- env-var/Secret data key name, not a credential
+	E2BKeyHashPepperSecretKey   = "E2B_KEY_HASH_PEPPER" // #nosec G101 -- env-var/Secret data key name, not a credential
 	QuotaRedisUsernameSecretKey = "QUOTA_REDIS_USERNAME"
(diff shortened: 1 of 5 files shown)


cmd/sandbox-manager/main.go
```diff
@@ -245,6 +272,35 @@ func main() {
...
+		klog.Fatalf("Failed to load secret config: %v", err)
+	}
+	e2bAdminKey = secretSettings.AdminKey
+	e2bKeyStorageDSN = secretSettings.KeyStorageDSN
+	e2bKeyStoragePepper = secretSettings.KeyHashPepper
+	quotaRedisUsername = secretSettings.RedisUsername
+	quotaRedisPassword = secretSettings.RedisPassword
+
+	if e2bEnableAuth && e2bAdminKey == "" {
+		klog.Fatalf("--e2b-admin-key is required when --e2b-enable-auth is true")
```
furykerry: the error message is now inaccurate, these flag may be provided in the secret

cmd/sandbox-manager/main.go
```diff
@@ -47,6 +48,7 @@ import (
 )
 
 const (
+	E2BAdminKeySecretKey     = "E2B_ADMIN_KEY"
```
furykerry: consider also change the following const name

PR #858, opened 2026-08-25
refactor(e2b): optional dedicated observability listener and drop empty debug endpoint

AiRanthem:
## Summary
- Add `--metrics-port` so sandbox-manager can serve `GET /metrics` and `GET /health` on a dedicated listener. `0` or the same value as `--port` keeps the current shared control-API listener; any other valid TCP port starts a separate observability server that shuts down together with the API server.
- Remove `GET /debug` and the Manager/Infra `DebugInfo` plumbing. That endpoint no longer returned useful data (`LoadDebugInfo` was already an empty map), and operators should use the Prometheus `/metrics` surface instead of a one-off debug dump.

## Test plan
- [ ] `go test ./cmd/sandbox-manager ./pkg/servers/e2b ./pkg/sandbox-manager` covers `--metrics-port` validation, dedicated vs shared listeners, and shutdown that waits for both HTTP servers before `manager.Stop`.
- [ ] Start without `--metrics-port` (or with `0` / the same port as `--port`) and confirm `/health` and `/metrics` are still on the control API listener.
- [ ] Start with `--metrics-port` set to a different port and confirm `/health` and `/metrics` are only on that listener, while control API routes stay on `--port`.
- [ ] Confirm `GET /debug` is gone.


pkg/servers/e2b/core.go
```diff
@@ -76,6 +78,10 @@ type ControllerOptions struct {
 	Domain string
 	// Port is the port the E2B HTTP server listens on.
 	Port int
+	// MetricsPort is the port for GET /metrics and GET /health. 0, a non-positive
+	// value, or the same value as Port serves them on the control API listener;
+	// any other positive port starts a dedicated observability listener.
```
furykerry: Issue: validateMetricsPort rejects negative values at startup, yet the ControllerOptions.MetricsPort doc explicitly promises "a non-positive value … reuses the control API listener", and core_test.go pins the negative case ("negative metrics port reuses API listener"). Two contradicting contracts for the same value.
Recommendation: Pick one contract. Simplest: tighten the option doc and test to "0 or the same value as Port"

PR #882, opened 2026-08-25
feat(api): add named HTTP callouts to SecurityProfile

Kuromesi:

Add `actions.httpCallout` to SecurityProfile and GlobalSecurityProfile. Rules reference an operator-managed EPEConfig provider through `provider.name`, so provider URL, timeout, TLS, and decision-response limits are configured once instead of repeated in every profile.

Provider references use the shared `ExtensionProviderRef` type (`name` only); the consuming action determines the required provider kind.

The action retains per-rule request/response phases, header disclosure modes, `body: true/false`, `maxBodyBytes`, and the existing `failStrategy` enum (default `Block`). At least one phase is required; header lists must agree with the selected mode. Provider names are required and validated as DNS subdomains.

Regenerate DeepCopy and both CRD schemas. Configuration examples are included in the API field comments.

NONE

- `make generate manifests`
- `go test ./api/v1alpha1` (package compilation; no package tests)
- `git diff --check`
- Review the HTTPCallout example in `api/v1alpha1/securityprofile_types.go`.

This replaces the unreleased `externalProcessing.backend.http` API with `httpCallout.provider.name`. Existing released actions are unchanged. `body` remains a boolean; `maxBodyBytes` remains on the action.

Provider availability is checked by the serving EPE at runtime, not by CRD admission. The API does not introduce a namespace or kind on the provider reference. Agentio runtime conversion and production filter registration will follow the published API; adding this CRD field alone does not enable execution.

The EPEConfig registry and configuration watch implementation is in https://github.com/openkruise/agentio/pull/93.


l1b0k (approved):

api/v1alpha1/securityprofile_types.go
```diff
@@ -375,6 +385,103 @@ type TokenTransformationAction struct {
...
+	Mode HTTPCalloutHeaderMode `json:"mode,omitempty"`
+	// AllowList names the only headers sent when Mode is AllowList. Header
+	// names are case-insensitive.
+	// +optional
+	// +listType=set
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=256
+	// +kubebuilder:validation:items:Pattern=`^[A-Za-z0-9!#$%&'*+\-.^_|~]+$`
+	AllowList []string `json:"allowList,omitempty"`
```
l1b0k: 语义，如果  allow list 为空，表达什么意思；

PR #894, opened 2026-08-28
feat(controller): emit CommitRunning and CommitSucceeded events for Commit lifecycle

Luckydog691:

The Commit controller currently only emits Warning events on edge failure paths (`PodNotFound`, `JobGenerationFailed`, `JobNotFound`). Lifecycle transitions are invisible: a user watching `kubectl describe commit` cannot tell when the commit Job actually started, when the image was successfully pushed, or when the job failed.

This PR emits events on the lifecycle transitions:

- **CommitRunning** (Normal) — the commit Job has been created and the Commit entered `Running`. Message: `Commit Job started for pod <podName>`.
- **CommitSucceeded** (Normal) — the commit Job completed and the image was pushed. Message: `Commit completed successfully, image <image> pushed`.
- **CommitFailed** (Warning) — the commit Job completed unsuccessfully (e.g. registry push rejected, non-zero exit). Message: `Commit job failed for pod <podName>`. This covers the previously-invisible job-failure path; `JobNotFound` and `JobGenerationFailed` keep their own specific Warning events.

Implementation notes:

- Events are emitted only after the phase transition is **successfully persisted** (`err == nil` from `ensureAndApply`). If the status update fails, the phase stays put in the API server, so the retry re-enters the same handler and emits then. Events are best-effort observability signals.
- Emission cannot repeat across reconciles: after Pending→Running the next reconcile is routed to `handleCommitRunning`, and terminal phases (Succeeded/Failed) are short-circuited in `Reconcile`.

NONE

### Ⅲ. Describe how you did it

Unit tests cover all paths:
- `TestReconcile_CommitPhasePending_EmitsCommitRunningEvent` — Pending→Running emits `CommitRunning`
- `TestReconcile_CommitPhaseRunning_EmitsCommitSucceededEvent` — Running→Succeeded emits `CommitSucceeded`
- `TestReconcile_CommitPhaseRunning_EmitsCommitFailedEvent` — Running→Failed emits `CommitFailed`
- `TestReconcile_CommitPhasePending_NoEventWhenPhaseUnchanged` / `TestReconcile_CommitPhaseRunning_NoEventWhenPhaseUnchanged` — no event while the phase stays put
- `TestReconcile_CommitPhasePending_NoEventOnError` — no event when the ensure function errors

### Ⅳ. Describe how to verify it

```bash
go test ./pkg/controller/commit/... -count=1
```

On a live cluster, `kubectl describe commit <name>` now shows:

```
Events:
  Type    Reason           Age   From               Message
  ----    ------           ----  ----               -------
  Normal  CommitRunning    5s    commit-controller  Commit Job started for pod my-sandbox-pod
  Normal  CommitSucceeded  83s   commit-controller  Commit completed successfully, image registry.example.com/env:v1 pushed
```

or on failure:

```
  Warning CommitFailed     83s   commit-controller  Commit job failed for pod my-sandbox-pod
```

### Ⅴ. Special notes for reviews

- No behavioral change beyond the `Recorder.Eventf` calls; status flow, requeue and error handling are untouched.
- `core` RBAC for events already exists (`+kubebuilder:rbac:groups=core,resources=events,verbs=create`).
- The `CommitFailed` event is added per review feedback to make the lifecycle symmetric (thanks @furykerry); the `JobNotFound` path will show both its specific event and `CommitFailed`.

### DIFF
--- a/pkg/controller/commit/commit_controller.go
+++ b/pkg/controller/commit/commit_controller.go
@@ -213,11 +213,29 @@ func (r *CommitReconciler) handleCommitPending(ctx context.Context, args *core.E
 		r.Recorder.Eventf(commit, corev1.EventTypeWarning, "PodNotFound", "Target pod %s not found or deleting", commit.Spec.PodName)
 		return ctrl.Result{}, r.updateCommitStatus(ctx, *args.NewStatus, commit)
 	}
-	return r.ensureAndApply(ctx, args, control.EnsureCommitRunning)
+	result, err := r.ensureAndApply(ctx, args, control.EnsureCommitRunning)
+	// Emit only after the transition is persisted: on a failed status update
+	// the retry re-enters this handler and emits then. Events are best-effort.
+	if err == nil && args.NewStatus.Phase == agentsv1alpha1.CommitPhaseRunning {
+		r.Recorder.Eventf(commit, corev1.EventTypeNormal, "CommitRunning",
+			"Commit Job started for pod %s", commit.Spec.PodName)
+	}
+	return result, err
 }
 
 func (r *CommitReconciler) handleCommitRunning(ctx context.Context, args *core.EnsureFuncArgs, control core.CommitControl) (ctrl.Result, error) {
-	return r.ensureAndApply(ctx, args, control.EnsureCommitUpdated)
+	result, err := r.ensureAndApply(ctx, args, control.EnsureCommitUpdated)
+	// Terminal phases are skipped in Reconcile, so these fire once per
+	// Running -> Succeeded/Failed transition.
+	if err == nil && args.NewStatus.Phase == agentsv1alpha1.CommitPhaseSucceeded {
+		r.Recorder.Eventf(args.Commit, corev1.EventTypeNormal, "CommitSucceeded",
+			"Commit completed successfully, image %s pushed", args.Commit.Spec.Image)
+	}
+	if err == nil && args.NewStatus.Phase == agentsv1alpha1.CommitPhaseFailed {
+		r.Recorder.Eventf(args.Commit, corev1.EventTypeWarning, "CommitFailed",
+			"Commit job failed for pod %s", args.Commit.Spec.PodName)
+	}
+	return result, err
 }
 
 // ensureAndApply calls the given ensure function, then applies the resulting status.
--- a/pkg/controller/commit/commit_controller_test.go
+++ b/pkg/controller/commit/commit_controller_test.go
@@ -19,6 +19,7 @@ package commit
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


Luckydog691:
The `sandbox (quota redis fault)` failure is unrelated to this change — it is a pre-existing flake in that matrix job.

**What failed**

```
FAILED test/e2b/test_sandbox.py::test_pause_connect_kill - httpx.ReadTimeout: timed out
1 failed, 50 passed, 4 skipped
```

The test pauses a sandbox, waits 30s, then calls `connect` (resume-on-connect). The POST hung for the full 120s client timeout with no server response.

**Why it cannot be this PR**

- This PR only adds two `Recorder.Eventf` calls in the Commit controller (`handleCommitPending` / `handleCommitRunning`), which only execute when a Commit CR transitions phase.
- `test_pause_connect_kill` never creates a Commit CR; the Commit reconciler is not in the pause/resume/connect path at all.

**Cross-PR evidence (same failure, same job, ~10 min earlier)**

| Run | PR | Failure |
|---|---|---|
| [33139105405](https://github.com/openkruise/agents/actions/runs/33139105405) | dependabot: bump otel group (deps-only change) | `test_pause_connect_kill - httpx.ReadTimeout` in `sandbox (quota redis fault)` |
| [33140604466](https://github.com/openkruise/agents/actions/runs/33140604466) | this PR | identical failure |

A dependencies-only PR failing the exact same spec rules out any controller-code causal link. Other PRs in the same window also hit different flakes in this workflow family (e.g. `test_wake_on_traffic` port-forward drop, `test_redis_data_loss_rebuilds...` assertion).

**Likely mechanism**

`test_api_key_quota.py` runs before `test_sandbox.py` (alphabetical) and in this matrix profile its fault tests scale Redis to 0 and back (`redis_stopped_temporarily`), while the manager runs with `--quota-redis-operation-timeout=1s` and anti-drift interval/grace of 30s/60s. The synchronous resume-on-connect request appears to exceed the 120s SDK timeout when Redis/quota state is still settling or pod rescheduling is slow on the shared runner.

I could not re-run the failed job from my account (needs admin rights), so I force-pushed the branch (same single commit, new SHA) to trigger a fresh run. Maintainers, please feel free to re-run just the failed job instead if you prefer.

Luckydog691:
The `golangci-lint` check on the re-triggered run failed due to a CI infrastructure issue, not a lint problem.

From the [job log](https://github.com/openkruise/agents/actions/runs/33145567437/job/98765696303), the failure happened while **downloading the golangci-lint binary itself** — the linter never started and no code was analyzed:

```
Downloading binary https://github.com/golangci/golangci-lint/releases/download/v2.9.0/golangci-lint-2.9.0-linux-amd64.tar.gz ...
##[error]Failed to run: Error: Unexpected HTTP response: 403
```

Evidence this is transient:
- The previous run of this same branch ([33140604403](https://github.com/openkruise/agents/actions/runs/33140604403), identical code, commit 012a20c) passed `golangci-lint` cleanly.
- Master ([33144456279](https://github.com/openkruise/agents/actions/runs/33144456279)) and other PRs that ran in the same window also passed.

Since fork PRs cannot re-run jobs (requires admin rights), it would be great if a maintainer could re-run the failed `golangci-lint` job when convenient. Alternatively I can re-push to re-trigger, but that would cancel the in-progress E2E matrix runs.

pkg/controller/commit/commit_controller.go
```diff
@@ -213,11 +213,26 @@ func (r *CommitReconciler) handleCommitPending(ctx context.Context, args *core.E
 		r.Recorder.Eventf(commit, corev1.EventTypeWarning, "PodNotFound", "Target pod %s not found or deleting", commit.Spec.PodName)
 		return ctrl.Result{}, r.updateCommitStatus(ctx, *args.NewStatus, commit)
 	}
-	return r.ensureAndApply(ctx, args, control.EnsureCommitRunning)
+	result, err := r.ensureAndApply(ctx, args, control.EnsureCommitRunning)
+	// Only emit the event after the phase transition is persisted: the next
+	// reconcile is routed to handleCommitRunning, so this fires once. On a
+	// failed status update err != nil and the event is deferred to the retry.
```
furykerry: **Warning: the "deferred to the retry" claim is wrong for conflict failures — the event can be silently lost.**

Trace through the actual race:
1. Reconcile #1 (Pending): creates the Job, patches status→Running, emits `CommitRunning`.
2. A Job watch trigger fires reconcile #2 while the informer cache still shows Pending: expectations are satisfied (Job observed in the list), `EnsureCommitRunning` takes the "job already exists" branch and sets Running, but `updateCommitStatus` issues a `MergeFrom(commit)` status patch whose base carries the stale `resourceVersion` → **409 Conflict** → `err != nil` → no event.
3. The retry re-fetches, sees phase Running, and is routed to `handleCommitRunning` — so `CommitRunning` is **never emitted**, not deferred.

The same race can drop `CommitSucceeded` on a conflicting patch in `handleCommitRunning`. Duplicates are impossible here (stale phase ⟹ stale RV ⟹ conflict), so the PR's "exactly once" holds or degrades to **zero**, never more.

Impact is observability-only and the window is narrow, so not blocking — but the comment shouldn't assert a guarantee the code doesn't have. Suggested rewrite:

```go
// Only emit after the phase transition is persisted: the next reconcile is
// routed to handleCommitRunning, so this normally fires once. On a failed
// status update the event is skipped; if the transition was nevertheless
// persisted (e.g. conflict), the event may be dropped — acceptable since
// events are best-effort observability signals.
```

pkg/controller/commit/commit_controller.go
```diff
@@ -213,11 +213,26 @@ func (r *CommitReconciler) handleCommitPending(ctx context.Context, args *core.E
...
+	}
+	return result, err
 }
 
 func (r *CommitReconciler) handleCommitRunning(ctx context.Context, args *core.EnsureFuncArgs, control core.CommitControl) (ctrl.Result, error) {
-	return r.ensureAndApply(ctx, args, control.EnsureCommitUpdated)
+	result, err := r.ensureAndApply(ctx, args, control.EnsureCommitUpdated)
+	// Terminal phases are skipped in Reconcile, so the success event fires on
+	// the single Running -> Succeeded transition and never again.
+	if err == nil && args.NewStatus.Phase == agentsv1alpha1.CommitPhaseSucceeded {
```
furykerry, line 230: **Warning: Running→Failed still emits nothing — the lifecycle is asymmetric.**

The PR's motivation is making lifecycle transitions visible in `kubectl describe commit`, but when the commit Job completes unsuccessfully, `EnsureCommitUpdated` (`core/common_control.go`) sets phase `Failed` with no event — users still see nothing on exactly the case they most need visibility. Only `JobNotFound` / `JobGenerationFailed` paths have Warning events today.

Recommend adding a `CommitFailed` Warning event here for symmetry:

```go
if err == nil && args.NewStatus.Phase == agentsv1alpha1.CommitPhaseFailed {
    r.Recorder.Eventf(args.Commit, corev1.EventTypeWarning, "CommitFailed",
        "Commit job failed for pod %s", args.Commit.Spec.PodName)
}
```

If you'd rather keep this PR scoped to the happy path, please file a follow-up issue and note it in the PR description.

furykerry (commented):
Overall this is a clean, correctly-gated observability enhancement: events fire only after the transition is persisted (`err == nil` + explicit phase check), which correctly suppresses them on the generation-failure path, and the once-per-transition reasoning holds against `Reconcile`'s routing. Verified locally: `go test ./pkg/controller/commit/... -count=1` passes including all 5 new tests; `gofmt`/`go vet` clean. Two warning-level findings inline below (W1: comment overstates a guarantee; W2: Running→Failed still emits nothing).

Luckydog691:
Thanks @furykerry for the thorough review — both points addressed in 18c9e23:

**W1 (comment overstates the guarantee):** Accepted, the comment is rewritten and no longer claims the event is deferred on any failed update. One factual note on the proposed mechanism: `updateCommitStatus` uses plain `client.MergeFrom(commit)`, not `MergeFromWithOptimisticLock`, so the status patch carries no resourceVersion precondition and a stale-base patch does not produce a 409 — it applies idempotently. The deviation actually runs the other way: a stale-cache double reconcile can re-apply the transition and emit a rare duplicate (at-least-once rather than at-most-once), which the event recorder's aggregation absorbs. The narrow true-loss window is patch-landed-but-response-lost (`err != nil` after the write hit the server). The new comment and PR description just say: events are best-effort.

**W2 (Running→Failed emits nothing):** Agreed this was the real gap — implemented the suggested `CommitFailed` Warning event in `handleCommitRunning`, with `TestReconcile_CommitPhaseRunning_EmitsCommitFailedEvent` covering it. Kept it in this PR rather than a follow-up since the change is small and makes the lifecycle symmetric in one shot. Note the `JobNotFound` path now emits both its specific event and `CommitFailed` (reason + transition marker), which seems fine for `kubectl describe`; happy to suppress `CommitFailed` there if you prefer.

Note the previous CI run's `golangci-lint` failure was the CDN 403 on downloading the linter binary (details in my earlier comment) — the re-push above also re-triggers it, and all E2E matrix jobs (including `sandbox (quota redis fault)`) passed on the previous run with identical controller code.

PR #895, opened 2026-08-28
feat(poolautoscaler): coordinate scale-up execution

### LINKED ISSUE #852: bug: SandboxSet maxUnavailable does not recognize failing sandboxes during scale-up
### Background

`SandboxSet.spec.scaleStrategy.maxUnavailable` is documented as a way to limit the rate of replica changes when scaling up, so warm pools can be expanded without overwhelming the cluster. In practice it only caps the number of concurrently in-flight unavailable creations. It does not detect whether the sandboxes being created are failing persistently, so when the template cannot start at all, scale-up keeps producing failing sandboxes at the maxUnavailable pace.

### Issue Type

bug

### Relevant Code

- `api/v1alpha1/sandboxset_types.go` — `SandboxSetScaleStrategy.MaxUnavailable`: "The scale will fail if the number of unavailable sandboxes were greater than this MaxUnavailable at scaling up."
- `pkg/controller/sandboxset/sandboxset_controller.go` — `calculateScaleDelta` only subtracts currently unavailable sandboxes (`Replicas - AvailableReplicas`) from `maxUnavailable` when computing the scale-up delta.
- `pkg/utils/utils.go` — `GetSandboxState` maps `Phase=Failed` sandboxes to the `Dead` state; `calculateSandboxSetStatusFromGroup` excludes them from `status.Replicas`, and `deleteDeadSandboxes` deletes them immediately.

### Notes

Failure scenarios that are not recognized include:

- Insufficient cluster resources (pod cannot be scheduled)
- Admission/quota rejection when creating the underlying resources
- Image pull failures (ErrImagePull / ImagePullBackOff)
- Sandboxes that are created but never reach Running for a long time

Observed consequences:

- A sandbox that fails into `Phase=Failed` is garbage-collected and excluded from the replica counts, so the controller creates a replacement in the next reconcile. This create → fail → delete → recreate cycle continues indefinitely at maxUnavailable concurrency. There is no failure counting, backoff, or circuit-breaking anywhere in this path.
- Sandboxes stuck in Pending are not `Failed`, so they stay counted as Creating and occupy the unavailable budget forever, with no status, condition, or event indicating why.
- When the pool cannot warm up, scale-up amplifies the scale of abnormal sandboxes instead of stopping. Each cycle adds load to the API server, scheduler, admission webhooks, and node-level components, which may further affect cluster stability on an already strained cluster.

ywExcellent:

Adds a scale-up execution coordination path between `SandboxSet` and `PoolAutoscaler` so that scale-up decisions do not keep piling on when the current startup budget is already exhausted by failing or long-pending sandboxes.

- `SandboxSet` publishes a new `ScalingLimited` condition on `.status`. It is set to `True` (with `observedGeneration` recorded) when the union of `StartContainerFailed` sandboxes and `Pending` sandboxes exceeding the process-wide max pending timeout has consumed the startup budget derived from `MaxUnavailable`.
- `PoolAutoscaler` observes that condition in `sandboxSetAllowsScaleUp` and gates scale-up only when `ScalingLimited=True` is reported against the current `SandboxSet` generation. Missing, `Unknown`, or stale reports are treated as fail-open so upgrades and first-time reconciles do not stall.
- Splits scale-up and scale-down cooldown handling in `PoolAutoscaler`; scale-up cooldown is normalized against the sandbox controller's pending timeout with a small safety margin (minimum 65s).
- New `--max-pending-timeout` flag on `agent-sandbox-controller` (default `60s`, clamped to `[15s, 3595s]`) drives the pending-side of the startup-failure classification.
- `calculateScaleDelta` keeps its historical contract: `MaxUnavailable` percentage still resolves against `Spec.Replicas` for the physical scale-up step. The new startup-budget accounting resolves the same percentage against observed replicas, and `resolveMaxUnavailable` documents which caller uses which base.
- Exports `SandboxStateReasonResourcePending` in `api/v1alpha1` to remove duplicated `"ResourcePending"` string literals.
- Proposal: `docs/proposals/20260827-sandboxset-scale-up-execution-coordination.md`.

fixes #852

```
go test ./pkg/controller/sandboxset/... -count=1
go test ./pkg/controller/poolautoscaler/... -count=1
go test ./pkg/controller/sandbox/... -count=1
```

Key new/updated tests: `TestCalculateScalingLimited`, `TestResolveStartupBudget`, `TestSandboxSetAllowsScaleUp`, `TestResolveScaleUpCooldown`, `TestMaxPendingTimeout`, and expanded `TestReconcile` cases in `PoolAutoscaler` covering the `ScalingLimited` gate.

- Please read the proposal first: `docs/proposals/20260827-sandboxset-scale-up-execution-coordination.md`.
- The scale-up gate is deliberately fail-open: only an explicit `ScalingLimited=True` observed against the current generation blocks scale-up. This avoids stalling when the condition has not yet been published (e.g. right after upgrade or on a fresh `SandboxSet`).
- `isStartupFailure` in `SandboxSet` only counts `SandboxReadyReasonStartContainerFailed` as a hard startup failure. Other pre-Ready conditions (image pulling, scheduling) are handled by the pending-timeout path so transient states do not immediately consume the budget.
- No new Prometheus metrics were added, per `AGENTS.md`.
- `resolveMaxUnavailable` accepts a `base` parameter that two callers use differently — physical scale-up step uses `Spec.Replicas`, startup-budget accounting uses observed replicas. This is intentional and documented on the function.
- CRD manifests were regenerated via `make manifests`. No clientset changes were needed; `client/` is unchanged.


api/v1alpha1/poolautoscaler_types.go
```diff
@@ -130,7 +130,9 @@ type CapacityScalingRules struct {
 	// action is immediate, subsequent actions must wait for the window to elapse since
 	// the most recent scale action (in either direction).
 	// Must be >= 0 and <= 3600 (one hour).
-	// Defaults: scale up: 0, scale down: 300.
+	// Scale-up defaults to 65 and is normalized at runtime to at least the
```
furykerry: 65 is a weird default number, can we change ti to 60?

furykerry:
### Code review: Critical findings

Reviewed at `ea3880a2f`. Verified locally: `go build ./...` clean, `gofmt -l` clean on all changed packages, and `go test ./pkg/controller/sandboxset/... ./pkg/controller/poolautoscaler/... -count=1` both `ok`.

The overall design is sound — in particular the **fail-open** gate is the right call, since `doScale` patches `spec.replicas` and therefore bumps `metadata.generation`, so a stale `ScalingLimited` report is the *normal* state immediately after every scale. A fail-closed gate would deadlock. Pairing that one-round-trip fail-open window with the new 65s scale-up cooldown is a thoughtful way to close it.

Below are the 3 issues I'd consider blocking. Warnings and suggestions are omitted here to keep the thread focused.

---

## 1. An invalid `maxUnavailable` now permanently wedges the entire SandboxSet reconcile

**Where:** `pkg/controller/sandboxset/scaling_limited.go:69-72`, propagated at `pkg/controller/sandboxset/sandboxset_controller.go:147-151` and `:171-175`

`calculateScaleDelta` previously discarded the `GetScaledValueFromIntOrPercent` error (`scaleMaxUnavailable, _ = ...`). Both new call sites now propagate it, and `Reconcile` returns *before* the status update, the scale step, and dead-sandbox GC.

This error is non-retriable — it comes from a malformed spec value that no amount of requeueing can fix — and it is reachable: `maxUnavailable` is `x-kubernetes-int-or-string` with **no pattern validation** (`config/crd/bases/agents.kruise.io_sandboxsets.yaml:116-125`), and there is no validating webhook for it, only a defaulter (`pkg/webhook/sandboxset/mutating/sandboxset_defaulter.go:111`) that fires solely when the field is `nil`. So `maxUnavailable: "50"` — a plausible YAML typo, since a bare string without `%` is rejected by `GetScaledValueFromIntOrPercent` — is accepted by the API server.

Before this PR that degraded to "scale-up capped at 0" while scale-down, dead-sandbox GC, status, and the scale-subresource selector all kept working. After this PR the object is frozen entirely and retried forever with exponential backoff, with no signal on `.status`.

Suggest surfacing it and degrading rather than aborting:

```go
startupBudget, err := resolveStartupBudget(sbs.Spec.ScaleStrategy.MaxUnavailable, status.Replicas)
if err != nil {
    // Non-retriable spec error: requeueing cannot fix it. Surface it and treat the
    // budget as unconstrained so the rest of the reconcile still runs.
    logf.FromContext(ctx).Error(err, "invalid scaleStrategy.maxUnavailable, skipping ScalingLimited accounting",
        "maxUnavailable", sbs.Spec.ScaleStrategy.MaxUnavailable)
    r.Recorder.Eventf(sbs, corev1.EventTypeWarning, "InvalidMaxUnavailable",
        "Invalid scaleStrategy.maxUnavailable %q: %v", sbs.Spec.ScaleStrategy.MaxUnavailable.String(), err)
    return 0, nil
}
```

The same treatment applies to the `calculateScaleDelta` error return at `sandboxset_controller.go:171-175`. Better still: add pattern validation (or a validating webhook) for `maxUnavailable` so the value can never reach the controller malformed — then both errors become genuinely unreachable.

## 2. The proposal added in this PR mandates fail-closed; the implementation is fail-open

**Where:** `docs/proposals/20260827-sandboxset-scale-up-execution-coordination.md:409` and `:556-557` vs. `pkg/controller/poolautoscaler/poolautoscaler_controller.go:379-394`

The proposal at `:409` states "Both Capacity and Cron scale-up require current-generation `ScalingLimited=False`", and the test plan at `:556-557` says "Verify Capacity and Cron increases **stop for missing, stale, `Unknown`, or `True`** `ScalingLimited`."

`sandboxSetAllowsScaleUp` does the opposite for three of those four cases — missing, stale, and `Unknown` all `return true, ""`. Only an explicit `True` against the current generation blocks.

This is not cosmetic drift: `docs/proposals/` is the normative design record per `AGENTS.md`, so a future maintainer "fixing the code to match the proposal" would introduce a hard deadlock. The implementation is right; the proposal text is wrong. Suggest correcting it, e.g.:

```markdown
- Capacity and Cron scale-up are blocked only by an explicit `ScalingLimited=True`
  observed against the current generation. Missing, stale, and `Unknown` conditions
  fail open: `doScale` bumps `metadata.generation`, so a stale report is the normal
  state immediately after any scale, and failing closed would deadlock scale-up
  whenever the SandboxSet controller lags or is unavailable.
```

The test-plan bullet at `:556-557` needs the same correction.

## 3. A gate-blocked scale-up leaves no signal on the PoolAutoscaler, and `reason` is a dead assignment

**Where:** `pkg/controller/poolautoscaler/poolautoscaler_controller.go:314-323`

The gate sets `limited, limitReason = false, ""` and then `reason = gateReason`. `reason` reaches `updateStatus(..., limitMessage, ...)` -> `setConditions`, which consumes `limitMessage` **only inside `if limited {`** (`:556-564`). Since `limited` was just forced to `false` on the previous line, `gateReason` is silently dropped. The `ScaledUp` event at `:346` is also skipped because `desiredReplicas == specReplicas`.

Net result: autoscaling silently stops with no condition, no event, and no log line on the `PoolAutoscaler`. `kubectl describe poolautoscaler` shows `ScalingActive=True`, `AbleToScale=True`, `ScalingLimited=False` while scale-up is in fact suppressed. The stabilization-window branch 10 lines above emits a `ScaleBlocked` event for exactly this reason. The information exists only on the `SandboxSet`, which is not where someone debugging their autoscaler looks.

Suggest mirroring the stabilization branch:

```go
if desiredReplicas > specReplicas {
    if allowed, gateReason := sandboxSetAllowsScaleUp(sbs); !allowed {
        desiredReplicas = specReplicas
        limited, limitReason = false, ""
        reason = gateReason
        logf.FromContext(ctx).Info("scale-up suppressed by SandboxSet startup budget",
            "sandboxSet", sbs.Name, "reason", gateReason)
        r.recorder.Eventf(pa, corev1.EventTypeNormal, "ScaleBlocked",
            "Scaling %s/%s blocked: %s", pa.Namespace, pa.Spec.ScaleTargetRef.Name, gateReason)
    }
}
```

Worth extending the existing `"scale up blocked by SandboxSet ScalingLimited"` case in `TestReconcile` to assert the event, so this plumbing cannot silently rot again.

api/v1alpha1/poolautoscaler_types.go
```diff
@@ -130,7 +130,9 @@ type CapacityScalingRules struct {
...
+	// process-wide Sandbox Pending timeout plus 5 seconds.
+	// Scale-down defaults to 300.
```
furykerry: we can enforce a more reasonable default for StabilizationWindowSeconds in the validation webhook

HARSHRAJ2789:
This and #881 both declare `fixes #852`, and neither references the other, so the two of us are
adding condition vocabulary to `SandboxSetStatus.Conditions` for the same signal. Flagging it here
because the API type keeps whatever both leave behind.

This PR adds `ScalingLimited`, typed through a new `SandboxSetConditionType`, with
`StartupBudgetExhausted` and `StartupBudgetAvailable`. It goes True when `StartContainerFailed`
sandboxes together with those sitting in `Creating`/`ResourcePending` past the pending timeout
consume the budget derived from `MaxUnavailable`.

#881 adds `ScaleUpBlocked`, with `TerminalFailures` and `WithinBudget`, counting `Phase=Failed` over
a rolling window; and `Progressing`, with `SandboxesStuckCreating` and `PoolProgressing`, for
sandboxes that never leave the creating state.

The overlap is the stuck-creating half. `ScalingLimited` covers it through
`Creating`/`ResourcePending` and `Progressing` covers it through a creating timeout, and both
conditions gate scale-up, here via `sandboxSetAllowsScaleUp` and in #881 at
`sandboxset_controller.go:202`. The failure halves are different, since this PR reads the
`StartContainerFailed` ready-condition reason while #881 counts terminal `Phase=Failed`.

I test-merged the two branches and they apply cleanly to each other, so nothing forces a decision.
That is what makes it worth raising: if both land, `SandboxSetStatus` carries three conditions
answering one question, under two naming conventions, since this PR introduces a typed
`SandboxSetConditionType` where #881 uses plain string constants.

So the question is which condition model the type should carry. I am not asking for #881 over this
one. If `ScalingLimited` is the model you want, I am happy to close #881, or to rebuild its
stuck-creating detection on top of this condition, whichever is more useful.

api/v1alpha1/poolautoscaler_types.go
ywExcellent: Parameters have been added, allowing for adjustments during the verification process

api/v1alpha1/poolautoscaler_types.go
ywExcellent: Done in 40fadae — the scale-up default is now 60s (the associated `fallbackScaleUpWindow` constant and the doc comment on `StabilizationWindowSeconds` were updated together). The runtime floor still normalizes to `maxPendingTimeout + 10s` so misconfigurations are pulled up rather than causing scale-up storms.

api/v1alpha1/poolautoscaler_types.go
ywExcellent: Good call. For this PR we kept the enforcement at the controller entrypoint — `validateDefaultStabilizationWindows` clamps the process-wide `--default-scale-up-window` / `--default-scale-down-window` flags into [0, 1h] at startup (falling back to 60s / 300s on out-of-range values), and the per-object window is further raised to `maxPendingTimeout + 10s` at runtime so a too-small value cannot cause scale-up storms. A validating webhook is the right long-term home for user-facing spec defaults/rejection, but the PoolAutoscaler CRD doesn't ship one yet; happy to spin that up as a follow-up so the same policy is enforced at admission time.

ywExcellent:
Thanks for the thorough pass — all three critical findings are addressed in bdd5483 + 818b578 + c747853 (rebased onto latest `master`; a Requeuing typo fix picked up by the `typos` gate is folded into c747853). Details inline:

**1. Invalid `maxUnavailable` no longer wedges the SandboxSet reconcile.**
Both call sites now degrade instead of aborting:
- `calculateScalingLimited` error path in `pkg/controller/sandboxset/sandboxset_controller.go` calls `reportInvalidMaxUnavailable(ctx, sbs, err, "startup-budget accounting")` and continues.
- `calculateScaleDelta` error path calls `reportInvalidMaxUnavailable(ctx, sbs, err, "scale-up for this pass")`, forces `delta = 0` and clears the shared `err` (c747853) so the downstream `allErrors` join does not re-surface the same non-retriable error out of `Reconcile`.
- `reportInvalidMaxUnavailable` emits a `Warning InvalidMaxUnavailable` event but suppresses duplicates via a `(uid, generation)` cache, so continuous child-Sandbox reconciles do not flood the event stream.
- End-to-end coverage: `TestReconcile_InvalidMaxUnavailable` asserts no error is returned, status is written, and exactly one warning is emitted per generation. `TestReconcile_ClearsInvalidMaxUnavailableOnNotFound` covers the eviction on delete. A validating webhook / CRD pattern for `maxUnavailable` is a good follow-up but out of scope for this PR.

**2. Proposal now matches the fail-open implementation.**
`docs/proposals/20260827-sandboxset-scale-up-execution-coordination.md` was updated to state that only an explicit current-generation `ScalingLimited=True` blocks scale-up, and that missing / stale / `Unknown` conditions fail open (both the "User-Visible Scaling Behavior" section and the PoolAutoscaler unit-test bullet). The wording matches your suggested language.

**3. Gate-blocked scale-up now leaves a first-class signal on the PoolAutoscaler.**
`Reconcile` in `poolautoscaler_controller.go` now:
- emits `klog.Info("scale-up suppressed by SandboxSet startup budget", ...)`,
- records a `Normal ScaleBlocked` event via `reportScaleBlocked`, deduped by `(paUID, sbsUID, sbs.Generation, condition.Reason, condition.LastTransitionTime)` so a steady blocked state does not spam events, and cleared as soon as the gate re-opens or the PA is deleted,
- stays silent when SandboxSet exposes no `ScalingLimited` condition at all (fail-open policy).

Tests added to lock this in:
- `TestReconcile` case "scale up blocked by SandboxSet ScalingLimited" now asserts `expectEventReason: "ScaleBlocked"`.
- `TestReportScaleBlockedDeduplicatesEvents` covers initial + duplicate + LastTransitionTime rotation + PA recreation.
- `TestReportScaleBlocked_NoConditionNoEvent` guards the fail-open path.
- `TestReconcile_ClearsScaleBlockedOnAllowed` / `TestReconcile_ClearsScaleBlockedOnNotFound` cover the two cache-eviction sites in `Reconcile`.

Two inline replies on the line comments (default 65s → 60s, and webhook-side enforcement) are also up. PTAL.

pkg/controller/sandboxset/scaling_limited.go
```diff
@@ -0,0 +1,142 @@
...
+			failed++
+			continue
+		}
+
+		state, reason := utils.GetSandboxState(sandbox)
+		if state != agentsv1alpha1.SandboxStateCreating || reason != agentsv1alpha1.SandboxStateReasonResourcePending {
+			continue
+		}
+		deadline := sandbox.CreationTimestamp.Add(sbxMaxPendingTimeout)
+		if !now.Before(deadline) {
```
furykerry: now.After(deadline)

pkg/controller/sandboxset/sandboxset_controller.go
```diff
@@ -345,33 +375,54 @@ func compareScaleDownPriority(a, b *agentsv1alpha1.Sandbox) int {
...
+		return delta, nil
 	}
 
-	// apply maxUnavailable limit only for scale up
-	scaleMaxUnavailable := math.MaxInt
+	// Apply maxUnavailable only to physical scale-up execution. The percentage
+	// base is Spec.Replicas so an empty pool can still ramp up at the configured
+	// rate; the startup-budget condition in scaling_limited.go uses observed
+	// replicas separately for its own accounting.
+	scaleMaxUnavailable, err := resolveMaxUnavailable(sbs.Spec.ScaleStrategy.MaxUnavailable, sbs.Spec.Replicas)
```
furykerry: when calculating delta, we should also consider sandbox that is  timeout and failed

pkg/controller/sandboxset/scaling_limited.go
```diff
@@ -0,0 +1,147 @@
...
+	if nextDeadline.IsZero() {
+		return 0
+	}
+	return max(nextDeadline.Sub(now), 0)
+}
+
+func isStartupFailure(sandbox *agentsv1alpha1.Sandbox) bool {
+	condition := apiMeta.FindStatusCondition(sandbox.Status.Conditions, string(agentsv1alpha1.SandboxConditionReady))
+	return condition != nil && condition.Status == metav1.ConditionFalse &&
+		condition.Reason == agentsv1alpha1.SandboxReadyReasonStartContainerFailed
```
furykerry: 1. plz also check SandboxReadyReasonPodCreateFailed
2. currently StartContainerFailed is assigned to any container in Waiting, including transient states the kubelet retries out of on its own. we need enforce the logic to set StartContainerFailed based on pod status.conditions[] (PodScheduled), status.containerStatuses[].state.Waiting.Reason e.g. CreateContainerConfigError, CreateContainerError, RunContainerError, CrashLoopBackOff

pkg/controller/sandboxset/scaling_limited.go
ywExcellent: Done in 5a7fc68 — the pending-timeout check now reads `if now.After(deadline)`. Semantics are unchanged (they only differ at the sub-nanosecond boundary); it just reads more naturally.

pkg/controller/sandboxset/sandboxset_controller.go
ywExcellent: Good catch — aligned in 5a7fc68. Extracted `countStartupBlocked` (startup-failed + pending-timeout sandboxes) and both `calculateScalingLimited` and `calculateScaleDelta` now consume that same count, so the scale-up delta and the `ScalingLimited` condition use one definition of "unavailable". Healthy in-flight creations within the pending window are no longer subtracted from the scale-up budget, and failures/timeouts consistently are.

pkg/controller/sandboxset/scaling_limited.go
ywExcellent: Addressed in 416f6e2.

1. `SandboxReadyReasonPodCreateFailed` is now considered a startup failure alongside `StartContainerFailed`: `isStartupFailure` in `pkg/controller/sandboxset/scaling_limited.go` treats a `Ready=False` condition with either reason as occupying the startup budget, and tests were extended to cover the `PodCreateFailed` case.

2. `StartContainerFailed` is no longer assigned to any container merely sitting in `Waiting`. The new `classifyStartupFailure` helper in `pkg/controller/sandbox/core/pod_failure.go` only reports a definitive failure when a container's `state.Waiting.Reason` matches an explicit allow-list: `CreateContainerConfigError`, `CreateContainerError`, `RunContainerError`, `CrashLoopBackOff`, `InvalidImageName`, `ErrImageNeverPull`. Transient states kubelet retries out of on its own (e.g. `ContainerCreating`, `PodInitializing`, `ImagePullBackOff`, `ErrImagePull`) intentionally do not flip Ready, so the ResourcePending timeout remains the safety net for them.

We deliberately kept the list narrow — we only classify what's unambiguously definitive; anything uncertain still falls through to the timeout path. `common_control.go` also clears the failure reason once a subsequent poll no longer sees a definitive waiting state, so the Ready condition recovers on its own when kubelet succeeds on a retry.

Ref: `pkg/controller/sandbox/core/pod_failure.go`, `pkg/controller/sandbox/core/common_control.go`.

pkg/controller/sandbox/core/pod_failure.go
```diff
@@ -0,0 +1,111 @@
...
+// internal module of kube and is not exposed as a stable API. We mirror the
+// exact string values here so we do not take a dependency on k8s.io/kubernetes.
+//
+// Keep this list deliberately narrow. Other states are handled by kubelet
+// retries and the SandboxSet ResourcePending timeout path.
+const (
+	waitingReasonCreateContainerConfigError = "CreateContainerConfigError"
+	waitingReasonCreateContainerError       = "CreateContainerError"
+	waitingReasonRunContainerError          = "RunContainerError"
+	waitingReasonCrashLoopBackOff           = "CrashLoopBackOff"
```
furykerry: container can switch quickly from CrashLoopBackOff to running which may result in a storm of status changes, consider remove this reason

pkg/controller/sandboxset/scaling_limited.go
```diff
@@ -0,0 +1,157 @@
...
+// intstr resolution cannot fail here; on the unreachable error path we fall
+// back to a budget of one so the pool can still make forward progress.
+func resolveStartupBudget(maxUnavailable *intstrutil.IntOrString, observedReplicas int32) int {
+	executionBase := max(int(observedReplicas), 1)
+	if maxUnavailable == nil {
+		return executionBase
+	}
+	resolved, err := intstrutil.GetScaledValueFromIntOrPercent(maxUnavailable, executionBase, true)
+	if err != nil {
+		return 1
```
furykerry, line 132: plz do not swallow the change, since we have validated the value in webhook, we can simply log the error.

pkg/controller/sandboxset/scaling_limited.go
```diff
@@ -0,0 +1,157 @@
...
+// resolveMaxUnavailable resolves MaxUnavailable against the supplied base.
+// Callers pass Spec.Replicas when sizing physical scale-up steps. Admission
+// enforces the value format; an unreachable resolution error degrades to a
+// zero delta so scale-up simply pauses this pass.
+func resolveMaxUnavailable(maxUnavailable *intstrutil.IntOrString, base int32) int {
+	if maxUnavailable == nil {
+		return math.MaxInt
+	}
+	resolved, err := intstrutil.GetScaledValueFromIntOrPercent(maxUnavailable, max(int(base), 1), true)
+	if err != nil {
```
furykerry, line 147: plz do not swallow the change, since we have validated the value in webhook, we can simply log the error.

pkg/controller/sandbox/core/common_control.go
```diff
@@ -222,11 +222,16 @@ func defaultSyncStatusFromPod(pod *corev1.Pod, newStatus *agentsv1alpha1.Sandbox
...
-			cond.Message = cStatus.State.Waiting.Message
+	// Only classify explicit container startup failures. Other Waiting reasons
+	// are left to kubelet retries and the SandboxSet ResourcePending timeout.
+	if cond.Status == metav1.ConditionFalse {
+		if reason, message, failed := classifyStartupFailure(pod); failed {
+			cond.Reason = reason
+			cond.Message = message
+		} else if cond.Reason == agentsv1alpha1.SandboxReadyReasonStartContainerFailed ||
+			cond.Reason == agentsv1alpha1.SandboxReadyReasonPodCreateFailed {
+			cond.Reason = agentsv1alpha1.SandboxReadyReasonPodReady
```
furykerry: Issue: Two problems. (a) Inside cond.Status == ConditionFalse, setting Reason = PodReady produces Ready=False, Reason=PodReady — a condition that reads as "not ready because ready." (b) cond.Message = "" discards the kubelet message that explained the failure, right at the moment an operator most wants it.

Recommendation: keep SandboxReadyReasonPodCreateFailed/ SandboxReadyReasonStartContainerFailed, or change SandboxReadyReasonPodReady to 'SandboxReadyReasonPodNotReady'

api/v1alpha1/sandboxset_types.go
```diff
@@ -138,6 +145,14 @@ type SandboxSetUpdateStrategy struct {
 	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
 }
 
+// SandboxSetConditionType is a valid value for SandboxSet conditions.
```
furykerry, line 155: sandboxSetScaleStrategy.MaxUnavailable godoc no longer matches its behaviour

pkg/controller/sandbox/core/pod_failure.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		if c.Status == corev1.ConditionFalse && c.Reason == corev1.PodReasonUnschedulable {
+			return agentsv1alpha1.SandboxReadyReasonStartContainerFailed,
+				c.Message,
+				true
+		}
+		break
+	}
+
+	for i := range pod.Status.ContainerStatuses {
```
furykerry: classifyStartupFailure skips InitContainerStatuses. An init container failing with CreateContainerConfigError blocks the pod just as definitively as an app container, but is invisible to the budget.

pkg/controller/sandboxset/scaling_limited.go
```diff
@@ -0,0 +1,157 @@
...
+// the admission webhook, so intstr resolution errors are no longer expected
+// here and callers do not need to handle them.
+func (r *Reconciler) calculateScalingLimited(
+	ctx context.Context,
+	sbs *agentsv1alpha1.SandboxSet,
+	status *agentsv1alpha1.SandboxSetStatus,
+	groups GroupedSandboxes,
+	now time.Time,
+) time.Duration {
+	failed, timedOut, nextDeadline := countStartupBlocked(groups, r.sbxMaxPendingTimeout, now)
```
furykerry, line 52: countStartupBlocked runs twice per reconcile. It is called at scaling_limited.go:51 (inside calculateScalingLimited) and again at sandboxset_controller.go:167, over the same groups and the same now. Have calculateScalingLimited return the counts it already computed:

pkg/controller/sandbox/core/pod_failure.go
ywExcellent: Fixed in cc02620. Dropped `CrashLoopBackOff` from the definitive-failure allow-list in `pod_failure.go`; a container can flip between `CrashLoopBackOff` and `Running` quickly and treating it as definitive would trigger a storm of `ScalingLimited` transitions. It is now left to kubelet retries and the `ResourcePending` timeout path like other transient states, and the test case for it was updated to assert transient.

pkg/controller/sandboxset/scaling_limited.go
ywExcellent: Fixed in cc02620. `resolveStartupBudget` now takes `ctx` and logs via `logf.FromContext(ctx).Error(err, ...)` on the unreachable resolution-error branch before falling back to a budget of 1, so if a value ever slips past the admission webhook we leave a trace instead of swallowing it.

pkg/controller/sandboxset/scaling_limited.go
ywExcellent: Fixed in cc02620. Same treatment for `resolveMaxUnavailable`: it now takes `ctx` and logs the (unreachable) intstr resolution error via `logf.FromContext(ctx).Error` before degrading to a zero delta so this pass simply pauses scale-up.

pkg/controller/sandbox/core/common_control.go
ywExcellent: Fixed in cc02620. Introduced `SandboxReadyReasonPodNotReady = "PodNotReady"` in `api/v1alpha1/sandbox_types.go` and switched every Ready=False path in `common_control.go` (initial default, pod cond transition to False, and startup-failure recovery) to use it instead of reusing `PodReady`, which read as "not ready because ready". On the False side we also preserve the upstream `pCond.Message` so operators keep the kubelet-level diagnostic context. Existing `common_control_test.go` assertions were migrated accordingly.

api/v1alpha1/sandboxset_types.go
ywExcellent: Fixed in cc02620. Rewrote the `MaxUnavailable` godoc in `sandboxset_types.go` to match the new behavior: the controller no longer fails scale when the budget is exhausted; it throttles physical scale-up (delta is clamped) and surfaces `ScalingLimited=True` with reason `StartupBudgetExhausted`. Also spelled out what counts against the budget (definitive startup failures plus `sbxMaxPendingTimeout`-expired creations) and that healthy in-flight creations do not consume it. CRD schema regenerated via `make manifests`.

pkg/controller/sandbox/core/pod_failure.go
ywExcellent: Fixed in cc02620. `classifyStartupFailure` now checks `InitContainerStatuses` before `ContainerStatuses` via a shared `firstDefinitiveWaitingMessage` helper: an init container with `CreateContainerConfigError` / `InvalidImageName` / etc. blocks app containers just as definitively, and surfacing it eagerly keeps the reason and message pointing at the actual root cause. Added a table case that pairs an init `InvalidImageName` with an app `CreateContainerError` to lock in the ordering.

pkg/controller/sandboxset/scaling_limited.go
ywExcellent: Fixed in cc02620. `calculateScalingLimited` now returns `(requeueAfter time.Duration, blocked int)`; `Reconcile` consumes the returned `blocked` and passes it into `calculateScaleDelta` directly, so `countStartupBlocked` runs exactly once per reconcile over the same groups instead of twice.

HARSHRAJ2789:
Read `cc02620` against master at `72353a0` and ran the accounting it changes.

Dropping `CrashLoopBackOff` from the definitive-failure list is right for the flapping raised in the review above. What I could not make hold is the fallback the new comment names. `pod_failure.go:34` now groups `CrashLoopBackOff` with `ImagePullBackOff`, `ErrImagePull`, `ContainerCreating` and `PodInitializing` as handled by "kubelet retries and the SandboxSet ResourcePending timeout path". For the other four that is true. For `CrashLoopBackOff` it is not, and the difference is the pod phase.

`countStartupBlocked` reaches its timeout branch only when `GetSandboxState` returns reason `ResourcePending` (`scaling_limited.go:101`), and `utils.go:239` returns that reason only while `Status.Phase == SandboxPending`. During startup the other four occur before any container has run, so the pod is still `PodPending`, which `k8s.io/api` defines as covering "time spent pulling images onto the host". `CrashLoopBackOff` requires the container to have started and be restarting, and that is `PodRunning`: "At least one container is still running or is in the process of being restarted". `common_control.go:138` sets `newStatus.Phase = SandboxRunning` as soon as the pod reports `PodRunning`, and no path sets it back to `SandboxPending`.

A crash-looping pooled sandbox therefore groups as `Creating` with reason `ResourceControlledBySbsButNotReady` (`utils.go:257`). `isStartupFailure` is false, because after this commit the reason is `PodNotReady`. The timeout branch skips it, because the reason is not `ResourcePending`. Running that against your branch:

| pod state | GetSandboxState reason | blocked |
|---|---|---|
| CrashLoopBackOff, after `cc02620` | ResourceControlledBySbsButNotReady | 0 |
| CrashLoopBackOff, before `cc02620` | ResourceControlledBySbsButNotReady | 1 |
| ImagePullBackOff, still Pending | ResourcePending | 1 |

A pool whose template has a bad entrypoint keeps scaling. Ten crash-looping sandboxes measure `blocked=0`, so at `Replicas: 10` the condition evaluates `0 >= 10` and `ScalingLimited` stays False. The same ten measured `blocked=10` before this commit, which made it True. The scale-up delta moves with it wherever `MaxUnavailable` is set: taking a pool from 10 to 20 with `MaxUnavailable: 3`, delta is 3 at `blocked=0` and 0 at `blocked=10`. Where `MaxUnavailable` is unset this count never throttled the delta anyway.

I checked the paths that might catch it anyway. `sbxMaxPendingTimeout` is applied as a deadline only inside `countStartupBlocked`; its other uses are `resolveScaleUpCooldown` in the poolautoscaler and a range warning at controller registration (`sandbox_controller.go:138`). `updateStatusIfPodCompleted` (`sandbox_controller.go:818`) fires only on `PodSucceeded` or `PodFailed`, and `Spec.Template` is a user-supplied `PodTemplateSpec` whose `restartPolicy` defaults to `Always`, so a crash-looping pod never reaches `PodFailed`. `RestartCount` is read only by `metrics.go:754`, for metrics.

The build is fine. 29/29 checks are green, and `go build ./...` and the changed-package tests pass on `cc02620` here, so what has the gap is the accounting. `pod_failure_test.go:51` pins the new classification on purpose. What no test covers is what the startup budget does afterwards.

One direction, if it is useful. Keep `CrashLoopBackOff` off the definitive list, and give `countStartupBlocked` a second entry that does not depend on `ResourcePending`: count anything in `groups.Creating` that is still unready past `sbxMaxPendingTimeout` since creation. That keys off age, so a container flipping between backoff and running does not move the count.

pkg/controller/sandboxset/scaling_limited.go
```diff
@@ -0,0 +1,168 @@
...
+// simply pauses this pass.
+func resolveMaxUnavailable(ctx context.Context, maxUnavailable *intstrutil.IntOrString, base int32) int {
+	if maxUnavailable == nil {
+		return math.MaxInt
+	}
+	resolved, err := intstrutil.GetScaledValueFromIntOrPercent(maxUnavailable, max(int(base), 1), true)
+	if err != nil {
+		logf.FromContext(ctx).Error(err, "unexpected error resolving maxUnavailable for scale delta; admission webhook should have rejected this value",
+			"maxUnavailable", maxUnavailable.String(), "base", base)
+		return 0
```
furykerry: return zero will disable scaling permanently, consider return more reasonable default

pkg/controller/sandboxset/sandboxset_controller_test.go
```diff
@@ -1083,7 +1083,7 @@ func TestCalculateScaleDelta(t *testing.T) {
 		name              string
```
furykerry: **`gofmt` violation — this will fail CI.**

The struct fields here (and the ~20 table cases below) are still aligned to an 18-column stop, left over from before a field was removed from each case. `gofmt` wants them realigned to the longest remaining name:

```
$ gofmt -l pkg/controller/sandboxset/
pkg/controller/sandboxset/sandboxset_controller_test.go

$ gofmt -d pkg/controller/sandboxset/sandboxset_controller_test.go | wc -l
336        # single hunk: @@ -1080,169 +1080,169 @@
```

Fix: `gofmt -w pkg/controller/sandboxset/sandboxset_controller_test.go`

This was also raised in my previous two rounds; it's purely mechanical, so worth wiring `gofmt -l` into your pre-push loop to stop it recurring.

pkg/controller/sandbox/core/pod_failure.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+		if !isDefinitiveWaitingReason(cs.State.Waiting.Reason) {
+			continue
+		}
+		return cs.State.Waiting.Message, true
+	}
+	return "", false
+}
+
+func isDefinitiveWaitingReason(reason string) bool {
```
furykerry, line 123: **Removing `CrashLoopBackOff` creates a permanent blind spot — the `ResourcePending` timeout path does not cover it.**

A crash-looping pod has `pod.Status.Phase == corev1.PodRunning`, so the Sandbox phase is set on pod phase alone (`common_control.go:138-140`), and `GetSandboxState` then returns `(Creating, "ResourceControlledBySbsButNotReady")` — **not** `ResourcePending`. So `countStartupBlocked` hits the `continue` at `scaling_limited.go:100-102`, no deadline is ever assigned, and nothing reaps it (`git grep ResourceControlledBySbsButNotReady` has exactly one non-test hit: the `return` itself; `--max-pending-timeout` is a read-only input).

Net effect: a crash-looping sandbox consumes **zero** startup budget, forever. `ScalingLimited` stays `False` and the autoscaler keeps creating more — the opposite of this PR's goal. Before this commit these were counted via `StartContainerFailed`.

But re-adding it to the allow-list isn't the fix either. My earlier "it flaps" note was accurate but I described the cause too loosely: `State.Waiting.Reason` is a **duty-cycle-dependent instantaneous signal**, not merely a value that toggles. One crash cycle is `Running` (~2s) -> `Terminated` (milliseconds, usually never even reaches the apiserver) -> `Waiting{CrashLoopBackOff}` (10s, 20s, 40s ... capped at 300s). So the observable duty cycle is ~83% on the first restart, ~98% at the 300s cap, but under 2% for a container that crashes every 11 minutes — kubelet resets the backoff after 10 minutes of clean running, while `RestartCount` does not reset. And our chain samples this **twice** (pod status -> `Ready.reason` -> `isStartupFailure`), which amplifies it.

`RestartCount` is monotonic and sampling-independent, so it is the only stable signal here. Note it must **not** be nested under `State.Waiting != nil`, or it still flaps during the `Running` window:

```go
func isDefinitiveContainerFailure(cs *corev1.ContainerStatus) bool {
	// RestartCount is monotonic and immune to the duty-cycle sampling problem
	// that makes State.Waiting.Reason == CrashLoopBackOff flap.
	if cs.RestartCount >= definitiveRestartThreshold && !cs.Ready {
		return true
	}
	if cs.State.Waiting == nil {
		return false
	}
	return isDefinitiveWaitingReason(cs.State.Waiting.Reason)
}
```

Also note `exitCode` cannot be used as the discriminator: a container that exits `0` under `restartPolicy: Always` still crash-loops (`lastState.terminated.reason: Completed`) and never becomes Ready. `RestartCount + !Ready` covers that case too.

Whichever way you go, the comment at lines 33-40 claiming these states are "handled by kubelet retries and the SandboxSet `ResourcePending` timeout path" needs correcting — that assertion is what hides the gap.

More generally, this is not specific to `CrashLoopBackOff`: **any** sandbox that reaches `Running` but never becomes `Ready` now escapes the budget entirely, whereas the old `replicas - availableReplicas` formula counted it. Worth closing with a readiness deadline (not just a pending deadline).

pkg/controller/sandboxset/scaling_limited.go
ywExcellent: Fixed in 4a66121. When `maxUnavailable` is nil or intstr parsing fails we now fall back to the default 25% budget via `resolveDefaultScaleMaxUnavailable`, so scaling is not permanently disabled.

pkg/controller/sandboxset/sandboxset_controller_test.go
ywExcellent: Fixed in 4a66121 with `gofmt -w`. `gofmt -l pkg/controller/sandboxset/` is now clean.

pkg/controller/sandbox/core/pod_failure.go
ywExcellent: Thanks for the detailed analysis — implemented as suggested in `0d0587e`.

Concrete changes:

1. `pkg/controller/sandbox/core/pod_failure.go`: dropped the "CrashLoopBackOff-via-Waiting" idea entirely and switched to `RestartCount + !Ready` as the stable, sampling-independent signal. The check is intentionally **not** nested under `State.Waiting != nil` so it still fires during the brief `Running` window between backoffs:

   ```go
   if cs.RestartCount >= definitiveRestartThreshold && !cs.Ready {
       return fmt.Sprintf("container %q has restarted %d times and is not ready", cs.Name, cs.RestartCount), true
   }
   ```

   Threshold set to `5`. Also fixed the misleading comment block that claimed these states were "handled by kubelet retries and the ResourcePending timeout path".

2. Per-lease baseline caveat recorded in code so future Sandbox reuse doesn't silently miscount:

   ```go
   // RestartCount currently spans the Pod lifetime. If Sandbox reuse is
   // introduced, compare against a per-lease restart baseline instead so
   // restarts from an earlier lease do not classify the next one as failed.
   ```

3. Readiness-deadline / "escape from budget" concern: addressed in `pkg/controller/sandboxset/sandboxset_controller.go` by restoring `unavailable = replicas - available` as the physical scale-up budget input. Anything that reaches `Running` but never `Ready` (crash loop, wedged readiness probe, etc.) is now counted against the budget again, matching the pre-refactor semantics. `ScalingLimited` continues to use the narrower definitive-failure + pending-timeout signal, so the two concerns stay decoupled.

`exitCode==0 under restartPolicy: Always` case is covered by (1) since `RestartCount + !Ready` doesn't depend on the exit code.

pkg/controller/sandbox/core/common_control.go
```diff
@@ -219,14 +219,31 @@ func defaultSyncStatusFromPod(pod *corev1.Pod, newStatus *agentsv1alpha1.Sandbox
 	if pCond != nil && string(pCond.Status) != string(cond.Status) {
 		cond.Status = metav1.ConditionStatus(pCond.Status)
 		cond.LastTransitionTime = pCond.LastTransitionTime
-		cond.Reason = agentsv1alpha1.SandboxReadyReasonPodReady
-		cond.Message = ""
+		if cond.Status == metav1.ConditionTrue {
+			cond.Reason = agentsv1alpha1.SandboxReadyReasonPodReady
+			cond.Message = ""
+		} else {
+			cond.Reason = agentsv1alpha1.SandboxReadyReasonPodReady
```
furykerry: these two if clause is identical, consider merge them

pkg/controller/sandboxset/scaling_limited.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	return max(resolved, 1)
+}
+
+// resolveMaxUnavailable resolves MaxUnavailable against the supplied base.
+// Callers pass Spec.Replicas when sizing physical scale-up steps. An absent or
+// invalid value falls back to the default 25% limit.
+func resolveMaxUnavailable(ctx context.Context, maxUnavailable *intstrutil.IntOrString, base int32) int {
+	if maxUnavailable == nil {
+		return resolveDefaultScaleMaxUnavailable(base)
```
furykerry: consider Reuse intstrutil for the default instead of hand-rolled rounding func

pkg/controller/sandboxset/scaling_limited.go
```diff
@@ -0,0 +1,167 @@
...
+		if isStartupFailure(sandbox) {
+			failed++
+			continue
+		}
+		state, reason := utils.GetSandboxState(sandbox)
+		if state != agentsv1alpha1.SandboxStateCreating || reason != agentsv1alpha1.SandboxStateReasonResourcePending {
+			continue
+		}
+		deadline := sandbox.CreationTimestamp.Add(sbxMaxPendingTimeout)
+		if now.After(deadline) {
```
furykerry: my mistake, plz restore to !now.Before(deadline), if now == deadline, now.After(deadline) will return false

pkg/controller/poolautoscaler/poolautoscaler_controller.go
```diff
@@ -357,6 +395,49 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	}
+	report := scaleBlockedReport{
+		poolAutoscalerUID: pa.UID,
+		sandboxSetUID:     sbs.UID,
+		generation:        sbs.Generation,
+		conditionReason:   condition.Reason,
+		conditionChanged:  condition.LastTransitionTime,
+	}
+	key := client.ObjectKeyFromObject(pa).String()
+	if previous, ok := r.scaleBlockedReported.Load(key); ok && previous == report {
```
furykerry, line 409: metav1.Time ==` comparison**  is fragile; since  conditionChanged is not logged anywhere, consider just remove conditionChanged field

pkg/controller/sandbox/core/common_control.go
ywExcellent: Done in `d820fbf`. The two branches are no longer identical:

```go
if pCond.Status != cond.Status {
    cond.LastTransitionTime = metav1.Now()
    if pCond.Status == metav1.ConditionTrue {
        cond.Reason = agentsv1alpha1.SandboxReadyReasonPodReady
        cond.Message = ""
    } else {
        // keep the classified failure Reason (e.g. StartContainerFailed);
        // only sync the human-readable message from the Pod side.
        cond.Message = pCond.Message
    }
    ...
}
```

Reason is now only overwritten when a new failure is classified upstream, or when readiness flips to True. It is no longer speculatively reset back to `PodReady` while the Pod is still not ready.

pkg/controller/sandboxset/scaling_limited.go
ywExcellent: Addressed in `d820fbf` by removing the hand-rolled helper entirely instead of porting it to `intstrutil`.

`resolveMaxUnavailable` now falls back to `base` when `MaxUnavailable` is `nil` or when `intstrutil.GetScaledValueFromIntOrPercent` returns an error, which is equivalent to the previous "100% default" without introducing another rounding path:

```go
fallback := max(int(base), 1)
if maxUnavailable == nil {
    return fallback
}
resolved, err := intstrutil.GetScaledValueFromIntOrPercent(maxUnavailable, fallback, true)
if err != nil {
    ...
    return fallback
}
return max(resolved, 0)
```

The webhook already enforces the accepted range on the field, so the controller no longer needs its own default rounding.

pkg/controller/sandboxset/scaling_limited.go
ywExcellent: Restored in `d820fbf`:

```go
deadline := sandbox.CreationTimestamp.Add(sbxMaxPendingTimeout)
if !now.Before(deadline) {
    timedOut++
    continue
}
```

Boundary tick (`now == deadline`) now correctly counts as timed out.

pkg/controller/poolautoscaler/poolautoscaler_controller.go
ywExcellent: Removed in `d820fbf`. `scaleBlockedReport` is now just:

```go
type scaleBlockedReport struct {
    paUID              types.UID
    sbsUID             types.UID
    observedGeneration int64
    conditionReason    string
}
```

Dedupe is driven by `conditionReason` (plus the identity tuple), so a new blocked episode surfaces as soon as the SandboxSet controller flips Reason. `metav1.Time` equality is no longer part of the key.

api/v1alpha1/sandboxset_types.go
```diff
@@ -113,10 +120,17 @@ type SandboxSetSpec struct {
...
 type SandboxSetScaleStrategy struct {
-	// The maximum number of sandboxes that can be unavailable for scaled sandboxes.
-	// This field can control the changes rate of replicas for SandboxSet so as to minimize the impact for users' service.
-	// The scale will fail if the number of unavailable sandboxes were greater than this MaxUnavailable at scaling up.
-	// MaxUnavailable works only when scaling up.
+	// MaxUnavailable caps the number of unavailable sandboxes during scale-up.
+	// It can be an absolute number (ex: 5) or a percentage of desired pods
+	// (ex: 10%); percentages are rounded up. Every non-Available sandbox counts
+	// against the physical scale-up budget. If unset or invalid, the controller
+	// uses 25%.
```
furykerry: API documentation contradicts code， plz change the comment and regenerated config/crd/bases/agents.kruise.io_sandboxsets.yaml

ywExcellent: Fixed in `ac27b42`. Updated the godoc to match actual behavior:

```go
// If unset or invalid, the controller uses the base (equivalent to 100%, i.e. no cap).
```

Regenerated CRD YAML via `make manifests`.

PR #899, opened 2026-08-28
feat(api): add Probes and AutoPausePolicy for sandbox auto-pause

zmberg:

Add the API surface and design proposal for sandbox auto-pause and resume. This PR is API + proposal only; the controller implementation will follow in a separate PR so the contract can be reviewed and merged first.

API changes (`api/v1alpha1/sandbox_types.go`):

- `SandboxSpec.Probes`: a list of named periodic probes that run while the sandbox is Running. Probe execution is delegated to the agent-runtime sidecar through the PodProbeMarker Serverless protocol (`kruise.io/podprobe` annotation); results land in `Pod.Status.Conditions` and are mirrored to `SandboxStatus.Conditions` as `agents.kruise.io/<probe-name>`.
- `SandboxSpec.AutoPausePolicy`: pause/resume decision rules driven by probe Conditions — `whenProbedIdleState` (pause after the probe message matches `messageRegex` for `thresholdDuration`) and `whenProbedScheduleTime` (resume `leadTime` before the timestamp reported by the probe).
- `SandboxStatus.Schedules`: upcoming pause/resume timing (`nextPauseTime` / `nextResumeTime`) keyed by `reason`, for observability.
- `ProbeValid` condition plus probe result / decision / schedule reason constants.
- `AnnotationPodProbe` for the `kruise.io/podprobe` protocol.

Proposal: `docs/proposals/20260626-sandbox-auto-pause.md`.

Generated outputs are regenerated from the API change: `zz_generated.deepcopy.go` and `config/crd/bases/agents.kruise.io_sandboxes.yaml`.

NONE

1. Run `make generate manifests` and confirm there is no diff — the committed deepcopy functions and Sandbox CRD match controller-gen output.
2. Apply the regenerated CRD and confirm a Sandbox with `spec.probes` and `spec.autoPausePolicy` is accepted, and that the `leadTime` default (`5m`) and the `timeFormat: unix` enum validation take effect.
3. All new fields are optional, so existing Sandbox objects remain valid — verify an existing Sandbox still applies unchanged.

- Field naming and semantics are the main thing to review: `autoPausePolicy.pause.whenProbedIdleState` / `resume.whenProbedScheduleTime`, `thresholdDuration` measured from the Condition's `lastTransitionTime`, `leadTime` defaulting to `5m`, and `timeFormat` restricted to `unix`.
- `SandboxStatus.Schedules` uses `patchMergeKey=reason` with `patchStrategy=merge`; the reason values are the `probedIdle` / `probedSchedule` constants.
- Probes are intentionally generic: their meaning (activity detection vs. scheduled-task detection) comes from `AutoPausePolicy`, not from the probe definition itself.
- Worth checking that the proposal and the API definitions stay consistent, since the implementation lands separately.


HARSHRAJ2789:
Read this against master at `e9e5e08` and ran the threshold contract through the helper that will
serve it. `make generate manifests` on this branch produces no diff, so verification step 1 holds.

Three things on the semantics you asked to have reviewed.

**1. `ThresholdDuration` measured from `lastTransitionTime` cannot work, and this PR freezes it into
the CRD.**

The doc on `ProbedIdleStateRule` says the threshold is "measured from the Condition's
lastTransitionTime", and the proposal repeats it at line 505 as `elapsed = now -
Condition.lastTransitionTime`.

Idleness is carried in the Condition **message**, not its status. The proposal is explicit about
that: probes "should always exit 0 and output semantic information to stdout", so `status` stays
`True` while the agent is busy and stays `True` once it goes idle. `utils.SetSandboxCondition`
(`pkg/utils/utils.go:58`) only moves `LastTransitionTime` when the status changes:

```go
if currentCond.Status != condition.Status || currentCond.LastTransitionTime.IsZero() {
    currentCond.LastTransitionTime = condition.LastTransitionTime
}
```

Driving that helper with a probe that goes green two hours ago and then flips its message to
`inactive`, with a 15 minute threshold:

```
after "active"   : status=True message="active"   lastTransitionTime=13:51:52
after "inactive" : status=True message="inactive" lastTransitionTime=13:51:52   (unchanged)
elapsed = now - lastTransitionTime = 2h0m0s
pause time = lastTransitionTime + 15m = 14:06:52   ->  already 1h45m in the past
```

So the sandbox pauses on the first idle report and `ThresholdDuration` is never honoured. This
applies to every sandbox whose probe was already succeeding before it went idle, which is the
normal case. The proposal documents the mechanism that breaks it, at line 569:
`lastTransitionTime: "2026-07-01T10:00:00Z"  # updated when status changes`.

The threshold needs an anchor stamped when the idle period begins.
`Status.Schedules` already carries `nextPauseTime` per reason and would hold it with no new API
surface. I posted a tested version of that on #612 on 27 August if it is useful.

Worth settling here, since this doc becomes the CRD text users read and the implementation PR
will be written to follow it.

**2. The proposal's example YAML does not validate against the CRD in this same commit.**

`config/crd/bases/agents.kruise.io_sandboxes.yaml:74` accepts `whenProbedIdleState`. The proposal's
example at line 140 writes `whenIdleProbeFires`, and its Go snippet at line 339 declares
`WhenIdleProbeFires *IdleProbeFireRule`. Across the proposal that is 24 occurrences of the old name
and none of the new; `sandbox_types.go` has 7 of the new and none of the old.

Anyone copying the example gets a field the API server drops.

**3. The nil `ThresholdDuration` contract contradicts the implementation it is written for.**

The field doc says "If nil, the sandbox is paused immediately when the message matches", and it is
`+optional`. `evaluatePauseSchedule` in #612 (`pkg/controller/sandbox/auto_pause.go:150`) treats nil
as a missing required field, emits an `InvalidPauseRule` warning and returns without pausing. One of
the two has to move, and the CRD is the side users read first.

PR #900, opened 2026-08-28
feat(api): OnIngressTraffic resume rule for wake-on-traffic

furykerry:

This PR promotes the wake-on-traffic configuration from annotations to a
first-class Sandbox spec field, and wires it through the control plane. The
design is described in `docs/proposals/20260627-wake-on-traffic-via-spec-patch.md`
(updated by this PR).

**API** (`api/v1alpha1`):

- New `Spec.AutoPausePolicy.Resume.OnIngressTraffic *IngressTrafficRule`.
  A non-nil rule means wake-on-ingress-traffic is enabled; there is no
  separate enable flag. The optional `PauseTimeout`, when positive,
  re-arms auto-pause after a traffic wake. It is never defaulted inside
  the setter — it is provided by the caller; when absent, the woken
  sandbox keeps running until it is paused or deleted again.
- Read helpers `WakeOnIngressTrafficEnabled` / `WakeOnIngressTrafficPauseTimeout` (in `pkg/utils`)
  give gateway, manager, and route projection a single resolution path.

**Control-plane wiring**:

- Write path: the E2B create handler writes the rule through the new neutral
  `infra.Sandbox.SetWakeOnIngressTraffic(enabled, pauseTimeout)` setter
  (implemented in `infra/sandboxcr`); it passes the request timeout as
  `PauseTimeout` for auto-pause sandboxes and leaves it unset for
  never-timeout/shutdown-only ones, whose wakes must not re-arm auto-pause.
- Read path: `route.WakeOnTraffic` projection and the gateway wake path read
  the spec via the helpers. On wake, a positive rule `PauseTimeout` re-arms
  auto-pause (with the resume timeout floor); without one the wake clears
  the stale `PauseTime` and extends `ShutdownTime` to the retention horizon
  so the controller does not immediately re-pause or delete the sandbox.
  The gateway `wake-timeout-seconds` config only bounds the wake operation
  itself.
- Recycle hygiene: `resetMetadataForPool` restores `Spec.Probes` and
  `Spec.AutoPausePolicy` verbatim from the SandboxSet so a recycled pool
  sandbox is byte-identical to a freshly-created one. Any per-tenant wake
  rule written by the E2B create path is overwritten; probe-driven resume
  rules declared by the SandboxSet template or operator survive because
  they are part of the SandboxSet policy.
- The wake annotation constants and their `AnnotationsClearedOnRecycle`
  entries are retired.

NONE

1. Unit tests (all pass):
   - `go test ./pkg/utils/` — read helper behavior (nil-safe, timeout validation)
   - `go test ./pkg/servers/e2b/...` — create modifier writes the rule for `autoResume`
   - `go test ./pkg/sandboxroute/...` — route projection from the spec rule
   - `go test ./pkg/sandbox-gateway/...` — wake enable check, re-arm vs no-re-arm, filter fallback
   - `go test ./pkg/controller/sandbox/core/...` — recycle restores `Probes` and `AutoPausePolicy` from the SandboxSet
   - `go test ./pkg/sandbox-manager/infra/sandboxcr/...` — setter enable/clear/prune
2. Create an auto-pause sandbox with `autoResume=true` via the E2B API and
   verify `spec.autoPausePolicy.resume.onIngressTraffic` is set with
   `pauseTimeout` equal to the request timeout; pause it and verify inbound
   traffic wakes it and re-arms auto-pause after that timeout. For a
   never-timeout sandbox, verify the rule carries no `pauseTimeout` and a
   wake does not re-arm auto-pause.

- `OnIngressTraffic` must NOT count toward `hasActiveAutoPausePolicy`:
  otherwise E2B `autoResume` sandboxes would silently lose timeout-based
  auto-pause once the `checkTimers` guard lands. The proposal documents
  this invariant.
- Wake without a rule `PauseTimeout` must clear the stale `PauseTime`
  (and extend `ShutdownTime`), otherwise `handlePauseTimeout` would
  re-pause the sandbox immediately after resume.
- `HashSandbox` hashes only `spec.template`, so adding the rule causes no
  SandboxSet revision churn.
- The E2B `autoResume` flag itself stays a delivery-scoped parameter; only
  its persistent representation moves to the spec.
- Layering: the API layer never touches CR spec directly — it goes through
  the new `infra.Sandbox` setter; the gateway/controller read through the
  shared `pkg/utils` helpers.


HARSHRAJ2789:
Read this against master at `72353a0`. The API shape reads well: nil meaning "not configured"
matches the sibling probed rules, and both helpers are nil-safe and guard a non-positive
`PauseTimeout`. One thing about retiring the annotations, and one merge note.

**Sandboxes that already carry the wake annotations lose wake-on-traffic on upgrade.**

The two constants are not vestigial on master today. `pkg/servers/e2b/create.go:456` writes
`wake-on-traffic` for every `autoResume` sandbox, `pkg/sandbox-gateway/wake/wake.go:98` and `:132`
read it and the timeout, and `pkg/sandboxroute/route.go:117` reads it for the route projection.

This PR removes the constants and all four sites. `basicSandboxCreateModifier` calls
`SetWakeOnIngressTraffic` on both branches, so a pooled Sandbox picks the rule up on its next claim.
What that leaves uncovered is the sandboxes already claimed and running at rollout. Nothing re-runs
the modifier until they are recycled and claimed again, the proposal says "No webhook is added", and
no conversion sets the rule in the meantime, so they lose wake-on-traffic for the remainder of that
delivery with no signal on the object.

The stale annotation also loses both of the paths that used to remove it. It comes out of
`AnnotationsClearedOnRecycle` (`api/v1alpha1/sandboxset_types.go:50` on master), and the modifier's
opening `delete(annotations, AnnotationWakeOnTraffic)` pair goes with it, which was there so a
previous delivery's value could not leak into the next one. Nothing takes the key off the object
again. That sits oddly beside the goal stated in the new comment here, that empty parents are pruned
"to keep the pooled spec byte-identical to a fresh one".

A read-time fallback in `WakeOnIngressTrafficEnabled`, or leaving the two keys on the recycle clear
list until the annotations are gone from live objects, would cover both.

**Merge note.** The recycle-hygiene part touches `resetMetadataForPool`, and #893 changes the same
function. `recycle.go` auto-merges, since this PR writes in Part 1 and #893 writes in the Part 2
manifest loop. The single conflict is the `TestResetForPool` field list in `recycle_test.go`: this
PR adds `expectAutoPausePolicy`, #893 adds `expectTemplateLabels` and `expectTemplateAnnotations`.
Keeping all three resolves it, and each PR is clean against master on its own.

Worth noting the two are the same idea applied to different fields. The comment here keeps
probe-driven rules because the SandboxSet template declares them, not a claim;
#893 deletes only the keys the claim recorded in `updated-metadata-in-claim`, so a template label
placed by the SandboxSet survives the same way.

api/v1alpha1/sandbox_types.go
```diff
@@ -270,12 +275,20 @@ type PausePolicy struct {
...
 	// +optional
 	WhenProbedScheduleTime *ProbedScheduleTimeRule `json:"whenProbedScheduleTime,omitempty"`
+
+	// OnIngressTraffic resumes the sandbox when the sandbox-gateway receives
+	// inbound traffic addressed to it while it is paused. Unlike the probed
+	// rules this one is event-driven: it needs no probe, it produces no
+	// Status.Schedules entry, and it is executed by the sandbox-gateway rather
+	// than by the sandbox controller.
+	// +optional
+	OnIngressTraffic *IngressTrafficRule `json:"onIngressTraffic,omitempty"`
```
zmberg, line 291: **[Issue]** `OnIngressTraffic` has not been included in the "at least one rule" check in `ValidateAutoPausePolicy` on master; it currently only counts `pause.whenProbedIdleState` and `resume.whenProbedScheduleTime`. As a result, a valid ingress-only policy is unconditionally rejected by the SandboxSet/SandboxTemplate validating webhooks, while a directly created Sandbox is incorrectly marked `ProbeValid=False` when `AutoPauseControllerGate` is enabled. Since this API explicitly defines `OnIngressTraffic` as probe-independent, please include `policy.Resume.OnIngressTraffic != nil` in the `hasAnyRule` check and add validation/webhook coverage for an ingress-only policy with no probes.

pkg/sandbox-gateway/wake/wake.go
```diff
@@ -104,84 +113,70 @@ func (w *Waker) HasWakeAnnotation(ctx context.Context, namespace, name string) b
...
 	// Resume refreshes from API reader, so the sandbox object is re-fetched
 	// with the latest state before patching.
 	sandbox := sandboxcr.AsSandbox(&sbx, w.cache)
 
 	// Determine the sandbox timeout mode, mirroring ParseTimeout in the
-	// E2B connect path. This ensures we only set PauseTime for auto-pause
+	// E2B connect path. This ensures we only touch deadlines of auto-pause
 	// sandboxes and never convert never-timeout or shutdown-only sandboxes
 	// into auto-pause mode.
 	autoPause := sbx.Spec.PauseTime != nil
```
zmberg, line 138: **[Issue]** The timeout mode and `PauseTimeout` are derived here from an informer-cache snapshot, but `Resume` subsequently refreshes the Sandbox through `APIReader` and applies these precomputed options to the latest object in `retryPatch`. If the cache lags behind a recent `PauseTime`, `ShutdownTime`, or `OnIngressTraffic` update, conflict retries refresh only the patched object and never recompute `opts`: an object that has most recently become shutdown-only/never-timeout may have a `PauseTime` reintroduced by stale options, while an object that has most recently become auto-pause may retain an expired `PauseTime` when stale `opts` is nil and immediately pause again after resuming. Please derive the timeout mode, rule timeout, and `ResumeOptions` from the same snapshot inside `retryPatch`, and add a test where the informer and `APIReader` return different object versions.

PR #901, opened 2026-08-29
fix(network): format IPv6 upstream addresses

cyrilcsr:

  This PR fixes ambiguous IPv6 host-port formatting in the sandbox traffic data
  plane.

  It uses `net.JoinHostPort` in the following address-construction paths:

  - `pkg/proxy/ext_proc.go`: Envoy `x-envoy-original-dst-host` values.
  - `pkg/proxy/utils.go`: sandbox-manager peer refresh HTTP URLs.
  - `pkg/sandbox-gateway/filter/filter.go`: Envoy `envoy.lb.original_dst`
    metadata.
  - `pkg/utils/runtime/runtime.go`: agent-runtime fallback URLs derived from the
    Sandbox Pod IP.
  - `pkg/utils/proxyutils/default.go`: direct HTTP requests to a Sandbox Pod IP.

  For example, an IPv6 destination is now formatted as `[2001:db8::1]:49999`
  instead of `2001:db8::1:49999`.

  IPv4 behavior remains unchanged. This PR does not change route synchronization,
  retry, concurrency, TLS authority, or memberlist behavior.

  NONE

  The full unit-test suite passes with Go 1.25.9:

  ```text
  GOTOOLCHAIN=go1.25.9 make test

  This runs all pkg/... tests with race detection and coverage.

  Focused tests also pass:

  go test ./pkg/proxy ./pkg/sandbox-gateway/filter \
    ./pkg/utils/runtime ./pkg/utils/proxyutils -count=1

  The tests cover:

  - IPv6 ext-proc original destinations.
  - IPv6 peer refresh HTTP requests.
  - IPv6 sandbox-gateway destinations from sandbox and host headers.
  - IPv6 agent-runtime fallback URLs.
  - Direct requests to an IPv6 Sandbox Pod using a real IPv6 loopback server.
  - Existing IPv4 behavior.

  This change is limited to host-port formatting at the five affected
  address-construction paths listed above. Memberlist IPv6 handling remains
  intentionally excluded.

### DIFF
--- a/pkg/proxy/ext_proc.go
+++ b/pkg/proxy/ext_proc.go
@@ -22,7 +22,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
+	"strconv"
 
 	configPb "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	extProcPb "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
@@ -136,7 +138,7 @@ func (s *Server) handleRequestHeaders(requestHeaders *extProcPb.ProcessingReques
 	}
 	// An adapter can set "x-envoy-original-dst-host" header to force route the request to a specific destination
 	if _, ok := extraHeaders[OrigDstHeader]; !ok {
-		extraHeaders[OrigDstHeader] = fmt.Sprintf("%s:%d", route.IP, sandboxPort)
+		extraHeaders[OrigDstHeader] = net.JoinHostPort(route.IP, strconv.Itoa(sandboxPort))
 	}
 
 	return s.logAndCreateDstResponse(requestHeaders.RequestHeaders, extraHeaders, log)
--- a/pkg/proxy/ext_proc_test.go
+++ b/pkg/proxy/ext_proc_test.go
@@ -174,6 +174,56 @@ func TestServer_Process(t *testing.T) {
 			},
 		},
 		{
+			name: "IPv6 sandbox upstream uses bracketed host port",
+			setupRoutes: []sandboxroute.Route{
+				{ID: "sandbox-ipv6", IP: "2001:db8::1", Owner: "user1"},
+			},
+			adapter: &testRequestAdapter{
+				isSandboxRequest: true,
+				mapResult: mapResult{
+					sandboxID:   "sandbox-ipv6",
+					sandboxPort: 49999,
+				},
+				entry: "127.0.0.1:8080",
+			},
+			requests: []*extProcPb.ProcessingRequest{
+				{
+					Request: &extProcPb.ProcessingRequest_RequestHeaders{
+						RequestHeaders: &extProcPb.HttpHeaders{
+							Headers: &corev3.HeaderMap{
+								Headers: []*corev3.HeaderValue{
+									{Key: ":scheme", RawValue: []byte("http")},
+									{Key: ":authority", RawValue: []byte("localhost:9002")},
+									{Key: ":path", RawValue: []byte("/sandbox")},
+								},
+							},
+						},
+					},
+				},
+			},
+			expectError: false,
+			expectResp: []*extProcPb.ProcessingResponse{
+				{
+					Response: &extProcPb.ProcessingResponse_RequestHeaders{
+						RequestHeaders: &extProcPb.HeadersResponse{
+							Response: &extProcPb.CommonResponse{
+								HeaderMutation: &extProcPb.HeaderMutation{
+									SetHeaders: []*corev3.HeaderValueOption{
+										{
+											Header: &corev3.HeaderValue{
+												Key:      "x-envoy-original-dst-host",
+												RawValue: []byte("[2001:db8::1]:49999"),
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		{
 			name:        "non-sandbox",
 			setupRoutes: []sandboxroute.Route{},
 			adapter: &testRequestAdapter{
--- a/pkg/proxy/routes_test.go
+++ b/pkg/proxy/routes_test.go
@@ -344,6 +344,25 @@ func TestSyncRouteWithPeers_TwoNodes_Success(t *testing.T) {
 	assert.Equal(t, route.ID, peer2.getReceived()[0].ID)
 }
 
+func TestSyncRouteWithPeers_IPv6Peer(t *testing.T) {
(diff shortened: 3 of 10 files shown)


HARSHRAJ2789:
Read this against master at `72353a0` and ran the fixed call sites alongside the ones that were left.

`net.JoinHostPort` is the right tool here. What I could not account for is two other sites that read the same field and keep the old formatting.

`Route.IP` is not an independent value. `RouteFromSandbox` sets it at `sandboxroute/route.go:99`:

```go
ip := sandbox.Status.PodInfo.PodIP
```

So the string this PR brackets at `ext_proc.go:141` and `filter.go:203` is `Status.PodInfo.PodIP`. Two other call sites read that same field and still concatenate:

- `pkg/utils/runtime/runtime.go:59` in `GetRuntimeURL`: `ip := sbx.Status.PodInfo.PodIP`, then `fmt.Sprintf("http://%s:%d", ip, utils.RuntimePort)`
- `pkg/utils/proxyutils/default.go:40` in `requestSandbox`: `fmt.Sprintf("http://%s:%d%s", s.Status.PodInfo.PodIP, port, path)`

Both are live. `GetRuntimeURL` is called from `runtime/client.go:314`, `client.go:343` and `controller/sandbox/core/lifecycle_handler.go:56`. `requestSandbox` is `proxyutils.DefaultRequestFunc`, called from `sandbox-manager/infra/sandboxcr/sandbox.go:471`. The `client.go:314` call is the plain-HTTP branch of `resolveBaseURL`; the TLS branch above it formats `r.authority` instead, so the runtime path is affected only while runtime TLS is off.

The failure there is harder than the one this PR fixes. In the ext-proc and filter paths the malformed value lands in an Envoy header or metadata field. In these two it lands in a URL, and the URL does not parse:

```
FIXED    net.JoinHostPort           -> [2001:db8::1]:49999
UNFIXED  runtime.go:59              -> http://2001:db8::1:49999
UNFIXED  proxyutils/default.go:40   -> http://2001:db8::1:49999/health

url.Parse("http://[2001:db8::1]:49999")  -> host="2001:db8::1" port="49999"
url.Parse("http://2001:db8::1:49999")    -> ERROR: invalid port ":db8::1:49999" after host
http.NewRequest("http://2001:db8::1:49999")
                                         -> ERROR: invalid port ":db8::1:49999" after host
```

On an IPv6 pod IP those two fail before a request leaves the process, where the three sites this PR fixes would have misrouted instead. Same field, same sandbox.

Two more concatenations sit at `pkg/peers/memberlist.go:105` and `:136`, over `FindPodIP()` and `peer.Status.PodIP`. The description puts memberlist behavior out of scope, so I am only noting those.

The rest of what the same sweep turned up is not IPv6 exposure, in case it saves you the walk. `runtime/client.go:312` and `:357` format `r.authority`, which `WithAuthority` documents as the TLS SNI and certificate-verification hostname and which defaults to `RuntimeServerSNI`. `webhookutils/configuration/configuration.go:137` formats a webhook service host outside the sandbox data plane.

`go build ./...` is clean on `75cb824`, and `pkg/proxy`, `pkg/sandbox-gateway/filter`, `pkg/utils/runtime`, `pkg/peers` and `pkg/utils/proxyutils` all pass.

cyrilcsr:
> Read this against master at `72353a0` and ran the fixed call sites alongside the ones that were left.
> 
> `net.JoinHostPort` is the right tool here. What I could not account for is two other sites that read the same field and keep the old formatting.
> 
> `Route.IP` is not an independent value. `RouteFromSandbox` sets it at `sandboxroute/route.go:99`:
> 
> ```go
> ip := sandbox.Status.PodInfo.PodIP
> ```
> 
> So the string this PR brackets at `ext_proc.go:141` and `filter.go:203` is `Status.PodInfo.PodIP`. Two other call sites read that same field and still concatenate:
> 
> * `pkg/utils/runtime/runtime.go:59` in `GetRuntimeURL`: `ip := sbx.Status.PodInfo.PodIP`, then `fmt.Sprintf("http://%s:%d", ip, utils.RuntimePort)`
> * `pkg/utils/proxyutils/default.go:40` in `requestSandbox`: `fmt.Sprintf("http://%s:%d%s", s.Status.PodInfo.PodIP, port, path)`
> 
> Both are live. `GetRuntimeURL` is called from `runtime/client.go:314`, `client.go:343` and `controller/sandbox/core/lifecycle_handler.go:56`. `requestSandbox` is `proxyutils.DefaultRequestFunc`, called from `sandbox-manager/infra/sandboxcr/sandbox.go:471`. The `client.go:314` call is the plain-HTTP branch of `resolveBaseURL`; the TLS branch above it formats `r.authority` instead, so the runtime path is affected only while runtime TLS is off.
> 
> The failure there is harder than the one this PR fixes. In the ext-proc and filter paths the malformed value lands in an Envoy header or metadata field. In these two it lands in a URL, and the URL does not parse:
> 
> ```
> FIXED    net.JoinHostPort           -> [2001:db8::1]:49999
> UNFIXED  runtime.go:59              -> http://2001:db8::1:49999
> UNFIXED  proxyutils/default.go:40   -> http://2001:db8::1:49999/health
> 
> url.Parse("http://[2001:db8::1]:49999")  -> host="2001:db8::1" port="49999"
> url.Parse("http://2001:db8::1:49999")    -> ERROR: invalid port ":db8::1:49999" after host
> http.NewRequest("http://2001:db8::1:49999")
>                                          -> ERROR: invalid port ":db8::1:49999" after host
> ```
> 
> On an IPv6 pod IP those two fail before a request leaves the process, where the three sites this PR fixes would have misrouted instead. Same field, same sandbox.
> 
> Two more concatenations sit at `pkg/peers/memberlist.go:105` and `:136`, over `FindPodIP()` and `peer.Status.PodIP`. The description puts memberlist behavior out of scope, so I am only noting those.
> 
> The rest of what the same sweep turned up is not IPv6 exposure, in case it saves you the walk. `runtime/client.go:312` and `:357` format `r.authority`, which `WithAuthority` documents as the TLS SNI and certificate-verification hostname and which defaults to `RuntimeServerSNI`. `webhookutils/configuration/configuration.go:137` formats a webhook service host outside the sandbox data plane.
> 
> `go build ./...` is clean on `75cb824`, and `pkg/proxy`, `pkg/sandbox-gateway/filter`, `pkg/utils/runtime`, `pkg/peers` and `pkg/utils/proxyutils` all pass.

Thanks for the detailed sweep. Agreed—both paths use the same Pod IP and should be covered by this PR. I’ll update them and add IPv6 tests.

PR #904, opened 2026-08-31
fix(runtime): honor refresh and retry in files write

RedZapdos123:
## Description:

Honor runtime refresh and retry in the multipart files write path.

Before this change, `FilesystemAPI.Write` resolved transport once from the bound sandbox and sent a single multipart `/files` request. As a result, a caller-provided `WithRefresh` hook could not surface a newly stamped runtime URL or TLS transport, and transient transport or HTTP 5xx failures ignored the configured `WithRetry` policy.

This PR keeps the existing outward `/files` error behavior, but refreshes the sandbox before each attempt, rebuilds the request from a cached multipart payload, and adds regression tests for the missing-runtime-URL refresh path and a transient 500 response.

### Validation Tests run:

- [x] `go test ./pkg/utils/runtime -run 'TestWriteFileWithRuntime_RefreshResolvesRuntimeURL|TestWriteFileWithRuntime_RetriesTransientServerError' -count=1 -v`
- [x] `go test ./pkg/utils/runtime -run 'TestWriteFileWithRuntime_(InputValidation|HTTPInteractions|TransportError|HonorsLargeArgsTimeout|RefreshResolvesRuntimeURL|RetriesTransientServerError)|TestTLSMode_CapabilityGroupsUseRuntimeTLS|TestTLSMode_CapabilityGroupsRejectInvalidBundle' -count=1`

## Notes:

- Scope is intentionally limited to `FilesystemAPI.Write`.
- `ListDir` and `Remove` are left for a follow up PR if needed.

### DIFF
--- a/pkg/utils/runtime/filesystem.go
+++ b/pkg/utils/runtime/filesystem.go
@@ -34,6 +34,7 @@ import (
 	"time"
 
 	"connectrpc.com/connect"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 
 	agentsv1alpha1 "github.com/openkruise/agents/api/v1alpha1"
@@ -228,13 +229,8 @@ func (f *filesystemAPI) Write(ctx context.Context, req WriteFileRequest) (WriteF
 	}
 	sbx := f.r.sbx
 	log := klog.FromContext(ctx).WithValues("sandbox", klog.KObj(sbx)).V(utils.DebugLogLevel)
-
-	// The multipart write shares the transport decision with every other
-	// capability group; runtimeFilesHTTPClient stays the plaintext client so the
-	// existing test seam keeps working.
-	base, httpClient, err := f.r.resolveTransport(sbx, runtimeFilesHTTPClient)
-	if err != nil {
-		return WriteFileResult{}, err
+	if f.r.tlsEnabled && f.r.tlsConfigErr != nil {
+		return WriteFileResult{}, fmt.Errorf("invalid runtime TLS configuration: %w", f.r.tlsConfigErr)
 	}
 
 	username := req.Username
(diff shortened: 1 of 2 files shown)


HARSHRAJ2789 (commented):
The failing `unit-tests` check is real and it reproduces locally. `TestWriteFileWithRuntime_ContextTimeout` is not new here: it passes on the parent `815a36d` and fails on `399299d`, at `runtime_test.go:833` with "An error is expected but got nil".

The timeout itself is fine. Instrumenting the call shows the deadline firing on schedule and the request being abandoned:

```
elapsed    = 50ms
err        = <nil>
statusCode = 0
```

The error is lost between the closure and the caller. `retry.OnError` at `filesystem.go:255` pairs with `retriableRuntimeError`, which returns false once the context is done (`client.go:578`). That combination has a hole:

```
closure returned : failed to call runtime files API: Post "...": context deadline exceeded
retriable(err)   : false        (ctx expired, so lastErr is never assigned)
wait.Interrupted : true         (errors.Is matches context.DeadlineExceeded)
OnError returned : <nil>
```

`OnError` ends with `if wait.Interrupted(err) { err = lastErr }`, and `lastErr` is only assigned on the retriable branch. A context error is non-retriable by this predicate, so it reaches that line with `lastErr` still nil and is replaced by nil. `wait.Interrupted` matches `context.Canceled` and `context.DeadlineExceeded` as well as the wait timeout (`apimachinery@v0.35.0`, `wait/error.go:52`), so a wrapped deadline error qualifies.

Worth saying that this is not something the PR introduced. The same pairing is already on master at `client.go:558`, and running it there with master's own predicate returns nil for a wrapped deadline error and `connection refused` for a plain one. What this PR does is bring the write path under that pattern, where a test already existed to catch it.

The consequence on this path is the part that matters: a write that times out or is cancelled returns `(WriteFileResult{}, nil)`, so a caller checking only the error treats it as a successful write with a zero status code.

Capturing the attempt error alongside the return makes the test pass:

```go
var lastAttemptErr error
err = retry.OnError(f.r.backoff, retriableRuntimeError(reqCtx), func() error {
    ...
    lastAttemptErr = fmt.Errorf("failed to call runtime files API: %w", err)
    return lastAttemptErr
})
if err == nil && lastAttemptErr != nil {
    err = lastAttemptErr
}
```

With that, `go test ./pkg/utils/runtime/ -run TestWriteFileWithRuntime` passes. Doing it in `call` instead would cover `client.go:558` and every capability group that goes through it, which looks like the better home for the fix.

RedZapdos123:
I think the current CI test failure is related to #791.

RedZapdos123:
This E2E failure is separate from this PR.

[sandbox (with sandbox-gateway)](https://github.com/openkruise/agents/actions/runs/33385670156/job/99467618054?pr=904) failed during quota pool startup: `quota-small` stayed `1/2 Running` and `AVAILABLE=0` before this PR's `FilesystemAPI.Write` path ran.

Tracked the issue in #908. Created the PR: #909 which adds a liveness probe so failed Code Interpreter startup restarts instead of remaining permanently unready.

The CI job lacks sandbox logs so the exact startup error is not proven there. Local reproduction matches this pod state.

PR #906, opened 2026-08-31
fix(e2e): raise envoy ext_proc message timeout and tolerate transient 504s

furykerry:

The E2B E2E suite intermittently fails with 504s from envoy right after deployment (e.g. [run 33379976653](https://github.com/openkruise/agents/actions/runs/33379976653/job/99449891153), and earlier runs 33350342490 / 33368030915 on unrelated branches). The root cause:

- All requests through the sandbox-manager envoy pass the `ext_proc` filter, whose per-message timeout is unset and therefore defaults to **200ms**. When the sandbox-manager pod is CPU-throttled during startup or warm-pool creation bursts, its in-memory ext-proc handler can miss that budget, and envoy replies `504 Gateway Timeout` with an empty body (verified against envoy v1.33 `Filter::onMessageTimeout`).
- The autouse `wait_for_sandbox` fixture raises on the first `list_sandbox()` error, so one transient 504 kills the whole run (pytest `-x`).

This PR:
1. Sets `message_timeout: 5s` on the ext_proc filter in `config/sandbox-manager/envoy-config.yaml`. The handler is a pure in-memory lookup (sub-ms normally), so 5s still fails fast on a truly stalled processor while absorbing scheduler latency spikes.
2. Hardens `test/e2b/utils.py`: the fixture now retries transient API errors (5xx / connection / timeout, parsed from the SDK's `"{status}: {body}"` message since it exposes no structured status) within the existing cleanup window; 4xx/unknown errors still fail fast. If listing never succeeds before the window expires, it dumps manager logs and raises the last error instead of force-killing on a broken API path.

NONE

1. Config: the rendered envoy ConfigMap parses and the ext_proc filter carries `message_timeout: 5s` (checked with YAML load of `envoy-config.yaml`).
2. Test helper: `_is_transient_api_error` validated against the real flake signatures — `504: b''` and `504: b'stream timeout'` → retried; `401/404` and unknown errors → fail fast; `ConnectionError`/`TimeoutError` → retried.
3. E2E: `E2E for E2B SDK` workflows should no longer die on a single post-deployment 504 during fixture setup.

- Only the "without sandbox-gateway" envoy config carries ext_proc; the sandbox-gateway deployment has no envoy ext_proc filter, so no other config needs the same change.
- The retry is bounded by the existing `sandbox_cleanup_timeout` window — no new unbounded wait is introduced; a persistent outage still fails (with manager logs dumped).
- No behavior change for non-transient errors: auth/config failures (4xx) raise immediately as before.

### DIFF
--- a/config/sandbox-manager/envoy-config.yaml
+++ b/config/sandbox-manager/envoy-config.yaml
@@ -65,6 +65,11 @@ data:
                           grpc_service:
                             envoy_grpc:
                               cluster_name: ext_proc_cluster
+                          # The ext-proc server (sandbox-manager) answers from memory in
+                          # milliseconds, but the default per-message timeout (200ms) is
+                          # too tight when the manager pod is CPU-throttled during
+                          # startup or sandbox-creation bursts, causing flaky 504s.
+                          message_timeout: 5s
                           processing_mode:
                             request_header_mode: SEND
                             response_header_mode: SEND
--- a/test/e2b/test_api_key_quota.py
+++ b/test/e2b/test_api_key_quota.py
@@ -1366,10 +1366,21 @@ def test_redis_data_loss_rebuilds_from_live_crs_and_reenforces(sandbox_context):
         wait_until(lambda: assert_redis_count(created_id, "all", 2), timeout=120)
 
         try:
-            redis_cli("DEL", quota_live_key(created_id))
-            redis_cli("DEL", quota_sum_key(created_id, "sandbox.count"))
-            redis_cli("DEL", quota_sum_key(created_id, "limits.cpu"))
-            redis_cli("DEL", quota_sum_key(created_id, "limits.memory"))
+            # Delete all quota keys atomically in a single redis-cli call.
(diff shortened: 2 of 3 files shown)


furykerry:
Second commit (`996cba575`) fixes an unrelated quota-redis-fault flake surfaced once the suite got past the ext_proc 504: `test_redis_data_loss_rebuilds_from_live_crs_and_reenforced` deleted `q:live` and the three `q:sum` hashes via four sequential `kubectl exec` redis-cli DELs. The gaps between DELs let the anti-drift driver re-acquire the live entries and rewrite the sums; the remaining DELs then left `q:live` intact but `q:sum` empty — a state anti-drift never repairs because it only diffs observed sandboxes against `q:live` entries (`count[all]=0` forever, see the diagnostic dump in run 33391726632).

The fix deletes all four keys with a single atomic `redis-cli DEL`, matching the spec's atomic data-loss model and the live+sums co-write invariant (`docs/specs/2026-06-17-api-key-sandbox-quota-redis.md` §7). Test-only change.

PR #907, opened 2026-08-31
fix(storage-cli): avoid copying CSI publish requests

### LINKED ISSUE #905: [BUG] go vet fails on master, breaking make build and make test-e2e; no CI job runs it
**What happened**:

#876 bumped `github.com/container-storage-interface/spec` from v1.9.0 to v1.13.0 and merged on 2026-08-28 with 29/29 checks green. It also broke `go vet ./...`, which now exits 1 on master at `5a5bc10` with 18 findings, five of them in non-test code:

```
pkg/agent-runtime/storage-cli/main.go:108:34: call of validateGeneralParams copies lock value: ...csi.NodePublishVolumeRequest contains ...protoimpl.MessageState contains sync.Mutex
pkg/agent-runtime/storage-cli/main.go:129:29: call of provider.Validate copies lock value: ...
pkg/agent-runtime/storage-cli/main.go:138:48: call of provider.Mount copies lock value: ...
pkg/agent-runtime/storage-cli/main.go:197:35: validateGeneralParams passes lock by value: ...
pkg/agent-runtime/storage-cli/storage/csi_runner.go:52:67: RunNodePublishVolume passes lock by value: ...
```

Two Makefile targets declare `vet` as a prerequisite, so both now stop before doing any work:

```
Makefile:81   build: generate fmt vet manifests
Makefile:110  test-e2e: setup-test-e2e manifests generate fmt vet
```

`make build` produces no binary on a clean checkout of master.

`go vet ./...` stays clean across a dependency bump, and `make build` builds `bin/agent-sandbox-controller`.

```
$ git checkout d553423        # parent of the bump, spec v1.9.0
$ go vet ./... ; echo $?
0

$ git checkout e9e5e08        # the bump itself, spec v1.13.0
$ go vet ./... ; echo $?      # 18 findings
1

$ git checkout 5a5bc10        # current master
$ make vet
make: *** [vet] Error 1
$ ls bin/agent-sandbox-controller
ls: no such file or directory
```

The mechanism is the regenerated protobuf. `csi.pb.go` in v1.9.0 contains `protoimpl.MessageState` zero times. v1.13.0 was regenerated with `protoc-gen-go v1.36.10`, and `NodePublishVolumeRequest` now opens with a `state protoimpl.MessageState` field tagged `protogen:"open.v1"`. `MessageState` embeds `pragma.DoNotCopy`, which `vet` treats as a lock. The repo passes that struct by value in the five places above, which was fine while the struct had no such field.

One point of precision, because "contains sync.Mutex" reads worse than it is. `pragma.DoNotCopy` is declared `[0]sync.Mutex`, a zero-length array that carries no lock at runtime and exists so that `vet` flags copies; its comment says it "does not rely on a Go language feature, but rather a special case within the vet checker". So this is protobuf's no-copy contract being reported, and I am not claiming an observed failure in the storage CLI. The concrete breakage is the vet exit code and the two targets that depend on it.

On why CI stayed green through the bump. No workflow invokes `vet`, `build`, or `test-e2e`. The complete set of make targets called from `.github/workflows` is `build-okactl`, `deploy-*`, `docker-*`, `generate`, `ginkgo` and `test`, and `test` runs `go test -race ./pkg/...` directly with no `vet` prerequisite. `golangci-lint` does enable `govet`, but the CI action sets `only-new-issues: true` at `.github/workflows/ci.yaml:77`.

The narrow fix is to take `*csi.NodePublishVolumeRequest` through `validateGeneralParams`, the `Validate`/`Mount`/`Unmount` provider methods and `RunNodePublishVolume`. The broader one is a CI step running `go vet ./...`. #876 touched only `go.mod` and `go.sum`, so no Go source line changed for the lint job to attribute a finding to, and the same is true of every bump: #869 landed `google.golang.org/protobuf` v1.36.12 on 31 August with all checks green while `go vet` was already failing, and eleven more dependabot PRs are open behind it.

**Environment**:
- OpenKruiseAgent version: master @ `5a5bc10`
- Kubernetes version: not applicable, reproduced at build time
- Others: reproduced with `GOTOOLCHAIN=go1.25.9` (the version in `go.mod`) and with go1.26.5; identical 18 findings on both

RedZapdos123:
## Description:

Avoid copying `csi.NodePublishVolumeRequest` through the storage CLI mount path.

Before this change, `runMount`, `validateGeneralParams`, `Provider.Validate`, `Provider.Mount`, and `RunNodePublishVolume` passed `csi.NodePublishVolumeRequest` by value. After `github.com/container-storage-interface/spec` moved to `v1.13.0`, the regenerated protobuf embedded `protoimpl.MessageState`, so those copies started tripping `go vet` copylocks diagnostics and made `make vet` fail on `main`.

This PR switches that request plumbing to `*csi.NodePublishVolumeRequest` and updates the affected tests and fakes to match. Runtime behavior stays the same, but the storage CLI path no longer violates the protobuf no-copy contract. Fixes #905.

### Validation Tests run:

- [x] `GOTOOLCHAIN=go1.25.9 go test ./pkg/agent-runtime/storage-cli/... -count=1`
- [x] `GOTOOLCHAIN=go1.25.9 go vet ./pkg/agent-runtime/storage-cli/...`
- [x] `GOTOOLCHAIN=go1.25.9 go vet ./...`
- [x] `GOTOOLCHAIN=go1.25.9 make vet`

## Notes:

- Scope is intentionally limited to the storage CLI CSI publish-request path.
- No functional behavior change is intended beyond removing the by-value protobuf copies that were breaking `go vet`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 build: generate fmt vet manifests ## Build manager binary.
-	go build -o bin/agent-sandbox-controller cmd/agent-sandbox-controller/main.go
+	go build -o bin/agent-sandbox-controller ./cmd/agent-sandbox-controller
 
 .PHONY: build-okactl
 build-okactl: ## Build okactl CLI binary.
--- a/pkg/agent-runtime/storage-cli/main.go
+++ b/pkg/agent-runtime/storage-cli/main.go
@@ -88,8 +88,8 @@ func runMount(cmd *cobra.Command) error {
 		return fmt.Errorf("failed to decode CSI request config: %w", err)
 	}
 
-	csiReq := csi.NodePublishVolumeRequest{}
-	if err := proto.Unmarshal(configRaw, &csiReq); err != nil {
+	csiReq := &csi.NodePublishVolumeRequest{}
+	if err := proto.Unmarshal(configRaw, csiReq); err != nil {
 		cmd.Help() // #nosec G104 -- help output error is non-actionable
 		return fmt.Errorf("failed to unmarshal CSI request: %w", err)
 	}
@@ -195,7 +195,7 @@ func main() {
 	}
 }
 
-func validateGeneralParams(csiReq csi.NodePublishVolumeRequest) error {
+func validateGeneralParams(csiReq *csi.NodePublishVolumeRequest) error {
 	if strings.TrimSpace(csiReq.VolumeContext["csi.storage.k8s.io/pod.uid"]) == "" {
 		return fmt.Errorf("Pod UID is required. Use csi.storage.k8s.io/pod.uid setting")
 	}
--- a/pkg/agent-runtime/storage-cli/main_test.go
+++ b/pkg/agent-runtime/storage-cli/main_test.go
@@ -283,8 +283,8 @@ func TestRootCmdExecute_VersionSubcommand(t *testing.T) {
 
 // newCSIRequestWithVolumeContext builds a minimal NodePublishVolumeRequest
 // with the provided volume context; used to drive validateGeneralParams.
-func newCSIRequestWithVolumeContext(ctx map[string]string) csi.NodePublishVolumeRequest {
-	return csi.NodePublishVolumeRequest{VolumeContext: ctx}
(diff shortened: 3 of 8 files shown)


RedZapdos123:
Closed this as #940 already fixed the issue #905.

PR #912, opened 2026-09-01
fix(sandbox): sync pod status before upgrade initialization

zmberg:

Syncs the resumed Pod status before running upgrade initialization.

For a paused Sandbox undergoing a Recreate upgrade with a PreUpgrade hook, the resumed Pod may have a different IP. Previously, initialization used the stale Pod information from Sandbox status, causing runtime `/init` requests to target the old Pod IP and leaving the upgrade stuck in Resuming.

This change updates SandboxIp and PodInfo from the Ready Pod before calling Initialize. It also adds regression coverage for the new Pod IP and UID, including the initialization failure path.

NONE

```bash
GOTOOLCHAIN=go1.26.3 go test -mod=mod ./pkg/controller/sandbox/core -run '^TestEnsureSandboxUpgraded_Resuming$' -count=1
```

Please verify the ordering in handleResuming and the status retained when Initialize returns an error.

### DIFF
--- a/pkg/controller/sandbox/core/upgrade_control.go
+++ b/pkg/controller/sandbox/core/upgrade_control.go
@@ -350,6 +350,10 @@ func (r *UpgradeControl) handleResuming(ctx context.Context, args EnsureFuncArgs
 		klog.InfoS("Waiting for pod ready before initialization", "sandbox", klog.KObj(box))
 		return nil
 	}
+	// Sync status from the resumed pod before Initialize so that runtime
+	// re-init targets the current Pod IP/UID. The resumed pod may differ
+	// from the paused one, and using stale status would leave the upgrade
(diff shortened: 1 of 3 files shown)


pkg/controller/sandbox/core/upgrade_control.go
```diff
@@ -350,6 +350,7 @@ func (r *UpgradeControl) handleResuming(ctx context.Context, args EnsureFuncArgs
 		klog.InfoS("Waiting for pod ready before initialization", "sandbox", klog.KObj(box))
 		return nil
 	}
+	r.syncStatusFromPod(pod, newStatus, false)
```
furykerry, line 357: plz leave a comment to explain syncStatusFromPod must be placed before any usage of sandbox ip

PR #920, opened 2026-09-02
feat(sandbox-manager): add network interface binding and reliable peer discovery

AiRanthem:

This PR adds an optional `--network-interface` flag for Sandbox Manager and uses its validated global-unicast address, preferring IPv4 over IPv6, consistently for the control API, peer route listener, memberlist bind/advertise address, and non-Sandbox proxy return path.

It also makes peer discovery resilient to startup ordering and temporary Kubernetes API failures:

- validates the namespace, label selector, REST configuration, and selected network interface at startup;
- lists selector-matching peer Pods with bounded retries and attempts every trusted seed in stable order during each cycle;
- validates, deduplicates, and stably orders trusted seed addresses;
- starts required listeners synchronously so bind failures reach the caller;
- serializes Start/Stop and gives one lifecycle owner responsibility for Leave and Shutdown;
- preserves upstream traffic-token and dedicated observability-listener behavior.

The design and lifecycle contract are documented in the accompanying proposal.

NONE

- `go test -mod=mod ./pkg/peers ./pkg/proxy ./pkg/sandbox-gateway/server ./pkg/sandbox-manager ./pkg/servers/e2b ./pkg/utils/network -count=1`
- `go build -mod=mod ./cmd/sandbox-manager`

- An empty `--network-interface` preserves the existing non-hosted behavior.
- A configured interface must be up and have exactly one global-unicast address in the preferred available family: IPv4 if any global-unicast IPv4 address exists, otherwise IPv6. Startup fails instead of falling back to another interface.
- The ext-proc listener, metrics, pprof, and other observability listeners remain outside the selected user-network address contract.
- Peer discovery attempts every trusted seed in a cycle, even after a successful Join. After the cycle completes, it stops Kubernetes listing and retrying if any Join succeeded; memberlist handles ongoing membership afterward.
- An in-flight memberlist Join is bounded by existing memberlist network deadlines because the upstream Join API has no context parameter.


pkg/utils/network/network.go
```diff
@@ -14,16 +14,71 @@ See the License for the specific language governing permissions and
...
 )
 
+// ListenAddress returns a TCP listen address for host and port.
+// An empty host yields ":port", which listens on all local addresses.
+func ListenAddress(host string, port int) string {
+	return net.JoinHostPort(host, strconv.Itoa(port))
+}
+
+// ResolveNetworkInterfaceAddress returns the single global-unicast IPv4 address
+// on the named interface. An empty name returns an empty address.
```
furykerry: consider add TODO for ipv6 support

pkg/peers/memberlist.go
```diff
@@ -171,57 +182,191 @@ func (m *MemberlistPeers) Start(ctx context.Context, bindPort int) error {
...
+		// Wait a full interval after each failed attempt, however long the
+		// attempt itself took.
+		retry := time.NewTimer(m.retryInterval)
+		select {
+		case <-ctx.Done():
+			retry.Stop()
+		case <-retry.C:
+		}
+	}
+	<-ctx.Done()
```
furykerry, line 224: <-ctx.Done() is redundant， When the for loop exits, ctx.Err() is already non-nil, meaning the context is canceled. <-ctx.Done() will return immediately because the channel is closed, so this line has no functional effect. It can be removed to reduce confusion.

pkg/utils/network/network.go
AiRanthem: Added a TODO on ResolveNetworkInterfaceAddress. IPv6 remains out of this proposal's scope.

pkg/peers/memberlist.go
AiRanthem: Keeping `<-ctx.Done()`. The retry loop also exits via `break` when tryJoin succeeds, at which point ctx is still active. That wait parks the lifecycle owner until Stop/parent cancel so Leave/Shutdown still run from here. Added a comment to make that explicit.

PR #930, opened 2026-09-03
fix(api): align CRD validation and status

furykerry:

- Bound probe and cron-policy inputs to protect periodically evaluated API fields.
- Make sandbox status schedules a reason-keyed map list for server-side apply,
  and require the list key.
- Align PoolAutoscaler status generation storage and condition constant names
  without changing condition wire values.
- Regenerate the API artifacts from the updated type definitions.

NONE

- `go test ./api/v1alpha1/... ./pkg/controller/poolautoscaler/... ./pkg/controller/sandbox/... ./pkg/autopause/...`
- `make generate manifests`
- `git diff --check`

- `Schedule.reason` is now required because it is the server-side-apply
  map-list key.
- The rebase retains upstream name-keyed list semantics for probes and cron
  policies while adding the new size limits.


zmberg (commented):
**[Suggestion]** This PR fixes recurring API schema omissions. Could we capture a checklist in `api/v1alpha1/AGENTS.md`? It should cover list topology, required and validated map keys, bounds for user-controlled collections and evaluated strings, shared-type propagation across CRDs, and `make generate manifests` followed by a clean second generation run.

PR #933, opened 2026-09-03
[Bug]: distinguish unsupported sandbox in-place resize

googs1025:

This PR refines Sandbox in-place resource resize failure reporting for clusters where the Pod resize path is not supported.

Before this change, a resize capability mismatch was reported with the same generic condition reason used by other in-place update failures:

```text
InplaceUpdate=False
Reason=Failed
Message=in-place pod resize not supported
```

That message is visible to operators, but the reason is not machine-readable enough to distinguish an unsupported cluster/runtime resize path from other generic in-place update failures.

This PR adds one structured `InplaceUpdate` reason:

- `UnsupportedResize`: the cluster/runtime resize path is not supported.

Other terminal resource resize failures, such as kubelet reporting an infeasible resize after the Pod has moved to the target revision, continue to use the existing `Failed` reason.

The claim wait-ready behavior intentionally remains compatibility-preserving. `NewSandboxWaitReadyTask` only waits while the Sandbox reports `InplaceUpdating`; terminal in-place update failure reasons, including `UnsupportedResize`, fall through to the existing usable-Sandbox readiness check. As a result, this PR improves status classification without changing the current best-effort claim semantics.

Broader API semantics, such as a future `spec.inplaceUpdate.failurePolicy` with `BestEffort`, `Fail`, or `CreateNew`, are tracked in #934 and should be designed separately.

Related to #934

```bash
go test ./api/v1alpha1 -count=1
go test ./pkg/cache -run 'TestNewSandboxWaitReadyTask|TestCache_CountActiveSandboxes' -count=1
go test ./pkg/controller/sandbox/core -run 'TestHandleInPlaceUpdateCommon|TestIsInplaceUpdateTerminal' -count=1
go test ./pkg/controller/sandboxclaim/core -run 'Test.*Claim|Test.*Inplace|Test.*buildClaimOptions' -count=1
make generate manifests
```

This PR does not introduce a fallback policy and does not make `UnsupportedResize` directly control SandboxClaim success. It only makes the unsupported resize path explicit while preserving the existing best-effort claim behavior.

### DIFF
--- a/api/v1alpha1/sandbox_types.go
+++ b/api/v1alpha1/sandbox_types.go
@@ -607,9 +607,10 @@ const (
 	SandboxReadyReasonPodCreateFailed      = "PodCreateFailed"
 
 	// SandboxConditionInplaceUpdate Reason
-	SandboxInplaceUpdateReasonInplaceUpdating = "InplaceUpdating"
-	SandboxInplaceUpdateReasonSucceeded       = "Succeeded"
-	SandboxInplaceUpdateReasonFailed          = "Failed"
+	SandboxInplaceUpdateReasonInplaceUpdating   = "InplaceUpdating"
+	SandboxInplaceUpdateReasonSucceeded         = "Succeeded"
+	SandboxInplaceUpdateReasonFailed            = "Failed"
+	SandboxInplaceUpdateReasonUnsupportedResize = "UnsupportedResize"
 
 	// SandboxConditionUpgrading Reason
 	SandboxUpgradingReasonResuming         = "Resuming"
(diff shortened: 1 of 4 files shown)


pkg/cache/tasks.go
```diff
@@ -132,6 +137,10 @@ func (c *Cache) NewSandboxWaitReadyTask(ctx context.Context, sbx *agentsv1alpha1
 	)
 }
 
+func isTerminalResourceResizeFailure(message string) bool {
```
googs1025: this is not good way to fix.
will disscus in issue

PR #936, opened 2026-09-04
fix(sandbox): propagate startup failures to wait-ready and ScalingLimited

ywExcellent:

Two changes that make observed sandbox startup failures visible sooner:

1. **Fail the wait-ready task fast on `PodCreateFailed`.** The cache wait-ready task previously failed fast only when the Ready condition reason was `StartContainerFailed`. It now also fails fast on `PodCreateFailed` — which the sandbox pod control sets when the K8s create call is rejected (e.g. invalid PVC, quota exceeded) — and the error reports the actual Ready reason and message instead of assuming a container startup failure.
2. **Requeue the SandboxSet when a Sandbox enters or leaves a startup-failure classification.** A Sandbox update used to be ignored by the event handler unless its state changed; since a startup failure happens while the Sandbox stays `Pending`, the `ScalingLimited` condition only reflected the new failure once the Pending timeout elapsed. The handler now also enqueues the owning SandboxSet when a Sandbox enters or leaves a startup-failure classification, so the `Failed` count is updated as soon as the failure is observed.

NONE

1. `go test ./pkg/cache/ -run TestNewSandboxWaitReadyTask -count=1`
2. `go test ./pkg/controller/sandboxset/ -run TestSandboxEventHandler -count=1`
3. `go test ./pkg/sandbox-manager/infra/sandboxcr/ -count=1`

- `isSandboxStartupFailureReason` in `pkg/cache/tasks.go` and the existing `isStartupFailure` in `pkg/controller/sandboxset/scaling_limited.go` intentionally enumerate the same two reasons — `StartContainerFailed` and `PodCreateFailed`; no new Sandbox Ready reasons are introduced.
- The wait-ready error string changed from `sandbox start container failed: %s` to `sandbox startup failed (reason=%s): %s`; `claim_test.go` assertions are updated accordingly.


pkg/cache/tasks.go
```diff
@@ -110,15 +110,25 @@ func (c *Cache) NewSandboxResumeTask(ctx context.Context, sbx *agentsv1alpha1.Sa
 	)
 }
 
+func isSandboxStartupFailureReason(reason string) bool {
```
furykerry: isSandboxStartupFailureReason and isStartupFailure has similar logic, consider move them to shared utility package e.g. pkg/utils/utils.go, and make isStartupFailure calling isSandboxStartupFailureReason

pkg/cache/tasks.go
```diff
@@ -110,15 +110,25 @@ func (c *Cache) NewSandboxResumeTask(ctx context.Context, sbx *agentsv1alpha1.Sa
...
+	switch reason {
+	case agentsv1alpha1.SandboxReadyReasonStartContainerFailed,
+		agentsv1alpha1.SandboxReadyReasonPodCreateFailed:
```
furykerry: checkSandboxReady also check SandboxReadyReasonStartContainerFailed, but checkSandboxReady is only used in unit test TestCheckSandboxInplaceUpdate, and TestCheckSandboxInplaceUpdate is actually test nothing , consider move checkSandboxReady and related ut TestCheckSandboxInplaceUpdate

ywExcellent:
Addressed both comments in 7228de1a:

1. Moved the startup-failure reason check to `pkg/utils` as `IsSandboxStartupFailureReason` (with a table-driven unit test); both the cache wait-ready task and the SandboxSet `isStartupFailure` now call it.
2. Removed `checkSandboxReady` and `TestCheckSandboxInplaceUpdate` — `checkSandboxReady` had no production callers left (the claim path waits via the cache `NewSandboxWaitReadyTask`), so the test only exercised the dead function.

Thanks for the review!

PR #940, opened 2026-09-04
fix: unbreak go vet and make build on master

### LINKED ISSUE #905: [BUG] go vet fails on master, breaking make build and make test-e2e; no CI job runs it

omlahore:
Fixes #905

Two separate things stop `make build` producing a binary on a clean checkout of master. The issue diagnosed the first; the second only becomes visible once the first is fixed.

## 1. go vet fails, and vet gates the build

#876 bumped `container-storage-interface/spec` from v1.9.0 to v1.13.0. In v1.13.0 `csi.NodePublishVolumeRequest` carries a `protoimpl.MessageState`, which embeds a `sync.Mutex`. The storage CLI passes that struct **by value** everywhere, so vet now reports 18 copylocks findings and exits 1:

```
pkg/agent-runtime/storage-cli/storage/csi_runner.go:52:67: RunNodePublishVolume passes lock by value
pkg/agent-runtime/storage-cli/main.go:198:35: validateGeneralParams passes lock by value
...
```

`Makefile:81` and `Makefile:110` both declare `vet` as a prerequisite, so `build` and `test-e2e` stop before doing anything.

Fixed by taking `*csi.NodePublishVolumeRequest` through the `Provider` interface, `RunNodePublishVolume` and `validateGeneralParams`. That is the ordinary way to handle a generated protobuf message, and it removes a real struct copy from the mount path rather than just silencing the check. `main.go` already did `proto.Unmarshal(configRaw, &csiReq)`, so it had the pointer in hand anyway.

No production `Provider` implementations exist outside the test fakes, so the interface change is contained.

## 2. The build target compiles a file, not a package

With vet clean, `make build` gets further and then fails:

```
cmd/agent-sandbox-controller/main.go:367:2: undefined: executeCABindings
```

`executeCABindings` lives in `cmd/agent-sandbox-controller/ca_binding.go`, same package. The target was:

```make
go build -o bin/agent-sandbox-controller cmd/agent-sandbox-controller/main.go
```

Naming a single `.go` file builds only that file, so its siblings are invisible. `go build ./cmd/agent-sandbox-controller/` has always worked, which is why this never showed up in `go build ./...` or in CI.

Changed to build the package. The `build-okactl` target two lines below already uses `./cmd/okactl`, so this makes the two consistent.

## Verification

```
$ go vet ./... ; echo $?
0

$ make build
...
go build -o bin/agent-sandbox-controller ./cmd/agent-sandbox-controller
$ ls -l bin/agent-sandbox-controller
-rwxrwxr-x 1 ... 103290525 ... bin/agent-sandbox-controller
```

`go test ./pkg/agent-runtime/storage-cli/...` passes across all four packages.

37 insertions, 39 deletions. Most of it is the mechanical `csi.NodePublishVolumeRequest` to `*csi.NodePublishVolumeRequest` change in the test fakes.

## One note on scope

`make build` runs `go fmt ./...`, which reformatted four unrelated test files on my machine (`pkg/identity/ca_cert_injector_test.go` and three others, all struct-field alignment). I reverted those rather than fold unrelated churn in here, but they are unformatted on master and something will keep re-flagging them. Worth a separate cleanup.

## Why CI missed it

No workflow runs `go vet`. The `undefined: executeCABindings` case is worse, since it only appears in `make build` and never in `go build ./...`. A CI job that runs `make build` would catch both.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 build: generate fmt vet manifests ## Build manager binary.
-	go build -o bin/agent-sandbox-controller cmd/agent-sandbox-controller/main.go
+	go build -o bin/agent-sandbox-controller ./cmd/agent-sandbox-controller
 
 .PHONY: build-okactl
 build-okactl: ## Build okactl CLI binary.
--- a/pkg/agent-runtime/storage-cli/main.go
+++ b/pkg/agent-runtime/storage-cli/main.go
@@ -88,8 +88,8 @@ func runMount(cmd *cobra.Command) error {
 		return fmt.Errorf("failed to decode CSI request config: %w", err)
 	}
 
-	csiReq := csi.NodePublishVolumeRequest{}
-	if err := proto.Unmarshal(configRaw, &csiReq); err != nil {
+	csiReq := &csi.NodePublishVolumeRequest{}
+	if err := proto.Unmarshal(configRaw, csiReq); err != nil {
 		cmd.Help() // #nosec G104 -- help output error is non-actionable
 		return fmt.Errorf("failed to unmarshal CSI request: %w", err)
 	}
@@ -195,7 +195,7 @@ func main() {
 	}
 }
 
-func validateGeneralParams(csiReq csi.NodePublishVolumeRequest) error {
+func validateGeneralParams(csiReq *csi.NodePublishVolumeRequest) error {
 	if strings.TrimSpace(csiReq.VolumeContext["csi.storage.k8s.io/pod.uid"]) == "" {
 		return fmt.Errorf("Pod UID is required. Use csi.storage.k8s.io/pod.uid setting")
 	}
--- a/pkg/agent-runtime/storage-cli/main_test.go
+++ b/pkg/agent-runtime/storage-cli/main_test.go
@@ -283,8 +283,8 @@ func TestRootCmdExecute_VersionSubcommand(t *testing.T) {
 
 // newCSIRequestWithVolumeContext builds a minimal NodePublishVolumeRequest
 // with the provided volume context; used to drive validateGeneralParams.
-func newCSIRequestWithVolumeContext(ctx map[string]string) csi.NodePublishVolumeRequest {
-	return csi.NodePublishVolumeRequest{VolumeContext: ctx}
+func newCSIRequestWithVolumeContext(ctx map[string]string) *csi.NodePublishVolumeRequest {
+	return &csi.NodePublishVolumeRequest{VolumeContext: ctx}
 }
 
 // captureStdout swaps os.Stdout for the duration of fn and returns the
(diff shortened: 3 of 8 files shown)


BH4AWS:
LGTM

PR #942, opened 2026-09-07
fix(sandbox): classify unschedulable startup failures

ywExcellent:

- Adds `Unschedulable` as a shared Sandbox Ready failure reason for `PodScheduled=False/Unschedulable`.
- Watches PodScheduled condition changes so scheduling failures and recoveries are reconciled promptly.
- Counts unschedulable sandboxes in the SandboxSet startup budget, fails wait-ready early, and reports the scheduling reason during claim diagnostics.

NONE

- `go test ./pkg/controller/sandbox -run '^TestIsActivePodUpdate$' -count=1`
- `go test ./pkg/controller/sandbox/core -run 'Test(ClassifyStartupFailure|DefaultSyncStatusFromPodStartupFailure)$' -count=1`
- Targeted tests for `pkg/utils`, `pkg/controller/sandboxset`, `pkg/cache`, and `pkg/sandbox-manager/infra/sandboxcr`.

- Recovery clears normalizer-owned `Unschedulable` and `StartContainerFailed` reasons, while preserving `PodCreateFailed`.
- CRD documentation is regenerated for the updated startup-budget contract.

---

- Regenerated SandboxSet CRD description (`config/crd/bases/agents.kruise.io_sandboxsets.yaml`).

### DIFF
--- a/pkg/sandbox-manager/infra/sandboxcr/claim.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/claim.go
@@ -1075,15 +1075,15 @@ func sandboxReadyFailureReason(sbx *v1alpha1.Sandbox, state string, readyCond, i
 	if inplaceCond.Reason == v1alpha1.SandboxInplaceUpdateReasonInplaceUpdating {
 		return "inplace update is still in progress"
 	}
-	if sbx.Status.PodInfo.PodIP == "" {
-		return "sandbox has no pod IP"
-	}
 	if utils.IsSandboxStartupFailureReason(readyCond.Reason) {
 		reason := fmt.Sprintf("ready condition reports %s", readyCond.Reason)
 		if readyCond.Message != "" {
 			reason = fmt.Sprintf("%s: %s", reason, readyCond.Message)
 		}
 		return reason
+	}
+	if sbx.Status.PodInfo.PodIP == "" {
+		return "sandbox has no pod IP"
 	}
 	if state != v1alpha1.SandboxStateRunning {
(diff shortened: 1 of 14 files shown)


pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -1078,7 +1078,7 @@ func sandboxReadyFailureReason(sbx *v1alpha1.Sandbox, state string, readyCond, i
 	if sbx.Status.PodInfo.PodIP == "" {
 		return "sandbox has no pod IP"
 	}
-	if readyCond.Reason == v1alpha1.SandboxReadyReasonStartContainerFailed {
+	if utils.IsSandboxStartupFailureReason(readyCond.Reason) {
```
zmberg, line 1078: **[Issue]** An Unschedulable Pod typically has not been assigned to a node and therefore has no PodIP. The preceding PodIP check returns `sandbox has no pod IP` first, so the newly handled Unschedulable reason and scheduler message cannot appear in timeout diagnostics. Please check startup failures before PodIP and add a test covering an empty PodIP with an Unschedulable condition.

ywExcellent: Good catch, thanks! Fixed in e940c285: the startup-failure reason check now runs before the pod IP check, so an unschedulable sandbox reports `ready condition reports Unschedulable: <scheduler message>` instead of the generic `sandbox has no pod IP`. Also added a test case covering an empty PodIP with an Unschedulable ready condition (`unschedulable reason reported without pod ip`).

PR #950, opened 2026-09-10
Tracing user operation

Liquorice-Ma:

Improve tracing observability across sandbox-manager and the controller to correlate user operations, controller logs, and Sandbox state transitions.

- **Trace-first logging:** Use single-line JSON controller logs with `traceID` as the first field when present, and include caller information.
- **Condition visibility:** Record `condition.<Type>=Status:Reason` attributes on Reconcile and updateSandboxStatus spans to help diagnose lifecycle delays and compare condition changes.
- **Operation propagation:** Carry user operations through OTel Baggage and CR annotations, exposing them as `traceOperation` in controller logs. Lifecycle endpoints use `create`, `pause`, `resume`, and `kill`; other endpoints default to the HTTP method and route pattern.
- **Single-line stdout spans:** Disable pretty-printing in std mode so newline-based log collectors do not split a span into multiple records.

NONE

The following build and unit tests passed:

```bash
go build ./...
go test ./pkg/tracing ./pkg/servers/web ./pkg/servers/e2b/... -count=1
```

New encoder tests cover field ordering, missing trace IDs, chained Logger.With context, cloning, and console fallback.

Suggested runtime verification with tracing enabled:
1. Perform create/pause/resume/kill operations and verify the corresponding controller logs contain a leading `traceID` field and the expected `traceOperation`.
2. Compare `condition.*` attributes on Reconcile and updateSandboxStatus spans.
3. Verify each span exported in std mode occupies a single JSON line.

- **Logging compatibility:** The controller explicitly selects the JSON encoder, so `--zap-encoder` no longer changes the output format. Log parsers relying on the previous console format may need adjustment.
- Connect is labeled `resume` only when it triggers recovery of a paused Sandbox; connecting to a running Sandbox retains the default route pattern.
- Baggage is propagated through the new `agents.kruise.io/trace-baggage` annotation without changing the CRD schema.
- Only std mode disables pretty-printing; file mode remains unchanged.
- No new tracing backend or cloud-provider-specific configuration is introduced.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -160,13 +160,13 @@ func main() {
 	// the calling site, so that structured logs (which by default omit it) are as
 	// locatable as klog-native logs.
 	//
-	// The explicit Encoder fixes the format to one JSON object per line with the
-	// trace ID as its first field (see tracing.NewTraceFirstJSONEncoder). It
-	// takes precedence over --zap-encoder, which therefore no longer switches
-	// the format; other zap flags (level, stacktrace) keep working.
+	// The explicit Encoder fixes the format to one JSON object per line with
+	// the trace ID first and ISO8601 timestamps (see tracing.NewTraceFirstJSONEncoder).
+	// Both --zap-encoder and --zap-time-encoding are overridden and have no effect;
+	// --zap-log-level and --zap-stacktrace-level remain effective.
 	opts.Encoder = tracing.NewTraceFirstJSONEncoder()
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts), zap.RawZapOpts(gozap.AddCaller())))
-	setupLog.Info("controller logger initialized with trace-first JSON encoder; --zap-encoder is overridden and has no effect")
+	setupLog.Info("controller logger initialized with trace-first JSON encoder and ISO8601 timestamps; --zap-encoder and --zap-time-encoding are overridden and have no effect")
 
 	if metricLabelsAllowlist != "" {
 		keys := strings.Split(metricLabelsAllowlist, ",")
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -28,6 +28,7 @@ import (
 	_ "time/tzdata" // Embed timezone database for scratch-based container images
 
 	"github.com/spf13/pflag"
+	gozap "go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -155,7 +156,17 @@ func main() {
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
 
-	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+	// AddCaller annotates each log line with the source file and line number of
+	// the calling site, so that structured logs (which by default omit it) are as
+	// locatable as klog-native logs.
+	//
+	// The explicit Encoder fixes the format to one JSON object per line with the
+	// trace ID as its first field (see tracing.NewTraceFirstJSONEncoder). It
+	// takes precedence over --zap-encoder, which therefore no longer switches
+	// the format; other zap flags (level, stacktrace) keep working.
+	opts.Encoder = tracing.NewTraceFirstJSONEncoder()
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts), zap.RawZapOpts(gozap.AddCaller())))
+	setupLog.Info("controller logger initialized with trace-first JSON encoder; --zap-encoder is overridden and has no effect")
 
 	if metricLabelsAllowlist != "" {
 		keys := strings.Split(metricLabelsAllowlist, ",")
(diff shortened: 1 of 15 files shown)


pkg/servers/e2b/routes.go
```diff
@@ -52,13 +53,16 @@ func (sc *Controller) registerRoutes() {
...
+	RegisterE2BRoute(sc.mux, http.MethodDelete, "/sandboxes/{sandboxID}", sc.DeleteSandbox, traceOperation(traceOpKill), sc.CheckApiKey)
 	RegisterE2BRoute(sc.mux, http.MethodPut, "/sandboxes/{sandboxID}/network", sc.UpdateSandboxNetwork, sc.CheckApiKey)
-	RegisterE2BRoute(sc.mux, http.MethodPost, "/sandboxes/{sandboxID}/pause", sc.PauseSandbox, sc.CheckApiKey)
-	RegisterE2BRoute(sc.mux, http.MethodPost, "/sandboxes/{sandboxID}/resume", sc.ResumeSandbox, sc.CheckApiKey)
+	RegisterE2BRoute(sc.mux, http.MethodPost, "/sandboxes/{sandboxID}/pause", sc.PauseSandbox, traceOperation(traceOpPause), sc.CheckApiKey)
+	RegisterE2BRoute(sc.mux, http.MethodPost, "/sandboxes/{sandboxID}/resume", sc.ResumeSandbox, traceOperation(traceOpResume), sc.CheckApiKey)
+	// Connect is labeled dynamically inside ConnectSandbox: it records
+	// "resume" only when it actually resumes a paused sandbox; connecting to
+	// a running sandbox keeps the default route pattern.
 	RegisterE2BRoute(sc.mux, http.MethodPost, "/sandboxes/{sandboxID}/connect", sc.ConnectSandbox, sc.CheckApiKey)
```
zmberg: RegisterE2BRoute(sc.mux, http.MethodPost, "/sandboxes/{sandboxID}/connect", sc.ConnectSandbox,  traceOperation(traceOpResume), sc.CheckApiKey)

Liquorice-Ma: done

cmd/agent-sandbox-controller/main.go
```diff
@@ -155,7 +156,17 @@ func main() {
...
-	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+	// AddCaller annotates each log line with the source file and line number of
+	// the calling site, so that structured logs (which by default omit it) are as
+	// locatable as klog-native logs.
+	//
+	// The explicit Encoder fixes the format to one JSON object per line with the
+	// trace ID as its first field (see tracing.NewTraceFirstJSONEncoder). It
+	// takes precedence over --zap-encoder, which therefore no longer switches
+	// the format; other zap flags (level, stacktrace) keep working.
+	opts.Encoder = tracing.NewTraceFirstJSONEncoder()
```
zmberg, line 167: **[Suggestion]** The comment and log message only mention that `--zap-encoder` is overridden. But once `opts.Encoder` is set explicitly, controller-runtime's `addDefaults()` no longer runs `o.Encoder = o.NewEncoder(o.EncoderConfigOptions...)`, so `--zap-time-encoding` (applied via `EncoderConfigOptions`) is silently disabled as well. The wording "other zap flags ... keep working" (and the L169 log line) is therefore inaccurate — please also note that `--zap-time-encoding` no longer takes effect.

pkg/tracing/zap_encoder.go
```diff
@@ -0,0 +1,160 @@
...
+		promoted = append(promoted, rest...)
+		return e.Encoder.EncodeEntry(entry, promoted)
+	}
+
+	spliced := splicePool.Get()
+	spliced.AppendByte('{')
+	spliced.AppendString(`"` + TraceIDLogKey + `":`)
+	spliced.AppendBytes(quoted)
+	spliced.AppendByte(',')
+	spliced.AppendBytes(encoded.Bytes()[1:])
```
zmberg, line 163: **[Suggestion]** If the base encoder emits an empty object `{}`, this splice produces `{"traceID":"...",}` (a trailing comma), which is invalid JSON. In practice zap always emits level/ts/msg so this never triggers, but the splice relies on an implicit assumption that the base output is a non-empty object. Consider falling back to re-encoding (or skipping the comma) when `encoded.Len() == 2` (i.e. `{}`).

zmberg (commented):
**[Issue]** `pkg/tracing/propagator.go` — `InjectTraceContext` (the `otel.GetTextMapPropagator().Inject(ctx, carrier)` call)

The composite propagator's `Baggage.Inject` does not call `carrier.Set` when the ctx carries no operation baggage. As a result, if `annotations` already holds a stale `trace-baggage` (left over from a previous operation), this Inject cannot overwrite or clear it. For background operations that write the CR without setting operation baggage (e.g. auto-pause / gateway wake / TriggerRecycle), `traceOperation` will be incorrectly attributed to the previous user operation. HTTP request paths are unaffected because they all pass through `WithTraceOperation`. Consider actively clearing the stale `trace-baggage` annotation when there is no operation baggage in ctx.

cmd/agent-sandbox-controller/main.go
Liquorice-Ma: done

pkg/tracing/zap_encoder.go
Liquorice-Ma: done

PR #955, opened 2026-09-11
feat(sandboxclaim): support claim-time auto-pause policy and probes

AiRanthem:

Allow SandboxClaim to configure absolute `pauseTime`, a complete `autoPausePolicy`, and named `probes` when claiming a Sandbox.

- Omitted fields inherit the existing Sandbox configuration. When either deadline is supplied, write `pauseTime` and `shutdownTime` together, clearing the omitted side.
- Replace an explicitly supplied auto-pause policy and merge probes by name, overriding matching probes and appending new ones with deep copies.
- Validate Claim configuration using the existing probe/policy rules and complete invalid Claims with `InvalidClaimSpec`.
- Check the actual candidate's probe compatibility during pool rollouts, including the merged 16-probe limit. Skip incompatible candidates so `createOnNoStock` can create a compatible Sandbox; recheck the limit on the creation path.
- Update the API, generated schema/deepcopy code, example, proposal, and focused tests.

NONE

The final candidate-limit fix passes:

```sh
go test -mod=readonly ./pkg/sandbox-manager/infra/sandboxcr \
  -run 'TestPickAnAvailableSandbox_CandidateCompatibilityUsesCandidateSpec|TestNewSandboxFromSandboxSet_TemplateRef' \
  -count=1
```

The regression cases cover oversized old candidates falling back to creation, same-name replacement at exactly 16 probes, and rejection of an oversized newly constructed Sandbox. Running the two rejection/fallback cases against the pre-fix implementation with a disposable Go overlay fails as expected.

Targeted auto-pause validation/merge, Claim option/completion, and Sandbox probe/pause tests also passed. `git diff --check` passes. The new Claim field schemas match the corresponding Sandbox schemas apart from their top-level descriptions.

This is a one-time claim overlay; completed Claims do not manage their delivered Sandboxes. Existing pause/resume execution and its feature-gate behavior are unchanged, and E2B HTTP does not expose Claim probes.

The existing controller-to-sandbox-manager dependency is intentionally deferred, with a TODO to move claim orchestration into SandboxClaim core and keep shared primitives in a neutral package.

### DIFF
--- a/pkg/sandbox-manager/infra/sandboxcr/claim.go
+++ b/pkg/sandbox-manager/infra/sandboxcr/claim.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"go.opentelemetry.io/otel/attribute"
 	"golang.org/x/time/rate"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -39,6 +40,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openkruise/agents/api/v1alpha1"
+	"github.com/openkruise/agents/pkg/autopause"
 	infracache "github.com/openkruise/agents/pkg/cache"
 	cacheutils "github.com/openkruise/agents/pkg/cache/utils"
 	"github.com/openkruise/agents/pkg/controller/sandboxset"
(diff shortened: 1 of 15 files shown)


pkg/sandbox-manager/infra/sandboxcr/claim.go
```diff
@@ -620,6 +625,17 @@ func pickAnAvailableSandbox(ctx context.Context, opts infra.ClaimSandboxOptions,
 			}
 			continue
 		}
+		if missing := missingRequiredProbeNames(obj.Spec.Probes, requiredProbeNames); len(missing) > 0 {
+			missingProbeNames = missing
```
furykerry, line 629: consider append instead of replace missingProbeNames

PR #967, opened 2026-09-15
feat: add memberlist encryption and peer mTLS for Manager and Gateway

AiRanthem:

Add independently configured memberlist encryption and peer mTLS to Sandbox Manager and Gateway. Unconfigured channels remain plaintext; invalid configured credentials fail instance startup instead of falling back to plaintext.

- Load a 32-byte memberlist key and existing runtime certificate bundles from explicitly referenced user-cluster Secrets. Validate certificate chains, usages, and the fixed runtime server name, and reject server certificates usable as accepted client credentials.
- Expose Manager flags and matching Gateway `PEER_*` environment variables. Require client certificates at the Manager TLS handshake and verified client certificates for Gateway refresh requests, while retaining Gateway GET health probes.
- Configure outbound peer HTTPS, disable redirects and environment proxies, and start the Gateway route receiver before memberlist. Preserve bounded route-update retries and close peer listeners and idle connections through existing cleanup paths.
- Add focused tests, Gateway configuration guidance, and the English peer-security and hosted-data-plane proposal.

NONE

- `git diff --check upstream/master...HEAD` passes.
- Added or updated tests cover credential inputs and loading, memberlist encryption, peer transport, Gateway authentication and health endpoints, Manager startup, and Secret references.
- Application tests and E2E tests were not rerun during PR preparation. Run the relevant changed-package tests and CI to validate the implementation.

- `--secret-config` now requires an explicit `namespace/name`; the implicit namespace fallback is removed.
- Plaintext peer updates also stop following HTTP redirects and using environment HTTP proxies.
- Credentials are startup snapshots and require restart to change. Remote failures do not guarantee readiness failure or member eviction.
- Certificate provisioning and production deployment rendering are outside this change. Gateway host-process shutdown wiring to `Stop` remains separate O6 work.

### DIFF
--- a/config/CLAUDE.md
+++ b/config/CLAUDE.md
@@ -1,0 +1,1 @@
+@./AGENTS.md
--- a/config/AGENTS.md
+++ b/config/AGENTS.md
@@ -1,0 +1,14 @@
+# Config
+
+## Deployment Configuration Scope
+
+- Deployment manifests and Kustomize deployment configuration in this subtree
+  are repository E2E test orchestration, not production configuration. Their
+  probes, arguments, and resource settings are not evidence of live deployments
+  or production requirements.
+- Keep concrete E2E orchestration changes with the corresponding deployment
+  work. A design that affects E2E execution should state the need to adjust test
+  orchestration without prescribing probe or deployment changes unless the user
+  explicitly includes that scope.
+- This deployment scope does not change the generated-artifact contract for
+  `config/crd/`; follow the repository generation rules for CRD manifests.
--- a/docs/proposals/20260908-sandbox-manager-peer-security-hosted-data-plane-CN.md
+++ b/docs/proposals/20260908-sandbox-manager-peer-security-hosted-data-plane-CN.md
@@ -1,0 +1,276 @@
+---
+title: Sandbox Manager Peer 安全与托管数据面关闭
+authors:
+  - "@AiRanthem"
+reviewers: []
+creation-date: 2026-09-08
+last-updated: 2026-09-09
+status: implementable
+see-also:
+  - "/docs/proposals/20260824-sandbox-manager-network-interface-peer-discovery.md"
+---
+
+# Sandbox Manager Peer 安全与托管数据面关闭
+
+## 摘要
+
+每个托管 sandbox-api 实例服务一个用户集群，关闭 ext-proc，保留 peer 通信。`7946` memberlist 共享密钥加密与 `7789` mTLS 是两个独立功能：Manager 用 flags、Gateway 用环境变量，各自通过 Secret 引用启用；未配置则保留对应通道的明文通信。Secret 不承载开关。已配置但本地凭据无效时，整个实例启动失败，不降级为明文。
+
+远端 Join 或同步失败只记录并继续其他 peer，不新增熔断或驱逐。仅 mTLS 配置不同不会隔离 memberlist 成员组；Ready 也不保证远端互通。托管部署终态要求两项保护都启用且配置兼容。Peer TLS 原样复用 runtime 凭据，保留独立加载与租户信任边界，并拒绝能冒充客户端的服务端证书。凭据按启动快照使用，变更通过重启生效；证书供应和部署不属于开源交付范围。
+
+## 背景
+
+Sandbox Manager 和 Sandbox Gateway 使用两条 peer 通道。`7946/TCP+UDP` 上的 memberlist 发现进程并维护成员关系；`7789/TCP` 上的路由同步在这些进程之间传递 Sandbox 状态。发现一个成员不等于认证其路由更新，保护路由更新也不能保护 memberlist 流量。
+
+托管 Sandbox Manager 称为 sandbox-api，部署在用户集群之外，每个托管实例只服务一个用户集群，通过接入该用户集群网络的网卡提供控制 API。它需要保留 peer 通信，同时不处理用户应用流量。因此，Envoy 外部处理服务（ext-proc）必须能够独立于路由同步关闭。
+
+Runtime 请求 TLS 消费已供应的凭据，Manager 和 Gateway 使用各自的客户端 bundle 格式。托管场景必须原样复用这些凭据及既有 runtime 服务端 bundle。证书签发与分发属于外部供应系统，不是 peer 安全要在开源仓库中新增的能力。
+
+## 设计终态
+
+### 范围与职责
+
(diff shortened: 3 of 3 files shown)


docs/proposals/20260908-sandbox-manager-peer-security-hosted-data-plane.md
```diff
@@ -0,0 +1,107 @@
...
+
+This check has a defined limit. It verifies the local server certificate, but it cannot prove that the same certificate authority has never issued some other accepted client certificate to an untrusted workload. Credential issuance and access controls must still prevent untrusted workloads from obtaining a peer client identity.
+
+Peer authorization is tenant-wide rather than role-based. Any holder of a client certificate accepted by the configured tenant trust may send route updates; the peer service does not infer Manager or Gateway roles from certificate names. This is the main authorization tradeoff of reusing runtime identities. A deployment that needs isolation between accepted credential holders requires a different identity boundary.
```
furykerry, line 59: role-based authorization is still required, only manager and gateway is allowed to send and receive route updates

config/CLAUDE.md
```diff
@@ -0,0 +1 @@
+@./AGENTS.md
```
furykerry, line 1: claude code should have read AGENTS.md, plz remove CLAUDE.md all together

cmd/sandbox-manager/main.go
```diff
@@ -205,15 +217,48 @@ func main() {
...
+	pflag.StringVar(&peerTLSServerKeyKey, "peer-tls-server-key-key", "",
+		"Secret data key for the private key of the server certificate. Empty uses \"tls.key\".")
+	pflag.StringVar(&peerTLSClientSecret, "peer-tls-client-secret", "",
+		"namespace/name of the Secret holding this process's agent-runtime client TLS bundle, used to send peer HTTPS "+
+			"(typically the same Secret as --runtime-client-cert-secret). "+
+			"Must be set together with --peer-tls-server-secret; leave both empty to keep plaintext peer HTTP. "+
+			"The --peer-tls-client-*-key data keys are read only when this is set.")
+	pflag.StringVar(&peerTLSClientCAKey, "peer-tls-client-ca-key", "",
+		"Secret data key for the CA trusted for outbound peer server certificates. Empty uses \"ca.crt\".")
+	pflag.StringVar(&peerTLSClientCertKey, "peer-tls-client-cert-key", "",
```
furykerry: it is not necessary to let user specify the secret data key , per https://github.com/openkruise/agents/pull/994, tls configure will try to load keys used by cert-manager , or keys used by identity provider

pkg/peersecurity/inputs.go
```diff
@@ -0,0 +1,123 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    10|Unless required by applicable law or agreed to in writing, software
```
furykerry: "10|" may be caused by mis operation, consider  Remove the 10| prefix so the header matches the boilerplate exactly

cmd/sandbox-manager/secretconfig.go
```diff
@@ -69,46 +70,26 @@ func parseSecretConfig(data map[string][]byte) (secretConfig, error) {
...
 // loadSecretConfig parses ref, then performs exactly one precise Get. It never
-// lists, watches, caches, or falls back to any other source.
-func loadSecretConfig(reader ctrlclient.Reader, ref, defaultNamespace string) (secretConfig, error) {
+// lists, watches, caches, or falls back to any other source. An empty ref is
+// rejected here even though resolveSecretSettings filters it out before calling.
+func loadSecretConfig(reader ctrlclient.Reader, ref string) (secretConfig, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), secretConfigLoadTimeout)
 	defer cancel()
-	namespace, name, err := parseSecretRef(ref, defaultNamespace)
+	secretRef, err := utils.ParseSecretRef(ref)
```
furykerry: secretRef, err := utils.ParseSecretRef(ref) is invoked before the if ref == "" guard. Consider Check ref == "" first, then parse, so the control flow reads naturally.

pkg/sandbox-gateway/server/server.go
```diff
@@ -160,38 +233,95 @@ func (s *Server) Start(ctx context.Context) error {
...
-		klog.InfoS("Starting sandbox-gateway peer server", "address", s.httpServer.Addr)
-		if err := s.httpServer.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+		klog.InfoS("Starting sandbox-gateway peer server", "address", lis.Addr().String())
+		if err := s.httpServer.Serve(lis); err != nil && !errors.Is(err, http.ErrServerClosed) {
 			klog.ErrorS(err, "Peer server failed to start")
 		}
 	}()
 
+	if err := s.peerManager.Start(ctx, "", s.memberlistBindPort); err != nil {
+		_ = s.httpServer.Shutdown(ctx)
```
furykerry: Issue: The shutdown error is discarded on an already-failing startup path.
Recommendation: Log it (klog.ErrorS) before returning the original error, so a shutdown failure isn't invisible during startup diagnosis.

pkg/peersecurity/inputs.go
AiRanthem: Fixed locally: removed the stray 10| prefix. The license header now matches hack/boilerplate.go.txt. I will include this in the next commit.

cmd/sandbox-manager/secretconfig.go
AiRanthem: Fixed locally: loadSecretConfig now checks for an empty ref before calling ParseSecretRef. The existing empty-ref error is unchanged. I will include this in the next commit.

pkg/sandbox-gateway/server/server.go
AiRanthem: Fixed locally: the startup failure path now logs a Shutdown error with klog.ErrorS and still returns the original memberlist startup error. I will include this in the next commit.

cmd/sandbox-manager/main.go
AiRanthem: Refactored

config/CLAUDE.md
AiRanthem: Shall we do this in a separate PR ?

PR #974, opened 2026-09-16
fix(storage): validate CSI mount paths

BH4AWS:


AiRanthem:
## Review

**Critical** — none.

**High**

1. **A malformed PATH can reject every mount.** (`validate.go:178`)
   An empty PATH entry (e.g. trailing colon) means cwd. The CLI runs with cwd `/`, so `/` gets protected, and every path "overlaps" with `/`. All mounts then fail with exit 64.
   Fix: skip empty PATH entries, or never protect `/` — it is already rejected as a mount target.

**Medium**

2. **One broken PATH entry blocks all mounts.** (`validate.go:185`)
   If any entry can't be resolved (e.g. `/etc/passwd/bin`), validation errors out — even for a safe target like `/data`.
   Fix: skip unresolvable entries, or document fail-closed as intentional.

3. **User paths can forge log lines.** (`validate.go:126`, `main.go:238`)
   The ENOTDIR reason embeds the raw path, which may contain `\n`. `main()` logs it unsanitized, and the new `symlink` / `validate-mount-path` subcommands go through this path.
   Fix: sanitize with `logs.SanitizeValue`, as done in #928.

**Low**

4. **Same failure, two exit codes.** With a file mid-path: `/x/file/sub/y` → exit 64 (permanent), but `/x/file/sub` passes validation, fails later in `MkdirAll` → exit 1 (retryable).

5. **TOCTOU.** Between validation and symlink creation, a sandbox process can swap path components. A full fix needs `openat2(RESOLVE_BENEATH)`; at minimum document the residual risk.

6. **`/dev` and `/sys` are not protected.** Confirm this is out of scope if intended.

7. **Housekeeping.** Exit 64 has no consumer in this repo — if it targets the runtime storage API, mention the companion change. The PR description is empty; please fill it in.

PR #993, opened 2026-09-21
fix(sandbox-manager): expose manager service on port 8080

furykerry:

Aligns the kustomize config under `config/` with the Helm chart so the sandbox-manager Service is exposed on `8080`, the manager Ingress splits control-plane vs data-plane traffic like the chart, and the gateway routes to the manager on `8080`. Also drops the now-obsolete gateway-less e2e matrix entries.

**Service / gateway port**
- `config/sandbox-manager/service.yaml`: Service port/targetPort `7788` -> `8080`, name `http-envoy` -> `manager`
- `config/sandbox-gateway/configmap.yaml`: envoy `manager_cluster` endpoint `sandbox-manager...svc:7788` -> `:8080`
- `config/sandbox-gateway-runtime-mtls/configmap-patch.yaml`: same `manager_cluster` endpoint update for the runtime-mTLS variant

**Ingress (matches chart `templates/ingress.yaml`)**
- Control Plane: `api.<domain>` (`/`) and `<domain>` (`/kruise/api`) -> `sandbox-manager:8080`
- Data Plane: `*.<domain>` (`/`) and `<domain>` (`/`) -> `sandbox-gateway:7788`
- TLS hosts now include `api.<domain>`, `*.<domain>`, `<domain>`

**CI**
- Remove the `without sandbox-gateway` matrix entries from the e2e workflows; the data plane now always goes through the gateway. The single-entry SDK-version workflows (`2.8.1`, `2.24.0`) are switched to run `with sandbox-gateway`.
- `hack/run-e2b-e2e-test.sh`: port-forward collapsed to always use `svc/sandbox-gateway 80:7788`; `e2e-e2b-latest.yaml` drops the `if: matrix.with_gateway` gate on the gateway build/deploy steps so the gateway is always available.

NONE

1. `kubectl apply -k config/sandbox-manager/` and `kubectl apply -k config/sandbox-gateway/`
2. Confirm the `sandbox-manager` Service exposes `8080` targeting the manager container (listens on `8080`, see `deployment.yaml`)
3. Confirm Ingress: `api.<domain>` and `<domain>/kruise/api` reach `sandbox-manager:8080`; `*.<domain>` and `<domain>/` reach `sandbox-gateway:7788`
4. Confirm the gateway envoy `manager_cluster` resolves `sandbox-manager.sandbox-system.svc.cluster.local:8080`

- The Service routes directly to the manager container, bypassing the manager pod's envoy sidecar (`7788`), matching the reference Helm chart.
- The gateway listener port `7788` (sandbox-gateway's own port) is intentionally unchanged; only the upstream `manager_cluster` endpoint is repointed.
- The e2e port-forward now always targets `svc/sandbox-gateway 80:7788` as the unified entry point, and `e2e-e2b-latest.yaml` builds/deploys the gateway unconditionally. The `quota redis fault` / `quota no redis` entries keep `with_gateway: false` only to skip gateway-specific feature tests; they still run behind the gateway.


AiRanthem:
### Blocker: `ingress_patch.yaml` indices were not updated

`config/sandbox-manager/kustomization.yaml` always applies `config/sandbox-manager/ingress_patch.yaml`. That patch still assumes the old 2-rule / 2-TLS-host Ingress:

```yaml
- op: replace
  path: /spec/rules/0/host
  value: "*.replace.with.your.domain"
- op: replace
  path: /spec/rules/1/host
  value: "replace.with.your.domain"
```

This PR adds `api.<domain>` as rules/tls index 0 and shifts `*` / apex to 1 / 2, but the patch is unchanged. Default `kubectl kustomize config/sandbox-manager` therefore clobbers the split:

| Rendered host | Backend | Intended |
|---|---|---|
| `*.replace.with.your.domain` | `sandbox-manager:8080` `/` | `api.` → manager |
| `replace.with.your.domain` | `sandbox-gateway:7788` `/` | `*.` → gateway |
| `replace.with.your.domain` | manager `/kruise/api` + gateway `/` | apex (kept, now duplicated) |

TLS hosts render as `*.domain`, `domain`, `domain` — `api.` is gone. Wildcard data-plane traffic would hit the manager instead of the gateway.

E2E does not catch this because it port-forwards and never exercises Ingress. Please extend `ingress_patch.yaml` with the `api.` host at index 0 (rules and tls) and shift the existing `*` / apex entries, then re-render with kustomize to confirm `api.` / `*.` / apex all survive.

### Nit: `CONTRIBUTING.md`

Still documents:

```
sudo kubectl port-forward services/sandbox-manager 80:7788
```

That port no longer exists on the Service. Point local testing at `svc/sandbox-gateway 80:7788`, matching the e2e entry point.

The Service `8080` change, gateway `manager_cluster` update (including runtime-mTLS), Ingress source split, and always-on gateway port-forward look good. Quota jobs keeping `with_gateway: false` only to skip gateway-specific tests is fine given they now still enter through the gateway.

